Issue #22, opened 2016-04-13
Consider removing the IO monad


Marwes:
From writing the REPL I get the feeling that a strict language does not work that well with monadic action once a few actions are chained. Adding better syntax for monadic actions could maybe make it convenient enough but as the language isn't pure anyway it is likely better to remove the IO monad for now.

Marwes:
Instead of removing the IO monad it may be worth adding some syntax sugar to make it easier to write monadic actions.

```fs
let test = 123
do let x = Some a
do None
None
```

```
do <expr> in <expr_2>
// Indentation works instead of in
do <expr>
<expr_2>
==>
<expr> >>= (\_ -> <expr_2>)
```

```
do let x = <expr> in <expr_2>
// Indentation works instead of in
do let x = <expr>
<expr_2>
==>
<expr> >>= (\x -> <expr_2>)
```


```
let opt =
    do let x = get map "abc"
    return x
match opt with
| Some y -> y
| None -> error "Expected Some"
```

bjz:
One idea for making this easier in the repl might be to have an `:x <expr>` command for executing IO actions, like in the Idris repl.

Imo, I would err on the side of keeping the ugliness for now, and figure out how to make it prettier later.

Marwes:
The repl already executes IO actions directly. Adding `:x <expr>` may make it more explicit though. I just mimicked ghci which also runs IO actions immediately.

PR #27, opened 2016-05-10
Remove the io monad

Marwes:
This PR removes the IO monad and allows all IO functions to simply be run as normal functions. Though this makes it easier to use IO it may still be worth exploring other less heavy handed approaches.
#22


puffnfresh:
Typed effects are more useful than untyped effects :-1:

Marwes:
I mostly agree with you though (which is why I haven't merged it). There are problems with the IO monad (and monads in general) though which I wrote a bit about in [this hackernews comment](https://news.ycombinator.com/item?id=12088664). With some novel ideas I think it could be solvable though.

PR #31, opened 2016-07-12
Minor fixes in README.md

TwoFX:
Just a few things that caught my eye when reading the readme :)


Marwes:
Thanks! Looks like I should be better at enabling spellchecking!

PR #35, opened 2016-07-16
Repl UX improvements

bjz:
- Use `"> "` as repl prompt
- Add basic help message at startup
- Skips parsing if an empty line is entered
- Adds previous expressions to linenoise history so that you can get back to them using the arrow keys
- Quit repl on ^C and ^D

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -77,9 +77,18 @@ fn find_info(args: WithVM<RootStr>) -> IO<Result<String, String>> {
 }
 
 fn compile_repl(vm: &Thread) -> Result<(), Box<StdError + Send + Sync>> {
-    fn input(s: &str) -> IO<Option<String>> {
-        IO::Value(::linenoise::input(s))
+    fn input(prompt: &str) -> IO<Option<String>> {
+        let input = ::linenoise::input(prompt);
+
+        if let Some(ref input) = input {
+            if !input.trim().is_empty() {
+                ::linenoise::history_add(input);
+            }
+        }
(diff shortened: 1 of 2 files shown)


bjz:
I've been having the same problem with `regex` building on nightly :(

Marwes:
Nice improvements! The repl hasn't seen much love for some time so I really appreciate this! I'd be happy to merge this as it is though I have one minor suggestion. I think the REPL could also skip over lines containing only whitespace (a call to `string.trim` before checking if its empty should do it).

I don't think we have to worry about the build failure on nightly either as it seems like a fix is about to get merged into rustc so hopefully that should be fixed in a day or two (and the failure on stable is just spurious timeouts which started after rust 1.10, may need to increase the timeouts a bit more until the next stable version of rust is released).

Feel free to squash some of the commits if you like but this is awesome regardless!

Marwes:
Oh, just a heads up since you made some changes to the linenoise integration. I started some work to replace the [linenoise](https://github.com/octplane/rust-linenoise/) dependency with [rustyline](https://github.com/kkawakam/rustyline). There shouldn't be any problems porting over any changes since it is mostly a rust implementation of linenoise but it also has a better completion api which should make it possible to add tab-completion in the REPL!

bjz:
Yeah - this compilation is taking forever... it is particularly slow for gluon_syntax. :(

bjz:
Done!

Marwes:
:tada:

PR #37, opened 2016-07-17
Add Category interface and piping operators

### LINKED ISSUE #34: Add F#/OCaml/Elm-style piping operators
This seems to be what all the cool kids are doing! It is also much easier for people to understand vs Haskell's `(.)` and `($)`:

```rust
/// Backward function application, where `f <| x == f x`
let (<|) f x : a -> (a -> b) -> b = f x

/// Forward function application, where `x |> f == f x`
let (|>) x f : (a -> b) -> a -> b = f x

/// Function composition, where `f >> g == (\x -> f (g x))`
let (>>) f g : (a -> b) -> (b -> c) -> a -> c = \x -> f (g x)

/// Function composition, where `f << g == (\x -> g (f x))`
let (<<) f g : (b -> c) -> (a -> b) -> a -> c = \x -> g (f x)
```

Seeing as we are thinking in higher kinds, I wonder if we could define these in terms of `Applicative` and `Category`?

bjz:
Closes #34

### DIFF
--- a/base/src/instantiate.rs
+++ b/base/src/instantiate.rs
@@ -8,119 +8,115 @@ use types;
 use types::{AliasData, BuiltinType, Type, Generic, TcType, TypeEnv, merge};
 use symbol::Symbol;
 
-pub struct AliasInstantiator<'a> {
-    pub inst: &'a Instantiator,
-    pub env: &'a TypeEnv,
+/// Removes type aliases from `typ` until it is an actual type
+pub fn remove_aliases(env: &TypeEnv, mut typ: TcType) -> TcType {
+    while let Ok(Some(new)) = maybe_remove_alias(env, &typ) {
+        typ = new;
+    }
+    typ
 }
 
-impl<'a> AliasInstantiator<'a> {
-    pub fn new(inst: &'a Instantiator, env: &'a TypeEnv) -> AliasInstantiator<'a> {
-        AliasInstantiator {
-            inst: inst,
-            env: env,
-        }
+pub fn remove_aliases_cow<'t>(env: &TypeEnv, typ: &'t TcType) -> Cow<'t, TcType> {
+    let mut typ = match maybe_remove_alias(env, typ) {
+        Ok(Some(new)) => new,
+        _ => return Cow::Borrowed(typ),
+    };
+    while let Ok(Some(new)) = maybe_remove_alias(env, &typ) {
+        typ = new;
     }
+    Cow::Owned(typ)
+}
 
-    /// Removes type aliases from `typ` until it is an actual type
-    pub fn remove_aliases(&self, mut typ: TcType) -> TcType {
-        while let Ok(Some(new)) = self.maybe_remove_alias(&typ) {
-            typ = new;
-        }
-        typ
-    }
+pub fn remove_alias(env: &TypeEnv, typ: TcType) -> TcType {
+    maybe_remove_alias(env, &typ).unwrap_or(None).unwrap_or(typ)
+}
 
-    pub fn remove_aliases_cow<'t>(&self, typ: &'t TcType) -> Cow<'t, TcType> {
-        let mut typ = match self.maybe_remove_alias(typ) {
-            Ok(Some(new)) => new,
-            _ => return Cow::Borrowed(typ),
-        };
-        while let Ok(Some(new)) = self.maybe_remove_alias(&typ) {
-            typ = new;
-        }
-        Cow::Owned(typ)
-    }
-
-    pub fn remove_alias(&self, typ: TcType) -> TcType {
-        self.maybe_remove_alias(&typ).unwrap_or(None).unwrap_or(typ)
-    }
-
-    pub fn maybe_remove_alias(&self, typ: &TcType) -> Result<Option<TcType>, ()> {
-        let maybe_alias = match **typ {
-            Type::Alias(ref alias) if alias.args.is_empty() => Some(alias),
-            Type::Data(ref alias, ref args) => {
-                match **alias {
-                    Type::Alias(ref alias) if alias.args.len() == args.len() => Some(alias),
-                    _ => None,
-                }
+pub fn maybe_remove_alias(env: &TypeEnv, typ: &TcType) -> Result<Option<TcType>, ()> {
+    let maybe_alias = match **typ {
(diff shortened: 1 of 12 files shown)


bjz:
Getting a type error in the prelude :/

```
std.prelude:Line: 298, Column: 40: Expected the following types to be equal
Expected: { id: 420 -> 420, compose: (429 -> 430) -> (431 -> 429) -> 431 -> 430 }
Found: { id: 433 434 434, compose: 433 435 436 -> 433 434 435 -> 433 434 436 }
4 errors were found during unification:
Types do not match:
	Expected: 420 -> 420
	Found: 433 434 434
Types do not match:
	Expected: 429 -> 430
	Found: 433 435 436
Types do not match:
	Expected: 431 -> 429
	Found: 433 434 435
Types do not match:
	Expected: 431 -> 430
	Found: 433 434 436
let category_Function: Category (->) = {
                                       ^
std.prelude:Line: 444, Column: 23: Expected the following types to be equal
Expected: 798 843 -> (843 -> 798 844) -> 798 844
Found: 798 843 -> 842 841 841 -> 798 844
Types do not match:
	Expected: 843 -> 798 844
	Found: 842 841 841
        join = \mm -> mm >>= id,
                      ^~~~~~~~~
```

Have I got something wrong, or is the unification having trouble with the function kind?

Marwes:
You are venturing into uncharted territory and found a compiler bug unfortunately. It first parses wrong and treats (->) as a generic variable (as it is called in the compiler there might be a better name). ie `(->) Int Int` looks like `a Int Int`). It probably also fails in the typechecker as `(->) Int Int` does not have the same internal shape as `(->) Int Int`. I have fixed the parser bug without to much trouble but the typechecking bug may be a bit trickier.

bjz:
Cool. Thankfully this is pushing the boundaries and forcing us to look at these bugs!

bjz:
Could you rebuild travis to see if it passes now? Otherwise I'll check tonight.

bjz:
So it's failing with this now:

```
	thread 'call_function' panicked at 'std.prelude:Line: 444, Column: 30: Could not resolve a binding for `id` with type `m010 m016 -> m010 m016`

Possibilities:

-> a a

-> a a

        join = \mm -> mm >>= id,

                             ^~
```

Marwes:
Meant to rebuild this before going to slow but I forgot.

Anyway that looks like an odd error. It looks like the typechecker fails to normalize a `Data((->), [a, a])` (displayed as `-> a a`) into `Function(a, a)` (displayed as `a -> a`), but that shouldn't even cause an error as there should only be a single `id` binding in scope which could be immediately selected! A bit busy today but I may have to to take a proper look sometime tonight.

https://github.com/Marwes/gluon/blob/22a4007d40a4107bd93c6c393a4a07d4b2ec2b30/base/src/instantiate.rs#L216
https://github.com/Marwes/gluon/blob/22a4007d40a4107bd93c6c393a4a07d4b2ec2b30/base/src/types.rs#L150

bjz:
I'll have a look. No promises though.

Marwes:
Yeah it is definitely a bug (or two), I may be able to respond from my cellphone if you have any questions.

bjz:
I switched to the `{:?}` formatter - certainly looks like it is normalizing it weirdly:

```
std.prelude:Line: 444, Column: 30: Could not resolve a binding for `id` with type `Function([Data(Generic(Generic { kind: Function(Star, Star), id: 0x10488eba0:m010 }), [Generic(Generic { kind: Star, id: 0x10488ef88:m016 })])], Data(Generic(Generic { kind: Function(Star, Star), id: 0x10488eba0:m010 }), [Generic(Generic { kind: Star, id: 0x10488ef88:m016 })]))`
Possibilities:
Data(Builtin(Function), [Generic(Generic { kind: Star, id: 0x104898028:a }), Generic(Generic { kind: Star, id: 0x104898028:a })])
Data(Builtin(Function), [Generic(Generic { kind: Star, id: 0x104898028:a }), Generic(Generic { kind: Star, id: 0x104898028:a })])

        join = \mm -> mm >>= id,
                             ^~
```

Marwes:
It would probably be a good idea to reduce the amount of normalising that needs to be done by removing the Function and App variants pf Type though that is not a sure way of fixing this though.

To fix the problem of two bindings it's probably a good idea to add in the specific bindings (at least the locations) to the error message (that would be good information to have not just for this issue)

Marwes:
Got a fix for the duplicate binding error which gets it past the renaming stage (since it properly has just a single `id` binding in scope it never needs to check that the types match and so the lack of normalizing is not noticed, the normalizing should still be fixed though). Got a panic in the compiler instead but it should be quite easy to fix though that will have to wait for tomorrow.

Marwes:
With #49 this PR will no longer have any compiler errors. It does however still fail two metadata tests which assume `id` exists in the prelude. While it is always possible to access `id` through `category_Function.id` I think it is a good idea to also export it directly through the prelude as it is useful function even if one does not know about `Category` (or category theory). Other than that :+1:

bjz:
Yup - good idea.

bjz:
Haha:

```
std.prelude:Line: 412, Column: 38: Expected the following types to be equal
Expected: { >>=: (737 -> 738) -> (738 -> 737 -> 740) -> 737 -> 740, return: 738 -> -> -> 737 738 }
Found: { >>=: -> 737 738 -> (738 -> -> 737 740) -> -> 737 740, return: 738 -> -> 737 738 }
Types do not match:
	Expected: ->
	Found: -> ->
let monad_Function: Monad ((->) a) = {
                                     ^~~
std.prelude:Line: 468, Column: 5: Expected the following types to be equal
Expected: a -> b -> a
Found: a -> -> -> b a
Types do not match:
	Expected: ->
	Found: -> ->
let const : a -> b -> a =
    ^~~~~
```

Marwes:
Hum. You just had to go and find another bug, didn't you :). I did test for that case in https://github.com/Marwes/gluon/commit/7281aefd48eae98f7eae056ec098d5cb91c0fe67 but maybe the type variable instead of a concrete type is what is messing it up. I can have a look in a bit

bjz:
Don't worry! This reminds me of the early days of Rust's generics - I was always finding bugs when trying to write cgmath! ^_^

Marwes:
And its finally merged!

PR #38, opened 2016-07-17
fix: Handle the function operator (->) being used in identifier position in types

Marwes:


parser/src/lexer.rs
```diff
@@ -350,10 +357,19 @@ impl<'a, 's, I, Id, F> Lexer<'a, I, F>
 
     /// Identifier parser which returns `(id, true)` if the identifier is a constructor
```
bjz: Stale comment

bjz:
Looks good!

parser/src/lexer.rs
Marwes: Oops, fixed.

Marwes:
@bjz I think this should fix the error in #37 but it unfortunately breaks two other tests in the typechecker. Should be solvable but I will have to think for a bit as I may need to refactor a bit.

base/src/instantiate.rs
```diff
@@ -196,24 +196,62 @@ pub fn instantiate<F>(typ: TcType, mut f: F) -> TcType
...
+        }
+        _ => return None,
+    };
     loop {
-        match *current {
+        match **current {
             Type::App(ref l, ref r) => {
                 args.push(r.clone());
-                current = &**l;
+                current = l;
```
bjz: Might be worth getting clippy to run on this project at some stage to see if you have any unnecessary refs/derefs...

Marwes: Those dereferences used to be necessary actually. In the commit I also changed the type of `current` to be `TcType` instead of `Type<Symbol>` (TcType derefs to `Type<Symbol>`). I suppose it might have been cleaner to declare the type of `current` to get automatic dereferences before though.

Marwes: I did run clippy some time ago on one or two of the crates. Only found style problems though so I didn't continue using it as it had some problems with multi crate projects at the time.

bjz: Would be cool to see if clippy could work with cargo workspaces - might make sense to set up a workspace for gluon anyway.

Marwes: As long as setting up a workspace still allows building on stable until it is stabilized then definitely.

Marwes: @bjz Side note: I noticed that I never mentioned the `test.sh` script in the readme. Just wanted to say to you can use that to run all tests in all the crates, it does output the binaries in `c-api/target` which is a bit annoying though. With workspaces I think this script should be obsolete though so I may test workspaces and if they work, just recommend nightly rustc for working on the library.

Marwes:
That should do it, I had to add the `SymbolRef` type (symbols identify a type, binding, function etc uniquely and they can be compared by pointer equality). It unfortunately takes 2 pointers worth of space which is a bit wasteful but that can be fixed later.

PR #44, opened 2016-07-17
Implement Applicative and Functor for State and Writer

### LINKED ISSUE #43: Implement Applicative and Functor for State and Writer
I reckon it would be nice to make `Monad` inherit from `Applicative`, and to remove `return` in favor of `pure`. In order to do that however, we need to implement the missing interfaces for the `State` and `Writer` types.

bjz:
Fixes #43


Marwes:
Looks good to me :+1:

PR #45, opened 2016-07-18
refactor: Replace linenoise with rustyline

Marwes:
Rustyline is a pure rust implementation of linenoise with some additional features which will make it possible to add completion to the repl.

The windows support in Rustyline is currently only partial (history support and modifier keys are not yet working). I will likely add in completion support as well before merging.

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -1,7 +1,15 @@
+extern crate rustyline;
+
 use std::error::Error as StdError;
+use std::fmt;
+use std::sync::Mutex;
+
+use self::rustyline::error::ReadlineError;
+
 use base::ast::Typed;
 use base::types::Kind;
-use vm::api::{IO, Function, WithVM};
+use vm::api::{IO, Function, WithVM, VMType, Userdata};
+use vm::gc::{Gc, Traverseable};
 use vm::thread::{Thread, RootStr};
 
 use gluon::{Compiler, new_vm};
@@ -66,8 +74,8 @@ fn find_info(args: WithVM<RootStr>) -> IO<Result<String, String>> {
         }
     }
     let maybe_comment = env.get_metadata(args)
-                           .ok()
-                           .and_then(|metadata| metadata.comment.as_ref());
+        .ok()
+        .and_then(|metadata| metadata.comment.as_ref());
     if let Some(comment) = maybe_comment {
         for line in comment.lines() {
             write!(&mut buffer, "\n/// {}", line).unwrap();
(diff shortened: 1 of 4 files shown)


std/repl.glu
```diff
@@ -94,14 +94,14 @@ let loop x: () -> IO () =
...
         match line_opt with
             | None -> return ()
             | Some line -> run_line line >>= \continue ->
-                if continue then loop () else return ()
+                if continue then loop editor else return ()
 
 let run x: () -> IO () =
     io.print "gluon (:h for help, :q to quit)"
-        >>= loop
+        >>= \_-> loop (rustyline.new_editor ())
```
bjz: Does `(>>)` work here?

src/repl.rs
```diff
@@ -76,25 +82,55 @@ fn find_info(args: WithVM<RootStr>) -> IO<Result<String, String>> {
...
+struct Editor(Mutex<::rustyline::Editor<()>>);
 
-        if let Some(ref input) = input {
-            if !input.trim().is_empty() {
-                ::linenoise::history_add(input);
-            }
-        }
+impl fmt::Debug for Editor {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "File")
```
bjz: Kind of a weird debug string... why not `"Editor(..)"`?

std/repl.glu
Marwes: Yes it should. I may be a bit averse to using >> since it may actually
change the semantics of a program since gluon isn't pure but in this
instance I should have used >>

src/repl.rs
```diff
@@ -76,25 +82,55 @@ fn find_info(args: WithVM<RootStr>) -> IO<Result<String, String>> {
...
+    Userdata(Editor(Mutex::new(editor)))
+}
 
-        IO::Value(input)
+fn readline(editor: &Editor, prompt: &str) -> IO<Option<String>> {
+    let mut editor = editor.0.lock().unwrap();
+    let input = match editor.readline(prompt) {
+        Ok(input) => input,
+        Err(ReadlineError::Eof) => return IO::Value(None),
+        Err(err) => return IO::Exception(format!("{}", err)),
```
bjz: Ending the program on `ReadlineError::Eof` (ie. ^D) makes sense - but what about `ReadlineError::Interrupted` (ie. ^C)? I would say if it is easiest, return `IO::Value(None)`, but in the future it would be nice for it to kill a long-running expression.

https://kkawakam.github.io/rustyline/rustyline/error/enum.ReadlineError.html

src/repl.rs
Marwes: Copy paste mistake, will fix. The implementation is only there to satisfy the userdata trait which requires it.

src/repl.rs
```diff
@@ -1,7 +1,13 @@
 use std::error::Error as StdError;
+use std::fmt;
+use std::sync::Mutex;
+
+use rustyline::error::ReadlineError;
```
bjz: Maybe scope the crate import to this module, seeing as it is not used elsewhere? Ie.

```rust
extern crate rustyline;
use self::rustyline::error::ReadlineError;

...

let editor = rustyline::Editor::new();
```

std/repl.glu
bjz: Oh, interesting. In what ways is it not pure?

src/repl.rs
Marwes: That would make more sense. Forgot that the ctrl-c case was wrapped as none in line noise. And a way to kill long running expressions would be really nice. I have a branch which I thought I had merged which allows users to restrict the max heap size. Should probably merge that as it is a useful part of sandboxing

std/repl.glu
Marwes: I did a pretty through explanation of my reasoning in a hackenews thread. Let me know if you have any thoughts or questions as it's not set in stone. https://news.ycombinator.com/item?id=12088664

src/repl.rs
Marwes: Wasn't aware you could put extern crate in other modules than the root actually.

bjz:
Looks good to me!

Marwes:
Merging this even though it still needs a git dependency and Windows support is not 100% as I have been unable to resolve some compilation errors I get with linenoise-rust (I think either my msvc installation is either broken or I didn't install everything needed). In any case that seems like it should be fixed soon so having the REPL on Windows miss a few things for a week or so is not a big problem

PR #49, opened 2016-07-20
fix(check): Fix errors uncovered from using (->) partially applied

Marwes:
Also fixes some fallout which was masked by fixing the double listing


Marwes:
@bjz Though this will fix the immediate error in #37 I think I will take a look at the normalizing issue as well before merging this. If you like I can merge this right away though.

bjz:
Take your time - I'm in no hurry! :)

bjz:
That works :)

Marwes:
Yep, I was just taking a look at the normalizing problem (which isn't actually caused by normalizing but by the unification traversal function changing the order of the arguments).

PR #54, opened 2016-07-21
fix(check): Don't bind uninstantiated type signatures to arguments

Marwes:
Without instantiating the type signature all uses of the function arguments would instantiate the argument's type with fresh variables on every use which easily masked the problem by allowing wrongly typed code to pass.


Marwes:
There is still at least one problem I need to fix before merging this as after modifying state.glu I am getting a stack overflow. Still pushing this PR up before going to sleep as a warning as the error that is being fixed is quite bad :sweat:

bjz:
Hooray for soundness issues!

Issue #50, opened 2016-07-20
Simplify the Type representation and reduce its heap usage


Marwes:
Currently there are quite a lot of different representation in the [Type enum](https://github.com/Marwes/gluon/blob/d1aaee3148d8954ccf37a1dd12d47c97a9597fd4/base/src/types.rs#L150-L181) this leads to complications in unification (and elsewhere) as types that should be equivalent can have multiple representations. To some extent these different representations are necessary either for convenience or for efficiency (`Function(Int, Float)` is a lot nicer to work with than `App(App(->, Int), Float)`).

Currently the `App` variant only exists to avoid allocation a one element vector for single argument constructors which are very common but with [smallvec](https://crates.io/crates/smallvec) (or equivalent) the `App` variant could be completely superseded by `Data`.

Other simplifications may be to replace`Function(a, b)` with `Data(-> [a, b])` and `Array(a)` with `Data([], [a]`.

Marking this as beginner as this change (though maybe a bit tedious) should just involve replacing all uses of `Type::App(a, b)` with `Type::Data(a, [b])`.

bjz:
Here is a checklist:

- [x]: Remove `Type::App` (#52)
- [ ]: Remove `Type::Function` (#55)
- [ ]: Remove `Type::Array`

(I would put it in the issue description, but I would need to be a contributor! hint hint... 😉)

Marwes:
Made you a contributor now, "With great power comes great responsibility" and all that :smile: .

bjz:
Hah, of course! ^_^

Marwes:
Also enabled the option that all forces travis checks to succeed before it being possible to merge but that might be a bad idea in case spurious errors (skeptic has been acting up the last week or so). Are you able to restart travis builds?

bjz:
Good idea. And yes, after force-syncing my travis account, I have a retry icon on builds now.

bjz:
Should `Array` be a `BuiltinType`?

Marwes:
Yes. I expect partially applying array does not work now then either. It may be better to remove the `[a]` syntax for array types as I am not sure the extra syntax carries its own weight.

bjz:
Are arrays even used anywhere? I guess I just haven't seen them in the tests...

bjz:
Yeah, I would prefer not to have a special syntax - what about just `Array a`?

Marwes:
There exists a few array tests in `tests/vm.rs` but thats about it.

[...]

Issue #41, opened 2016-07-17
Improve sharing of types and kinds


Marwes:
Anytime `Type::int`, `Type::float`, `Kind::star`, is called a new reference count type or kind is allocate which is rather wasteful. As these types are quite common they could be cached in the typechecker (and else where) so that the same pointer could be reused. The kindchecker already does some of this but in the typechecker this is not done.

https://github.com/Marwes/embed_lang/blob/master/check/src/kindcheck.rs

[PrimitiveEnv](https://github.com/Marwes/gluon/blob/80d300a33c791aa34e91f0794b08e967af98c2b9/base/src/types.rs#L115-L117) might be possible to extend for this (though it was originally created to retrieve a canonical boolean).

zjhmale:
i can see `Kind::start` is always called a new reference, but `Type::Int` and `Type::Float` seems always use the same reference, am i missing some code here?

Marwes:
While `Type::Int` does not allocate usually one will use [Type::int](https://github.com/Marwes/gluon/blob/51b85f44b6a2939714d40daf0d6fc67704f76c6d/base/src/types.rs#L489) which forwards to `Type::builtin`[https://github.com/Marwes/gluon/blob/51b85f44b6a2939714d40daf0d6fc67704f76c6d/base/src/types.rs#L457). This allocates a new reference counted pointer every time it is called and it is that waste I am talking about.

Marwes:
#275 does the easy/obvious improvements towards this issue. It can however be further improved by not only caching the leaf types (`Int`, `Bool`, `()` etc) but also dynamically caching nested types (`a -> Int` etc). This kind of dynamic caching requires a lot more work since all type creation will need to go through the `TypeCache` type. But if it is done it would be possible to greatly improve performance of unification in some cases as pointer equality could be used to determine if two types are equal without needing to check their contents.

PR #55, opened 2016-07-22
Remove Type::Function

bjz:
Working towards #50

### DIFF
--- a/base/src/types.rs
+++ b/base/src/types.rs
@@ -160,8 +160,6 @@ pub enum Type<Id, T = ASTType<Id>> {
     /// Variant for "generic" variables. These occur in signatures as lowercase identifers `a`, `b`
     /// etc and are what unbound type variables are eventually made into.
     Generic(Generic<Id>),
-    /// A function type `<type> -> <type>`
-    Function(Vec<T>, T),
     /// A builtin type
     Builtin(BuiltinType),
     /// An array type `Array T`
@@ -495,10 +493,13 @@ impl<Id, T> Type<Id, T>
         })
     }
 
-    pub fn function(args: Vec<T>, ret: T) -> T {
+    pub fn function(args: Vec<T>, ret: T) -> T where T: Clone {
+        let function: T = Type::builtin(BuiltinType::Function);
         args.into_iter()
             .rev()
-            .fold(ret, |body, arg| T::from(Type::Function(vec![arg], body)))
+            .fold(ret, |body, arg| {
+                Type::app(function.clone(), vec![arg, body])
+            })
     }
 
     pub fn generic(typ: Generic<Id>) -> T {
@@ -553,6 +554,17 @@ impl<Id, T> Type<Id, T>
 impl<Id, T> Type<Id, T>
     where T: Deref<Target = Type<Id, T>>
 {
+    pub fn as_function(&self) -> Option<(&T, &T)> {
+        if let Type::App(ref app, ref args) = *self {
+            if args.len() == 2 {
+                if let Type::Builtin(BuiltinType::Function) = **app {
+                    return Some((&args[0], &args[1]));
+                }
+            }
+        }
+        None
+    }
+
     pub fn as_alias_symbol(&self) -> Option<&Id> {
         match *self {
             Type::App(ref id, _) => {
@@ -609,25 +621,14 @@ impl<'a, Id, T> Iterator for ArgIterator<'a, T>
 {
     type Item = &'a T;
     fn next(&mut self) -> Option<&'a T> {
-        match **self.typ {
-            Type::Function(ref arg, ref return_type) => {
-                self.typ = return_type;
-                Some(&arg[0])
-            }
-            _ => None,
-        }
+        self.typ.as_function().map(|(arg, ret)| {
+            self.typ = ret;
+            arg
+        })
     }
 }
 
 impl<Id> ASTType<Id> {
-    /// Returns the type which this type would return if it was fully applied.
-    pub fn return_type(&self) -> &ASTType<Id> {
-        match **self {
-            Type::Function(_, ref return_type) => return_type.return_type(),
-            _ => self,
-        }
-    }
-
     /// Returns the lowest level which this type contains. The level informs from where type
     /// variables where created.
     pub fn level(&self) -> u32 {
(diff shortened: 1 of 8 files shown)


bjz:
Don't merge yet - I messed something up. Still figuring out my way around the code...

base/src/types.rs
```diff
@@ -498,7 +496,9 @@ impl<Id, T> Type<Id, T>
     pub fn function(args: Vec<T>, ret: T) -> T {
         args.into_iter()
             .rev()
-            .fold(ret, |body, arg| T::from(Type::Function(vec![arg], body)))
+            .fold(ret, |body, arg| {
+                Type::app(Type::builtin(BuiltinType::Function), vec![arg, body])
```
Marwes: It would be a bit more efficient to lift `Type::builtin(BuiltinType::Function)` out of the fold and `.clone()` it. (#41 is a bit more relevant after removing this change as well).

base/src/types.rs
```diff
@@ -610,24 +619,16 @@ impl<'a, Id, T> Iterator for ArgIterator<'a, T>
     type Item = &'a T;
     fn next(&mut self) -> Option<&'a T> {
         match **self.typ {
-            Type::Function(ref arg, ref return_type) => {
-                self.typ = return_type;
-                Some(&arg[0])
+            Type::App(ref app, ref args) if app.is_function() => {
```
Marwes: Maybe it is worth it to have `Type::as_function(&self) -> Option<(&T, &T)>` to make this a bit cleaner and less error prone?

check/src/unify_type.rs
```diff
@@ -370,51 +342,6 @@ fn unify_app_<'a, 's, U>(unifier: &mut UnifierState<'a, 's, U>,
...
-        None => {
-            let new_fn = matcher(unifier,
-                                 fn_prim,
-                                 &Type::app(Type::builtin(types::BuiltinType::Function),
-                                             vec![arg.clone()]));
-            let new_ret = matcher(unifier, ret, other_ret);
-            Ok(merge(fn_prim, new_fn, ret, new_ret, |l, r| Type::app(l, vec![r])))
-        }
-    }
-}
```
bjz: I removed this because it wasn't used, but maybe we are missing some logic somewhere?

check/src/typecheck.rs
```diff
@@ -983,10 +983,10 @@ impl<'a> Typecheck<'a> {
             return Ok(typ);
         }
         match *typ {
-            Type::Function(ref argument_types, ref return_type) => {
-                assert!(argument_types.len() == 1);
-                self.stack_var(args[0].id().clone(), argument_types.last().unwrap().clone());
-                self.typecheck_pattern_rec(&args[1..], return_type.clone())
+            Type::App(ref app, ref arg_types) if app.is_function() => {
+                assert!(arg_types.len() == 2);
+                self.stack_var(args[0].id().clone(), arg_types.last().unwrap().clone());
```
Marwes: Should be `arg_types[0]` (no idea why I used `last`)

check/src/typecheck.rs
```diff
@@ -1293,19 +1293,16 @@ impl<'a, 'b> Iterator for FunctionArgIter<'a, 'b> {
     type Item = TcType;
     fn next(&mut self) -> Option<TcType> {
         loop {
-            let (arg, new) = match *self.typ {
-                Type::Function(ref args, ref ret) => (Some(args[0].clone()), ret.clone()),
```
Marwes: I think the `Type::Function` case still needs to be handled explicitly.

check/src/unify_type.rs
```diff
@@ -193,10 +168,7 @@ fn do_zip_match<'a, 's, U>(self_: &TcType,
 // Retrieves the alias name of an application if it exists.
 // (Result e Int ==> Result, a -> b ==> ->
 fn alias_sym(l: &TcType) -> Option<&SymbolRef> {
```
Marwes: I think you could just remove this function.

check/src/unify_type.rs
Marwes: Its fine to remove, it just existed to handle the special case of `Function <=> App` unification.

base/src/types.rs
Marwes: Where `(&T, `&T)` is `(argument type, return type)`

Marwes:
There is one thing I forgot to mention in #50 about `Type::Function` which is that it may be worth it to keep the `Type::Function` representation to represent functions more efficiently by using the vector represent multiple arguments (`Type::app(->, [a, Type::app(->, [b, c)])])` vs `Type::Function([a, b,], c)`. As functions are the most common type it may be a worthwhile optimization.

Premature optimization is the root of all evil though so might be best to do that optimization after determining that it is actually needed.

bjz:
Addressed the requested issues! :)

bjz:
Wait - don't merge this. Experimentation in the repl revealed some problems, namely:

```
λ cargo run -- -i
     Running `target/debug/repl -i`
gluon (:h for help, :q to quit)
> :l std/prelude.glu
thread '<unnamed>' panicked at 'ICE: expected a function application with two arguments, found 1', base/src/types.rs:560
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Any', ../src/libcore/result.rs:788
error: Process didn't exit successfully: `target/debug/repl -i` (exit code: 101)
```

base/src/types.rs
```diff
@@ -553,6 +554,16 @@ impl<Id, T> Type<Id, T>
 impl<Id, T> Type<Id, T>
     where T: Deref<Target = Type<Id, T>>
 {
+    pub fn as_function(&self) -> Option<(&T, &T)> {
+        if let Type::App(ref app, ref args) = *self {
+            if let Type::Builtin(BuiltinType::Function) = **app {
+                assert!(args.len() == 2, "ICE: expected a function application with two arguments, found {}", args.len());
```
bjz: Should I return `None` here? :/

Marwes:
Feel free to merge it after fixing that panic!

base/src/types.rs
Marwes: Yep that seems to be correct behavior as it can only handle fully applied functions/functions that have both argument and return value.

Issue #47, opened 2016-07-19
Explore possibilities for modular implicits


bjz:
At the moment module polymorphism is kind of clunky. Whilst our priority is more focused on polishing what we already have, it might be good to note down our thoughts here.

bjz:
Here is an idea that uses extensible records to implement module inheritance:

```ocaml
type (f: Functor) =
    { t : Type -> Type -> Type
    , map : [a, b] -> (a -> b) -> f.t a -> f.t b
    }

type (f: Applicative) =
    { Functor
    | apply : [a, b] -> f.t (a -> b) -> f.t a -> f.t b
    , const : [a, b] -> f.t a -> f.t b -> f.t b
    }

type (m : Monad) =
    { Applicative
    | flat_map : [a, b] -> (a -> m.t b) -> m.t a -> m.t b
    }

let (>>=) : [m : Monad] -> m.t a -> (a -> m.t b) -> m.t b
    (>>=) [m] = flip m.flat_map
```

bjz:
One issue with the above is how to handle multi-parameter inheritance.

Maybe:

```ocaml
type (m : Monad) =
    { foldable : Foldable
    , applicative : Applicative
    , t : foldable.t & applicative.t
    , flat_map : [a, b] -> (a -> m.t b) -> m.t a -> m.t b
    }
```

bjz:
One issue with that last approach is traversing up the hierarchy to get to the function you want:

```haskell
foo.monad.applicative.functor.map f xs
```

blegh

Marwes:
```f#
m.applicative.functor.map f xs
```
Yeah that does not look fun. Row polymorphism might be useful here which purescript uses.

```f#
type Functor f = {
    map : (a -> b) -> f a -> f b,
    | r
}

type Applicative f = {
    map : (a -> b) -> f a -> f b,
    (<*>) : f (a -> b) -> f a -> f b,
    pure : a -> f a,
    | r
}
```

This would mean any `Àpplicative` is also a `Functor`.

With the ability to 'splat' records it might be pretty easy to write.

```f#
let applicative_list = {
    (<*>) = ..,
    pure = ...,
    .. functor_list // splat the functor instance into this record
}
```

https://leanpub.com/purescript/read#leanpub-auto-record-patterns-and-row-polymorphism

bjz:
Your need another type parameter in your examples. ie:

```haskell
type Functor (f : Type -> Type) (r : Row) = {
[...]

Issue #11, opened 2015-10-31
Fix the Type::Function variant


Marwes:
Currently the `Function` variant has holds a `Vec<Type>` for function arguments but it is always assumed that the vector only has one element as functions are curried and are always treated as 1-argument functions.

The variant should either just hold one argument type, mirroring how the type checker sees it, or the code should take advantage of the `Vec` to store multiple arguments (possibly for efficency's sake).

Marwes:
`Type::Function` was removed in #55

PR #56, opened 2016-07-23
Library overhauls

### LINKED ISSUE #36: Favor giving names to functions over operators
It might be nice to provide named functions for interfaces first, then have the operators as shorcuts. So for example, `(>>=)` would become `flat_map` (I personally prefer this over `bind`, heh), and `make_Monad` would define the operator based on that.

bjz:
Fixes #36

This is a huge overhaul of the prelude API. It removes duplication, and give names to most of the more obscure operators. I've also enforced interface inheritance using fields, for example:

```diff
 type Applicative f = {
-    (<*>) : f (a -> b) -> f a -> f b,
+    functor : Functor f,
+    apply : f (a -> b) -> f a -> f b,
     pure : a -> f a
 }
```

This allows us to remove some of the parameters from the `make_*` functions, and is more self-documenting. It is also a nice step along the road to modular implicits (see #47). Despite my concerns (https://github.com/Marwes/gluon/issues/47#issuecomment-233562336), using fields doesn't seem too onerous, at least at this stage. It was neat to see that the language as it stands already allowed us to do this! 😄 

I have a feeling that the new API is beginning to hurt compile times, but I would rather we tackle that in a separate pass. There are already some PRs out about this (#50, #41, #11), but it might also pay to start profiling it to see if there are any places we can improve things.

I'm guessing the continual calling of the `make_*` functions is causing lots of allocations. Maybe the VM could be smarter about this? If we had a better effect system, perhaps we could figure out what is constant, and cache values based on their input args.... I dunno. Probably needs to be a separate issue. And as always, speculating on performance is usually pretty ineffective without profiling!


src/io.rs
```diff
@@ -204,7 +204,7 @@ pub fn load(vm: &Thread) -> Result<()> {
     vm.add_bytecode("io_bind", io_bind_type, 3, io_bind);
```
bjz: I would like to rename this to `io_flat_map`, but it wasn't very clear to me how to rewrite it with the arguments swapped. Still haven't got my head around the VM's bytecode! :o

std/prelude.glu
```diff
@@ -322,169 +361,169 @@ let functor_Result: Functor (Result e) = {
...
-    let { (>>=), return } = m
+    let { applicative, flat_map } = m
+    let { pure } = applicative
     let { id } = category_Function
 
-    let (>>) l r = l >>= \_ -> r
-    let forM_ xs f =
+    let (=<<) : (a -> m b) -> m a -> m b = flat_map
+    let (>>=) : m a -> (a -> m b) -> m b = flip flat_map
+    let (>>) l r : m a -> m b -> m b = l >>= \_ -> r
```
bjz: Still need to remove this in favor of `(*>)`

src/io.rs
Marwes: I believe this should work (untested)
```
// Stack
//  [f, m, ()]
let io_bind = vec![Call(1), // [f, m_ret] // Call m ()
                   PushInt(0), // [f, m_ret, ()] // Add a dummy argument ()
                   TailCall(2)]; // [f_ret] // Call f m_ret ()

```
IO actions are implement as simply having an extra argument (`IO a = () -> a` so invoking an action just needs to add an extra dummy argument and calling the function.

I can add some documentation to the bytecode though if you want, currently the documentation is just "what the vm does".

bjz: That would be amazing! I reckon the right place would be on the variants.

bjz: That worked as expected 👍

Marwes:
> Despite my concerns (#47 (comment)), using fields doesn't seem too onerous, at least at this stage...

This looks good indeed. One way to otherwise get some niceness back if we feel it is necessary would be to add the fields of an "inherited record" to the result of the `make_*` functions.

> I have a feeling that the new API is beginning to hurt compile times, but I would rather we tackle that in a separate pass.

Definitely need to properly profile this as I don't have a good idea of where the time is spent at the moment. Wonder if there is a good way to setup automatic ratcheting of benchmarks with travis so that it is possible to track performance regressions automatically (travis might have to variable times though).

> I'm guessing the continual calling of the make_* functions is causing lots of allocations...

Yep. A sufficiently smart compiler would be able to remove the allocations but the current compiler is as dumb as they come :). Actually implementing the necessary those optimizations may take some time (and I don't see gluon as a language which should be doing much optimizations, at least at the moment). I have an idea for how we may provide a workaround in the meantime though with macros (I will post an issue in a bit).

bjz:
Agreed about being conservative about optimisations. Less magic would be great, if possible.

bjz:
Speak of the devil... https://travis-ci.org/Marwes/gluon/jobs/146827271#L632

> thread 'tutorial_13' panicked at 'command failed', /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/skeptic-0.4.0/lib.rs:308

Retrying the build...

bjz:
Succeeded on nightly and stable though. Probably a spurious failure.

Marwes:
And I have already restarted that build once...

bjz:
Seems to pass now.

bjz:
Rebuilt based on the latest master - is there anything that you had issues with? Sorry for it being such a large PR - got a little carried away :(

Marwes:
Nope, looks good to me!

PR #61, opened 2016-07-23
Handle infinite recursive types

Marwes:
Still struggling to get the right semantics of when to bail out in the unifier. Currently I am getting either bad error messages or errors in typechecking. Making a PR to show WIP though.

### DIFF
--- a/base/src/instantiate.rs
+++ b/base/src/instantiate.rs
@@ -8,6 +8,11 @@ use types;
 use types::{AliasData, BuiltinType, Type, Generic, TcType, TypeEnv, merge};
 use symbol::Symbol;
 
+pub enum Error<I> {
+    SelfRecursive(I),
+    UndefinedType(I),
+}
+
 /// Removes type aliases from `typ` until it is an actual type
 pub fn remove_aliases(env: &TypeEnv, mut typ: TcType) -> TcType {
     while let Ok(Some(new)) = maybe_remove_alias(env, &typ) {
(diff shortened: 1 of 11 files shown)


bjz:
I've merged the function branch - you might want to rebase.

base/src/instantiate.rs
```diff
@@ -8,6 +8,11 @@ use types;
 use types::{AliasData, BuiltinType, Type, Generic, TcType, TypeEnv, merge};
 use symbol::Symbol;
 
+pub enum Error<I> {
+    SelfRecursive(I),
+    UndefinedType(I),
+}
```
bjz: Not using `quick-error` here?

Marwes: Thats a bit sloppy of me, only needed a way to distinguish the two error cases from each other as the only place I use it is to convert it into another error. Should be quick to add though.

Marwes:
This should be a bit better. The first refactoring commit may confuse things a bit but look at the added test cases to see the types of cases that should now be handled instead of causing a stack overflow by continuing to expand aliases.

In most languages this handled by not allowing recursive aliases but I am hoping that the way gluon handles it may work out to be convenient in the end. The short version is just that it expands an alias only when it is not ambiguous to do so and "aliases" are otherwise treated as opaque types. I will try to add some more through explanation on this (and overloading) to the tutorial sometime next week though.

check/tests/fail.rs
```diff
@@ -0,0 +1,280 @@
...
+";
+    let result = typecheck(text);
+    assert_unify_err!(result, Other(SelfRecursive(..)));
+}
+
+#[test]
+fn detect_self_recursive_aliases() {
+    let _ = ::env_logger::init();
+    let text = r"
+type A a = A a
```
bjz: Just clarifying: so this an alias, not a union of one variant, because it isn't preceded by a pipe?

Also, does this mean that types aren't loaded into the context top-down?

Marwes: Yes that is just an alias, variants always have to be preceded by a pipe.

Types are put in scope as they are declared by using `and` it is possible to have two or more types which are mutually recursive though (just like functions).

base/src/instantiate.rs
Marwes: Added quick-error for this error type

check/tests/fail.rs
bjz: But how can `type A a = A a` be recursive, when `A : Type -> Type` has not been defined yet?

Marwes: `A` is put into scope *before* doing checking the right hand side. `kindcheck.rs` also does kind inference so the kind of `A` is set to a kind variable when it is put in scope.

Marwes: Oh, and any kind variables that does not get their kind inferred [defaults](https://github.com/Marwes/gluon/blob/de751ebece1ab252af07320ac1aafd21b0070c17/check/src/kindcheck.rs#L248) to `Type` so there is no kind polymorphism.

check/src/unify_type.rs
```diff
@@ -205,63 +262,72 @@ fn find_alias<'a, 's, U>(unifier: &mut UnifierState<'a, 's, U>,
...
         };
     }
     Ok((l, r))
 }
 
+/// As a last ditch effort attempt to unify the types again by expanding the aliases (if the types
+/// are alias types).
 fn try_with_alias<'a, 's, U>(unifier: &mut UnifierState<'a, 's, U>,
                              expected: &TcType,
                              actual: &TcType)
```
bjz: Thanks for all these comments!

Marwes:
@brendanzab Unless you have some more questions or reservations I think this can be merged?

PR #62, opened 2016-07-23
Remove Type::Array variant

brendanzab:
Working towards #50


bjz:
Bump!

Marwes:
For some reason I thought you weren't done since it was so tiny compared to the previous removals!

Looks good, merge if it is indeed done!

Marwes:
Oops, sec noticed a mistake.

check/src/kindcheck.rs
```diff
@@ -139,14 +139,9 @@ impl<'a> KindCheck<'a> {
                 Ok((gen.kind.clone(), Type::generic(gen)))
             }
             Type::Variable(_) => panic!("kindcheck called on variable"),
+            Type::Builtin(BuiltinType::Array) |
```
Marwes: Array and function do not have the same kind (`Type -> Type` vs `Type -> Type -> Type`). If you can add a test to catch this as well that would be great!

bjz: Ah, durp

bjz:
Fixed the array kind issue, but still need to do some tests. Should there be a separate kindcheck test module?

Marwes:
There probably should be a separate kind check module. Currently the tests are just dumped into a single file (did split them into pass and fail in #61 though)

bjz:
arrrg skeptic arrg!

Marwes:
It might be that travis_wait is messing it up, gonna run some builds on travis in #66

Marwes:
@bjz Should this be merged? Or are you working on adding a kindcheck test?

brendanzab:
Yeah - would be nice to merge it - won't have much time to do stuff on gluon this week. Maybe we should make an issue to add a test...

Marwes:
Ok, I think its fine to merge it. Opening an issue for adding kindcheck tests would be a good idea.

PR #64, opened 2016-07-24
Update dependencies and add workspace field

bjz:
I find the map version of specifying dependencies easier to scan and update. Not sure if you agree!

I've also added a workspace so that if you do a build in a sub-directory it will reuse the root target directory.

Rustyline has been pegged to a specific git revision so as to ensure repeatable builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,15 +45,16 @@ path = "vm"
 version = "0.1.2"
 
 [dependencies.compiletest_rs]
-version = "0.1.3"
+version = "0.2"
 optional = true
 
 [dependencies.env_logger]
-version = "0.3.3"
+version = "0.3.4"
 optional = true
 
 [dependencies.rustyline]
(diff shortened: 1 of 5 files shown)


bjz:
ugh - something is weird with skeptic >_<

bjz:
Gonna roll back the workspace change to see if it fixes it...

Marwes:
I have had that error occur to me locally actually though in that case it occured after changing rust versions. It might have something to do with the caching travis does

bjz:
Would you like me to re-add it?

Marwes:
Maybe leave it out and submit that as a separate PR. If I am wrong and clearing out the cache does not solve the problem it may be a good idea to test it a bit to see if it is reproducible in which case a bug report to cargo may be needed.

bjz:
Sounds good - I had that error occur locally when I updated to the latest nightly too... maybe that is it.

Issue #33, opened 2016-07-16
Allow nested block comments


lfairy:
Nested block comments (e.g. `/* /* foo */ */`) are my favorite feature from Rust. It would be nice if Gluon supports them too.

Marwes:
Seems like a good idea to me. I can't see any reason not to not have nested comments so this might be best to implement directly in https://github.com/Marwes/combine-language.

bjz:
Have you thought of switching to something like LALRPOP? Or rustpeg?

Marwes:
It has been useful to dogfood `combine` but yes I have thought that the language has outgrown `combine`. It is relatively rare that I modify the parser though so it hasn't been much of a problem in practice.

To use any other parser library though they must support a custom lexer as it is rather involved (it uses combine as which should probably be changed regardless of which parser that is used as it is quite slow).

Also, I should add this to the README as well, but I really recommend to use either nightly or beta to compile gluon as they have massive performance improvements, just take a look at the travis builds https://travis-ci.org/Marwes/gluon.

bjz:
Hah, sorry! I didn't realise it was your lib! ^_^

I think the nightly builds are fast because they keep failing on `regex` ;)

Marwes:
"The fastest code is the one not ~~run~~ usable"

PR #66, opened 2016-07-24
chore: Remove travis_wait

Marwes:
This is just a hunch but it might be that travis_wait messes up something when it writes to stdout which causes skeptic to fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,9 @@ before_script:
   export PATH=$HOME/.local/bin:$PATH
 script:
 - |
-  travis_wait 40 sh ./test.sh &&
+  sh ./test.sh &&
   travis-cargo --only nightly test -- --features "test nightly" -p gluon compile_test &&
-  travis_wait 40 cargo build --release &&
+  cargo build --release &&
   travis-cargo --only nightly bench &&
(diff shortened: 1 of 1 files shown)


Marwes:
Test gitter

Marwes:
Ran beta and nightly three times without failing so I am reasonably sure it is travis_wait that is messing it up. Unfortunately stable now times out instead! Not sure what the best way of dealing with this is now, might be a good idea to temporarily disable skeptic tests until Rust 1.11 releases in four weeks (after which the compile times fix reaches stable).

bjz:
We really need to tackle the compile times of `gluon_parser` at some stage... but good to know.

Marwes:
I'd be happy to have some discussion about rewriting the parser. You mentioned lalrpop and rustpeg in #33. Between those I think lalrpop is probably better as rustpeg requires nightly, not sure if there are any other viable candidates. As a start the lexer could be rewritten, I think that could give a pretty large run speed and compile speed increase relative to the effort of rewriting.

bjz:
rustpeg doesn't require nightly if you use a build script ;)

bjz:
I can take up the parser if you like

PR #67, opened 2016-07-24
docs: Document the Instruction enum

Marwes:
In addition to adding doc comments I also removed the unused `GetIndex` instruction and made the two argument variants into struct variants to avoid ambiguity.

### DIFF
--- a/vm/src/types.rs
+++ b/vm/src/types.rs
@@ -9,36 +9,81 @@ pub type VMIndex = u32;
 pub type VMTag = u32;
 pub type VMInt = isize;
 
+/// Enum which represent the instructions executed by the virtual machine.
+///
+/// The binary arithmetic instructions pop two values of the stack and then push the result.
 #[derive(Copy, Clone, Debug)]
 pub enum Instruction {
+    /// Push an integer to the stack
     PushInt(isize),
+    /// Push a byte to the stack
     PushByte(u8),
+    /// Push a float to the stack
     PushFloat(f64),
+    /// Push a string to the stack by loading the string at `index` in the currently executing
+    /// function
     PushString(VMIndex),
+    /// Push a variable to the stack by loading the upvariable at `index` from the currently
+    /// executing function
+    PushUpVar(VMIndex),
+    /// Push the value at `index`
     Push(VMIndex),
+    /// Push the value at `index`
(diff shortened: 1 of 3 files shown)


vm/src/types.rs
```diff
@@ -9,36 +9,82 @@ pub type VMIndex = u32;
...
     PushString(VMIndex),
+    /// Push a variable to the stack by loading the upvariable at `index` from the currently
+    /// executing function
+    PushUpVar(VMIndex),
+    /// Push the value at `index`
     Push(VMIndex),
+    /// Push the value at `index`
     PushGlobal(VMIndex),
+    /// Call a function by passing it `args` number of arguments. The function is at the index in
+    /// the stack just before the arguments. After the call is done the function all arguments are
```
bjz: Remove 'the function' here - 'After the call is done all arguments are removed and the function is replaced by the result of the call.'

vm/src/types.rs
```diff
@@ -9,36 +9,82 @@ pub type VMIndex = u32;
 pub type VMTag = u32;
 pub type VMInt = isize;
 
+/// Enum which represent the instructions executed by the virtual machine.
+///
+/// The binary arithmetic instructions pop two values of the stack and pushes the result of
+/// adding/comparing/etc the values.
```
bjz: 'The binary arithmetic instructions pop two values of the stack and then push the result.'

vm/src/types.rs
```diff
@@ -9,36 +9,82 @@ pub type VMIndex = u32;
...
     Slide(VMIndex),
 
-    // Creates a closure with 'n' upvariables
-    // Pops the 'n' values on top of the stack and creates a closure
-    MakeClosure(VMIndex, VMIndex),
-    // Creates a closure but does not fill its environment
-    NewClosure(VMIndex, VMIndex),
-    // Fills the previously allocated closure with `n` upvariables
+    /// Creates a closure with the function at `function_index` of the currently executing function
+    /// and `upvars` upvariables poped from the top of the stack.
```
bjz: 'popped'

Marwes:
@bjz Fixed your other comments. Squashing the commits before merging if everything looks good.

bjz:
👍

PR #71, opened 2016-07-25
perf(check): Avoid recreating identical App instances in unroll_app

Marwes:
Gives a ~65% speedup on the typecheck_prelude benchmark


Marwes:
Currently re-profiling the benchmark to see if there are any other simple improvements.

PR #73, opened 2016-07-25
Fix a few error cases when using `:i`, `:k` and `:t` from the repl

Marwes:
Fixes a few things from https://github.com/Marwes/gluon/issues/58#issuecomment-234916434

### DIFF
--- a/vm/src/vm.rs
+++ b/vm/src/vm.rs
@@ -8,8 +8,8 @@ use std::string::String as StdString;
 use base::ast::{Typed, ASTType};
 use base::metadata::{Metadata, MetadataEnv};
 use base::symbol::{Name, Symbol, SymbolRef};
-use base::types;
-use base::types::{Type, KindEnv, TypeEnv, PrimitiveEnv, TcType, RcKind};
+use base::types::{Alias, AliasData, Generic, Type, Kind, KindEnv, TypeEnv, PrimitiveEnv, TcType,
+                  RcKind};
 
 use macros::MacroEnv;
 use {Error, Result};
@@ -145,7 +145,7 @@ impl TypeEnv for VMEnv {
                     .map(|ctor| ctor)
             })
     }
-    fn find_type_info(&self, id: &SymbolRef) -> Option<&types::Alias<Symbol, TcType>> {
+    fn find_type_info(&self, id: &SymbolRef) -> Option<&Alias<Symbol, TcType>> {
         self.type_infos
             .find_type_info(id)
     }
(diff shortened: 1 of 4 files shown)


vm/src/vm.rs
```diff
@@ -187,10 +187,15 @@ fn map_cow_option<T, U, F>(cow: Cow<T>, f: F) -> Option<Cow<U>>
...
-        }
         let name = Name::new(name);
+        let module_str = name.module().as_str();
+        if module_str == "" {
+            return self.type_infos
+                .id_to_type
+                .get(name.as_str())
+                .map(|alias| Cow::Borrowed(alias))
+                .ok_or_else(|| Error::UndefinedBinding(name.as_str().into()));
+        }
```
bjz: What does this do?

vm/src/vm.rs
```diff
@@ -326,19 +331,32 @@ impl GlobalVMState {
...
         use api::generic::A;
         use api::Generic;
+        fn add_type<T: Any>(ids: &mut HashMap<TypeId, TcType>,
+                            env: &mut VMEnv,
+                            name: &str,
+                            typ: TcType) {
+            ids.insert(TypeId::of::<T>(), typ);
+            // Insert aliases so that `find_info` can retrieve information about the primitives
+            env.type_infos.id_to_type.insert(name.into(),
+                                             types::Alias::from(types::AliasData {
```
bjz: Might look nicer if you imported these things from `types` in the top of the scope.

vm/src/vm.rs
```diff
@@ -326,19 +331,32 @@ impl GlobalVMState {
...
-            ids.insert(TypeId::of::<f64>(), Type::float());
-            ids.insert(TypeId::of::<::std::string::String>(), Type::string());
-            ids.insert(TypeId::of::<char>(), Type::char());
+            let ids = self.typeids.get_mut().unwrap();
+            let env = self.env.get_mut().unwrap();
+            add_type::<()>(ids, env, "()", Type::unit());
+            add_type::<VMInt>(ids, env, "Int", Type::int());
+            add_type::<u8>(ids, env, "Byte", Type::byte());
+            add_type::<f64>(ids, env, "Float", Type::float());
+            add_type::<::std::string::String>(ids, env, "String", Type::string());
```
bjz: Why the need for the full qualification of `String`?

vm/src/vm.rs
```diff
@@ -326,19 +331,32 @@ impl GlobalVMState {
...
+            add_type::<char>(ids, env, "Char", Type::char());
         }
         let _ = self.register_type::<IO<Generic<A>>>("IO", &["a"]);
         let _ = self.register_type::<Lazy<Generic<A>>>("Lazy", &["a"]);
```
bjz: Is this you hiding from the unused `Result` error? Why not `unwrap`?

vm/src/vm.rs
Marwes: type_infos contain globally defined types so if `module_str == ""` the input is just an identifier (`Int` not `std.prelude.Option`) and the only way to return a result is a global type. I should probably handle the case of a bare global variable though (`test.test` works but `test` wont)

bjz: I'm wondering if a raw `match` might be clearer here...

vm/src/vm.rs
Marwes: The `Value::String` variant is reexported and shadows `std::string::String`. Its a holdover from the time when enum variants existed in the same scope as the enum itself. Haven't gotten around to adding explicit `Value::` prefixes to all places where the enum is used yet.

Marwes:
Rebased and addressed all comments (except the qualification of String which would require more extensive changes).

Issue #23, opened 2016-04-21
Add an environment variable which specifies directories to look for imported files


Marwes:
The import.rs module would should be extended to look for an environment variable (maybe `GLUON_PATH`?) and use that for import lookup as well.

leshow:
I'd like to contribute something to your project, this issue seems easy enough. Do you have a particular way you might like to see this implemented?

I'm a beginner at Rust, so please let me know if this is way off base

https://github.com/leshow/gluon/blob/master/src/import.rs#L104

bjz:
Thanks for the interest!

I'm wondering if it would be possible to pass the path down from a higher level - ie. closer to the entrypoint. That way we can push the global state out to the edge of the code, making things more testable. Just an idea!

bjz:
Feel free to open a WIP PR, referencing this issue - that way we can talk over there.

Marwes:
I think it may be, as @bjz says that the addition of GLUON_PATH should be a bit higher up to keep the creation of an `Import` pure. I am thinking that it could be added in [new_vm](https://github.com/Marwes/gluon/blob/9058c253c23358cdfc249e88c5789bd3b4f6e98c/src/lib.rs#L543-L555) which serves to create a vm initialized with a sane set of defaults.

Also see https://github.com/Marwes/gluon/blob/9058c253c23358cdfc249e88c5789bd3b4f6e98c/tests/vm.rs#L39-L42 for how to retrieve and add a new path.

leshow:
awesome! I'll make some changes and open a PR

Marwes:
Closed by #79

brendanzab:
@leshow In the future, if you put 'Closes #N' of 'Fixes #N' in the PR description, the corresponding issue will be auto-closed. :)

PR #79, opened 2016-07-27
read GLUON_PATH from env var and add to new_vm

leshow:
@brendanzab 
I tried to do this as was said in the issue. I've exported a GLUON_PATH variable and it appears to get read while the vm is created. 

Feel free to completely reject this code if it's not up to snuff, as I said before I'm just a beginner with Rust.

In the interest of adding the path without any explicit conditionals, I added a method `insert_and_get` to `vm/src/macros` so that the function calls could be chained. If no path is found, it should just grab the current directory. 

Rebased off of master just prior.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,6 +22,7 @@ pub use vm::thread::{RootedThread, Thread};
 
 use std::result::Result as StdResult;
 use std::string::String as StdString;
+use std::env;
 
 use base::ast;
 use base::error::Errors;
@@ -538,12 +539,26 @@ pub fn filename_to_module(filename: &str) -> StdString {
     name.replace("/", ".")
 }
 
+fn get_env_path() -> Option<String> {
+    match env::var("GLUON_PATH") {
+        Ok(val) => Some(val),
+        Err(_) => None,
+    }
+}
+
 /// Creates a new virtual machine with support for importing other modules and with all primitives
 /// loaded.
 pub fn new_vm() -> RootedThread {
     let vm = RootedThread::new();
-    vm.get_macros().insert(String::from("import"),
-                           ::import::Import::new(::import::DefaultImporter));
+    let gluon_path = get_env_path().unwrap_or(String::from("."));
+    vm.get_macros()
+        .insert_and_get(String::from("import"),
+                        ::import::Import::new(::import::DefaultImporter))
+        .as_ref()
+        .and_then(|import| import.downcast_ref::<::import::Import>())
+        .expect("Import macro")
+        .add_path(gluon_path);
+
     Compiler::new()
         .implicit_prelude(false)
         .run_expr::<Generic<A>>(&vm, "", r#" import "std/types.glu" "#)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -539,25 +539,17 @@ pub fn filename_to_module(filename: &str) -> StdString {
     name.replace("/", ".")
 }
 
-fn get_env_path() -> Option<String> {
-    match env::var("GLUON_PATH") {
-        Ok(val) => Some(val),
-        Err(_) => None,
-    }
-}
-
 /// Creates a new virtual machine with support for importing other modules and with all primitives
 /// loaded.
 pub fn new_vm() -> RootedThread {
+    use ::import::{DefaultImporter, Import};
+
     let vm = RootedThread::new();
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -538,12 +539,26 @@ pub fn filename_to_module(filename: &str) -> StdString {
...
+    }
+}
+
 /// Creates a new virtual machine with support for importing other modules and with all primitives
 /// loaded.
 pub fn new_vm() -> RootedThread {
     let vm = RootedThread::new();
-    vm.get_macros().insert(String::from("import"),
-                           ::import::Import::new(::import::DefaultImporter));
+    let gluon_path = get_env_path().unwrap_or(String::from("."));
```
Marwes: No need for a separate function here. You can write this as `env::var("GLUON_PATH").unwrap_or(String::from("."))` as `Result` have the same [unwrap_or](https://doc.rust-lang.org/core/result/enum.Result.html#method.unwrap_or) method.

vm/src/macros.rs
```diff
@@ -57,6 +57,15 @@ impl MacroEnv {
         self.macros.write().unwrap().insert(name, Box::new(mac));
     }
 
+    /// Inserts a `Macro` then returns an instance of Option<Box<Macro>> using self.get
+    pub fn insert_and_get<M>(&self, name: String, mac: M) -> Option<Box<Macro>>
```
Marwes: No need to return an `Option` here since inserting the macro guarantees that it exists.

src/lib.rs
```diff
@@ -538,12 +539,26 @@ pub fn filename_to_module(filename: &str) -> StdString {
...
+    vm.get_macros()
+        .insert_and_get(String::from("import"),
+                        ::import::Import::new(::import::DefaultImporter))
+        .as_ref()
+        .and_then(|import| import.downcast_ref::<::import::Import>())
+        .expect("Import macro")
+        .add_path(gluon_path);
+
```
Marwes: If you want to clean up my old code a bit you could write it as. This avoids the need to downcast the macro. :smile:
```rust
let import = ::import::Import::new(::import::DefaultImporter);
import.add_path(gluon_path);
vm.get_macros().insert(String::from("import"), import);
```

vm/src/macros.rs
Marwes: If you make the above mentioned cleanup to inserting the path I guess this method may not be needed though.

Marwes:
Good work! Unfortunately it does not quite work as is due to the macro code being super weird. It uses `Box` and clones it when returning it from `.get`. Thus when `add_path` is called the path is added to a copy of the actual macro stored in the vm. This didn't matter for the [example](https://github.com/Marwes/gluon/blob/e4f6f66bdfa64c053bcdc7e7695cd9f41881a7e6/tests/vm.rs#L36-L44) that I linked to in #23 because all files actually imported in tests have been included directly in the binary (letting them always be accessible with `import "std/prelude.glu"`).

If you do the change in https://github.com/Marwes/gluon/pull/79#discussion_r72400370 it should work though as then the path is added before inserting the macro.

Fixing it so that paths can be added afterwards may be a bit complicated as the import macro currently interacts a bit subtly with the macro system itself. I think I will tackle that myself as it requires one to know a bit about how the compilation is set up.

src/lib.rs
brendanzab: Might also be cleaner to import the types at the top of the function. Generally in Rust the convention is to import types directly, and keep functions qualified by their parent module.

brendanzab: Also, I tend to start off importing things within a function if it is only needed there in  t,he entire file. I then bump it up to a module level import if I need it anywhere else. Just a personal preference though.

leshow:
I fear this has been a selfish endeavour and you all have spent more time explaining than probably would have taken to implement the feature yourself, lol. However, I have made the changes suggested anyway. Let me know if there's anything else I can do.

Marwes:
@leshow Don't worry about it, its just cool that you have taken the time to contribute to the project. I also find it fun just to help out!

This looks good in any case, will merge after travis finishes.

Marwes:
Merged! Thanks for the PR and don't be afraid of asking for help if you find something else that seems interesting!

brendanzab:
Don't worry! It's great to have help! :)

leshow:
Thanks guys!

PR #80, opened 2016-07-27
Fix a few error cases when using `:i`, `:k` and `:t` from the repl

Marwes:
Re-submitted PR of #73 since I apparently closed that instead of merging


brendanzab:
Was wondering about that!

PR #86, opened 2016-07-30
Clean up tests for check module

brendanzab:
- Moved common functions to a support module, as is common in most Rust projects.
- Prefer qualifying functions by module, and remove glob imports.
- Assign most expected values to `expected` binding before asserting equality.

### DIFF
--- a/check/tests/pass.rs
+++ b/check/tests/pass.rs
@@ -4,20 +4,19 @@ extern crate gluon_base as base;
 extern crate gluon_parser as parser;
 extern crate gluon_check as check;
 
-use base::ast;
-use base::ast::Typed;
-use base::types::{Generic, Kind, Type};
+use base::ast::{self, Expr, Pattern, Typed};
+use base::types::{Field, Generic, Kind, Type};
 use base::types;
 
-mod functions;
-use functions::*;
+mod support;
+use support::{alias, intern, typ};
 
 macro_rules! assert_pass {
     ($e: expr) => {{
         if !$e.is_ok() {
             panic!("assert_pass: {}", $e.unwrap_err());
         }
-    }}
+    }};
 }
 
 #[test]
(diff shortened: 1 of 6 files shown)


brendanzab:
Need to run rustfmt again.

brendanzab:
I'm also thinking it might be nice to move to using [expectest](https://crates.io/crates/expectest) - the matchers are a great deal nicer for testing.

check/tests/support/mod.rs
```diff
@@ -1,66 +1,71 @@
-extern crate env_logger;
+#![allow(private_in_public, dead_code)]
```
Marwes: Won't private_in_public be an error eventually? What is the reason for adding it?

check/tests/pass.rs
```diff
@@ -1,225 +1,265 @@
...
+
     let text = r"
 \x y -> 1 #Int+ x #Int+ y
 ";
-    let result = typecheck(text);
-    assert_eq!(result,
-               Ok(Type::function(vec![typ("Int"), typ("Int")], typ("Int"))));
+    let result = support::typecheck(text);
+    let expected = Ok(Type::function(vec![support::typ("Int"), support::typ("Int")],
+                                     support::typ("Int")));
```
Marwes: The type construction functions have intentionally short names as they are used everywhere. I don't think it helps readability  to qualify them.

Marwes: That includes alias and intern etc.

check/tests/support/mod.rs
brendanzab: I was getting errors with the reexports of the crates:

Eg.

```
warning: extern crate `gluon_base` is private, and cannot be reexported (error E0364), consider declaring with `pub`, #[warn(private_in_public)] on by default
 --> /Users/brendan/dev/rust/gluon/check/tests/support/mod.rs:7:9
  |
7 | pub use self::gluon_base as base;
  |         ^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see the explanation for E0446 (`--explain E0446`)
```

I couldn't put a `pub` on the `extern crate` though. I didn't see the deprecation notice in my haste though. I'll move the crate imports to the top instead.

Marwes:
> Moved common functions to a support module, as is common in most Rust projects.

> Prefer qualifying functions by module, and remove glob imports.

:+1:

> Assign most expected values to expected binding before asserting equality.

For large assertions this is good but for small ones I don't find it more readable. I don't feel strongly enough about it to have you change it either though. Any good reason for writing small asserts like that?

check/tests/pass.rs
brendanzab: Do you mind if I just do a hard import at the top of the module?

Marwes: Nope that is how it used to be (might have been a global though)

Marwes: *glob

Marwes: Clarification.  I don't mean that it should be a glob  now

brendanzab:
I was just trying to be consistent - I understand your reasoning though. I can inline them.

check/tests/pass.rs
brendanzab: Ok, done. That looks nicer ^_^

Marwes:
Consistency is a fair argument. Do it whichever way you feel best (taking into account changing it now)

brendanzab:
If we eventually go with expectest (which gives better errors), it will probably change again. I'll switch back now to reduce churn.

brendanzab:
Ugh, it's too hard to figure out which ones to change, and which ones to leave. I'm gonna stick to consistency.

check/tests/support/mod.rs
```diff
@@ -1,66 +1,64 @@
-extern crate env_logger;
+#![allow(dead_code)]
```
brendanzab: I think I'm going to revert this change - I don't like global allows in general...

PR #89, opened 2016-07-30
Print type along with value when when evaluating in the repl

### LINKED ISSUE #83: Print type along with value when when evaluating in the repl
Eg.

screen shot 2016-07-29 at 1 42 39 pm

brendanzab:
Closes #83 

Note that this changes `run_expr` and `run_io_expr` to return the inferred type along with the value. Not sure if that's ok :/

### DIFF
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -1,11 +1,9 @@
 extern crate env_logger;
 extern crate gluon;
 
-use gluon::vm::api;
-use gluon::vm::api::generic::A;
-use gluon::vm::api::{Generic, VMType, FunctionRef};
+use gluon::base::types::Type;
+use gluon::vm::api::{self, VMType, FunctionRef};
 use gluon::vm::thread::{RootedThread, Thread, Traverseable, Root, RootStr};
-use gluon::vm::internal::Value;
 use gluon::vm::types::VMInt;
 use gluon::Compiler;
 use gluon::import::Import;
@@ -13,13 +11,6 @@ use gluon::import::Import;
 fn load_script(vm: &Thread, filename: &str, input: &str) -> ::gluon::Result<()> {
     Compiler::new()
         .load_script(vm, filename, input)
-}
-
-fn run_expr(vm: &Thread, s: &str) -> Value {
-    Compiler::new()
-        .run_expr::<Generic<A>>(vm, "<top>", s)
-        .unwrap_or_else(|err| panic!("{}", err))
-        .0
 }
 
 fn make_vm() -> RootedThread {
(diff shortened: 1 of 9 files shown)


tests/api.rs
```diff
@@ -90,45 +83,52 @@ fn root_data() {
...
           let test: fn(_) -> _ = test;
           test
       })
       .unwrap();
-    let result = run_expr(&mut vm, expr);
-    match result {
-        Value::String(s) => assert_eq!(&s[..], "hello world"),
-        x => panic!("Expected string {:?}", x),
+
+    let (value, typ) = Compiler::new().run_expr::<Generic<A>>(&vm, "<top>", expr).unwrap();
```
Marwes: Since you changed this code you may also want to change `Generic <A>` to `String`.

Marwes:
I think it's fine to return both type and value, adding separate methods for only the value does not seem worth it.

tests/api.rs
brendanzab: I thought that par of the test was to match on the value stored in `Generic`? Same goes for the `array` test. If not I'll use the specific type though, seeing as it's cleaner!

Marwes: What is tested will end up being the same, the only way the current test succeeds is if `result == Value::String(..)` which is precisely the same thing as using `String` over `Generic<A>` as the only way to get a string out is if the resulting type where inferred to be a `String`.

That being said using `Generic<A>` to mean 'any value' is not really intuitive so that should probably be changed.

Marwes:
It might be possible to use the functions provided by [Executable](https://github.com/Marwes/gluon/blob/9207d325047d103ab34a62015a42e76823a53cda/src/lib.rs#L225-L298) directly instead to avoid changing the signatures of the `Compiler::*` methods. The `Executable` trait didn't end up exactly where I wanted it though so it may be a bit tricky to use (the `compiler_pipeline` module is intended to allow compiled programs to be compiled incrementally up to a specific stage without missing anything).

tests/api.rs
brendanzab: Yeah, it confused me at first too! ^_^

brendanzab:
Updated the tests, but haven't got to going back and looking at implementing it in a different way. Looks to be a bit more involved ;)

Marwes:
Fair enough, I have to take another stab at the `compiler_pipeline` module some time to make it simpler to use :grinning: . Merged

brendanzab:
Yay! Maybe we should make an issue?

Marwes:
Opened #93

PR #94, opened 2016-07-31
Remove Typed::type_of function

brendanzab:
Done as part of working towards #50.

`env_type_of` could probably be renamed, but I think it's ok for now.

I also removed some trait object indirection, and some unused impls for the env traits.

### DIFF
--- a/base/src/types.rs
+++ b/base/src/types.rs
@@ -14,13 +14,15 @@ use symbol::{Symbol, SymbolRef};
 pub type TcType = ast::ASTType<Symbol>;
 pub type TcIdent = ast::TcIdent<Symbol>;
 
+pub struct EmptyTypeEnv;
+
 /// Trait for values which contains kinded values which can be refered by name
 pub trait KindEnv {
     /// Returns the kind of the type `type_name`
     fn find_kind(&self, type_name: &SymbolRef) -> Option<RcKind>;
 }
 
-impl KindEnv for () {
+impl KindEnv for EmptyTypeEnv {
     fn find_kind(&self, _type_name: &SymbolRef) -> Option<RcKind> {
         None
     }
@@ -57,7 +59,7 @@ pub trait TypeEnv: KindEnv {
     fn find_record(&self, fields: &[Symbol]) -> Option<(&TcType, &TcType)>;
 }
 
-impl TypeEnv for () {
+impl TypeEnv for EmptyTypeEnv {
     fn find_type(&self, _id: &SymbolRef) -> Option<&TcType> {
         None
     }
--- a/base/src/types.rs
+++ b/base/src/types.rs
@@ -13,18 +13,10 @@ use symbol::{Symbol, SymbolRef};
 pub type TcType = ast::ASTType<Symbol>;
 pub type TcIdent = ast::TcIdent<Symbol>;
 
-pub struct EmptyTypeEnv;
-
 /// Trait for values which contains kinded values which can be refered by name
 pub trait KindEnv {
     /// Returns the kind of the type `type_name`
     fn find_kind(&self, type_name: &SymbolRef) -> Option<RcKind>;
-}
-
-impl KindEnv for EmptyTypeEnv {
-    fn find_kind(&self, _type_name: &SymbolRef) -> Option<RcKind> {
-        None
-    }
 }
 
 impl<'a, T: ?Sized + KindEnv> KindEnv for &'a T {
@@ -50,18 +42,6 @@ pub trait TypeEnv: KindEnv {
     /// Returns a record which contains all `fields`. The first element is the record type and the
     /// second is the alias type.
     fn find_record(&self, fields: &[Symbol]) -> Option<(&TcType, &TcType)>;
-}
-
-impl TypeEnv for EmptyTypeEnv {
-    fn find_type(&self, _id: &SymbolRef) -> Option<&TcType> {
-        None
-    }
-    fn find_type_info(&self, _id: &SymbolRef) -> Option<&Alias<Symbol, TcType>> {
-        None
-    }
-    fn find_record(&self, _fields: &[Symbol]) -> Option<(&TcType, &TcType)> {
-        None
-    }
 }
 
 impl<'a, T: ?Sized + TypeEnv> TypeEnv for &'a T {
--- a/base/src/types.rs
+++ b/base/src/types.rs
@@ -25,20 +25,14 @@ impl<'a, T: ?Sized + KindEnv> KindEnv for &'a T {
     }
 }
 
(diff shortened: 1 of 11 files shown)


base/src/types.rs
```diff
@@ -40,95 +27,25 @@ impl<T: KindEnv, U: KindEnv> KindEnv for (T, U) {
...
-    fn find_type(&self, id: &SymbolRef) -> Option<&TcType> {
-        (**self).find_type(id)
-    }
-    fn find_type_info(&self, id: &SymbolRef) -> Option<&Alias<Symbol, TcType>> {
-        (**self).find_type_info(id)
-    }
-    fn find_record(&self, fields: &[Symbol]) -> Option<(&TcType, &TcType)> {
-        (**self).find_record(fields)
-    }
-}
```
brendanzab: These kind of scream out at me as a warning sign... so much dereferencing! :)

Marwes: It was necessary for allowing `(&TypeEnv, Stack)` to work but that might not be necessary anymore

brendanzab: Seems like it is fine now ^_^

Marwes:
I think its best to refrain from removing the trait object indirection as there is quite a lot of code that is duplicated. It could be kept for `env_type_of` as that does not contain that much code but maybe it is still best to leave it as is, the trait object should not actually be used that much so any speed improvement should be negligible. Otherwise it looks good

Marwes:
For `OnFound` and `Suggest` there is nothing to revert to so you could keep those. Code duplication shouldn't be a problem there anyway as they are really only used in https://github.com/Marwes/gluon_language-server. They should maybe be moved there instead but then the language-server specific parts should maybe also be split off into a crate...

check/src/completion.rs
```diff
@@ -1,78 +1,81 @@
...
+
+    fn ident(&mut self, context: &LExpr<TcIdent<Symbol>>, ident: &TcIdent<Symbol>);
 }
 
-struct GetType(Option<TcType>);
-impl OnFound for GetType {
-    fn expr(&mut self, expr: &ast::LExpr<ast::TcIdent<Symbol>>) {
-        self.0 = Some(expr.type_of());
+struct GetType<'a, T: TypeEnv + 'a> {
+    env: &'a T,
```
Marwes: You could have `env: T` instead and remove the lifetime parameter if the implementation for `&T: TypeEnv` are kept. Then `GetType<&TypeEnv>` would work for instance.

Marwes: Specifying trait bounds in the struct definition is a bit of an anti-pattern (unless it is absolutely required because you need access to an associated type). It forces any user of the struct to specify it as well even if is not required in that instance (say if we wanted to implement ``PartialEq` manually for some reason, then ``T: TypeEnv` would need to be specified even if it was not needed).

brendanzab:
Pushed those fixes!

check/src/completion.rs
```diff
@@ -1,78 +1,81 @@
...
+struct Suggest<E> {
+    env: E,
     stack: ScopedMap<Symbol, TcType>,
-    result: Vec<ast::TcIdent<Symbol>>,
+    result: Vec<TcIdent<Symbol>>,
 }
-impl OnFound for Suggest {
-    fn on_ident(&mut self, ident: &ast::TcIdent<Symbol>) {
+
+impl<'a, E: TypeEnv> OnFound for Suggest<&'a E> {
```
Marwes: If you kept `impl<T: ?Sized + TypeEnv> for &'a T { .. }` you could have `impl<E: TypeEnv> OnFound for Suggest<E>`

vm/src/compiler.rs
```diff
@@ -226,38 +213,40 @@ impl CompilerEnv for TypeInfos {
     }
 }
 
-pub struct Compiler<'a> {
-    globals: &'a (CompilerEnv + 'a),
+pub struct Compiler<'a, E> {
```
Marwes: `Compiler` (just as `Typecheck`) are one of the places where two monomorphizations can actually mean a signficant amount of duplicated code. I think its better to keep the trait object here for that reason

Marwes:
Sorry for adding a few more nits, should have been clearer before :disappointed:

vm/src/compiler.rs
brendanzab: Ugh - now the vtables don't match... or something...

```
error[E0308]: mismatched types
   --> vm/src/compiler.rs:356:64
    |
356 |                                  TcType::from(expr.env_type_of(self.globals).clone()));
    |                                                                ^^^^^^^^^^^^ expected trait `base::types::TypeEnv`, found trait `compiler::CompilerEnv`
    |
    = note: expected type `&base::types::TypeEnv`
    = note:    found type `&'a compiler::CompilerEnv + 'a`
```

brendanzab: Guess I need to add those & impls again...

Marwes: Yeah, Rust does not downcast `&CompilerEnv` -> `&TypeEnv` automatically. An alternative solution may be to add an explict downcast method on `CompilerEnv`. 

```rust
// Could be a default method on CompilerEnv
fn as_type_env(&self) -> &TypeEnv { self }
```

brendanzab: Hmm, that gives an error about `self` not being `Sized`:

```
λ cargo run -- -i
   Compiling gluon_vm v0.1.2 (file:///Users/brendan/dev/rust/gluon/vm)
error[E0277]: the trait bound `Self: std::marker::Sized` is not satisfied
   --> vm/src/compiler.rs:184:9
    |
184 |         self
    |         ^^^^
    |
    = help: consider adding a `where Self: std::marker::Sized` bound
    = note: required for the cast to the object type `base::types::TypeEnv`
```

Marwes: I Guess it can't be a default method? A normal trait method should work.

Marwes: I guess one could implement `impl CompilerEnv for [T]` so `&self: &[T]` which cannot be casted to `&TypeEnv` (as that would require it to return `(v-table, size, data-ptr)` or something)

brendanzab: Eh, let's keep the weird &T impls for now. We should really do some profiling anyway before optimizing for stuff we don't need to. It would be nice to have cut out that extra code though...

Marwes: Yeah they unfortunately make it rather easy to accidentally add extra indirections. On the other hand they are quite crucial when it comes to doing composable generic APIs such as iterators or parser combinators (as in combine which is perhaps why I don't mind them) .

Marwes:
test

Issue #77, opened 2016-07-26
Type signatures causes variables to be generalized to early


Marwes:
```f#
let make x =
    let y: m = x
    x

make
```
Gives a type of `m00 -> m01` whereas it should be `m00 -> m00`. This is caused by the type signature of `y` which unifies the type of `x` with the 'generic' variable `m` in the binding. Subsequent use of `x` instantiates `m` with a fresh variable which gives the resulting type where the variables are not equal.

Marwes:
I think it is best to forbid this as it tries match a type signature to a variable which comes from a scope above the type declaration. Forbidding it outright may be a bit restricting however so there are a few things that can be done to allow something close to this to be written.

Make type variables scoped so that a type variable from an outer scope can be used in an inner scope.
```f#
let make x: m -> m =
    // `m` here is the same as `m` above
    let y: m = x
    x

make
```
In some cases the type signature of an outer scope may be so large that it is inconvenient to type it out (which could be the reason for the inner signature). A second improvment taken from the [1ML paper](http://www.mpi-sws.org/~rossberg/1ml/) could be to add `_` to signatures which means "I don't know what type this is, please fill it in for me"

```f#
// `_` is filled in with `m` after inference
let make x: m -> _ =
    let y: m = x
    x

make
```

We could however allow the example to be written as is but I am less sure how the semantics should be in that case. Probably variables from an outer scope that are matched to a variable in a signature needs to just ignore that the match?

PR #96, opened 2016-08-01
refactor: Document and cleanup unification

Marwes:


Marwes:
Nothing to exiting here, just some cleanup to make #77 easier and a few comments added.

check/src/typecheck.rs
```diff
@@ -1235,8 +1237,8 @@ impl<'a> Typecheck<'a> {
         debug!("Unify {} <=> {}",
                types::display_type(&self.symbols, expected),
                types::display_type(&self.symbols, &actual));
-        let mut state = ::unify_type::State::new(&self.environment);
-        match unify::unify(&self.subs, &mut state, expected, &actual) {
+        let state = ::unify_type::State::new(&self.environment);
```
brendanzab: Maybe import this as `UnifyState` or something?

check/src/unify.rs
```diff
@@ -31,122 +31,133 @@ impl<T, E> fmt::Display for Error<T, E>
...
-        let mut errors = &mut unifier.unifier.errors;
+    fn try_match(unifier: &mut UnifierState<S, Self>, l: &T, r: &T) -> Option<T> {
+        let subs = unifier.unifier.subs;
+        // Retrieve the 'real' types by resolving
         let l = subs.real(l);
         let r = subs.real(r);
-        match (l.get_var(), r.get_var()) {
-            (Some(l), Some(r)) if l.get_id() == r.get_id() => None,
+        // `l` and `r` must have the same type, if one is a variable that variable is
+        // 'unioned' with whatever the other type is
```
brendanzab: unioned -> unified?

check/src/typecheck.rs
Marwes: There were a bunch of other places with `::unify_type::*` so I added `use unify_type;` at the top instead. `unify_type` is not an ideal name but just using `unify` may confuse it with the `unify` module itself.

check/src/unify.rs
Marwes: Fixed

PR #97, opened 2016-08-01
chore: Temporarily disable stable Rust on travis to avoid timeouts

Marwes:


brendanzab:
Sad, but oh well... :(

Marwes:
:/ At least it is only for 3 more weeks

brendanzab:
Hah, true that!

Issue #98, opened 2016-08-01
Syntax for specifying and ascribing implicit arguments


brendanzab:
At the moment we have one form of implicit argument - type variables. For example:

``` elm
let compose : cat b c -> cat a b -> cat a c = ...
```

Has the implicit arguments:
- `cat : Type -> Type`
- `a : Type`
- `b : Type`

At the moment there is no way to explicitly specify the types of those arguments, or to explicitly ascribe them when mentioning the binding.

As an example, Idris has the following syntax for it:

``` idris
compose : {cat : Type -> Type} -> {b : Type} -> cat b c -> cat a b -> cat a c

foo = compose {a = Int} {cat = (->)}
```

Note in particular that:
- For better or for worse (ask @puffnfresh), variables in lower case are assumed to be implicit args of `Type` - the explicit definition of `b` is just for illustration.
- The brace syntax conflicts with our record syntax.
- This also requires later arguments to depend on previous arguments - perhaps only types would be allowed initially. I think this feature also subsumes [rank-n types](https://wiki.haskell.org/Rank-N_types) - not sure though.

Whatever syntax is chosen, this could eventually be used for modular implicits (see #47). But the restrictions to only using types might have to be relaxed somewhat. I think the [1ML paper](https://www.mpi-sws.org/~rossberg/1ml/1ml-extended.pdf) has some ideas on how to restrict it while retaining decidability (see the section, 'Predicativity').

I'm thinking I might have to start working on a spec for our inference rules before we tackle all this stuff though... I tried using a combination of records and implicit arguments to implement modular implicits, but it couldn't infer the module (EDIT: that was in Idris). :/

Marwes:
> ... variables in lower case are assumed to be implicit args of Type ...

Is this in Idris? In Haskell (and gluon atm) variables default to `Type`/`*` if they are unbound after inference.

>  .... I think this feature also subsumes rank-n types - not sure though.

Haskell has `forall` for this so `(forall a. a ->a) -> () == ({ a: Type } -> a -> a) -> ()`?

> Whatever syntax is chosen ...

Just tossing out ideas.

OCaml has `?` for default arguments so `compose : (cat : Type -> Type)? -> (b : Type)? -> cat b c -> cat a b -> cat a c` may have some precedence. 

Or `compose : [at : Type -> Type] -> [b : Type] -> cat b c -> cat a b -> cat a c` ? Would conflict with array patterns if those were added though.

[...]

PR #101, opened 2016-08-05
fix(check): Don't generalize variables from an outer scope ...

### LINKED ISSUE #77: Type signatures causes variables to be generalized to early

Marwes:
... when merging with a type signature

This commit tracks all type variables introduced in type signatures and uses this information to emit an error when a variable from a signature is matched to a variable from an outer scope.

Fixes #77

### DIFF
--- a/check/src/substitution.rs
+++ b/check/src/substitution.rs
@@ -294,6 +294,10 @@ impl<T: Substitutable + PartialEq + Clone> Substitution<T> {
     pub fn union(&self, id: &T::Variable, typ: &T) -> Result<(), ()>
         where T::Variable: Clone
     {
+        // Nothing needs to be done if both are the same variable already (also prevents the occurs check from failing)
+        if typ.get_var().map_or(false, |other| other.get_id() == id.get_id()) {
+            return Ok(());
+        }
         if occurs(typ, self, id) {
             return Err(());
         }
--- a/check/src/unify_type.rs
+++ b/check/src/unify_type.rs
@@ -1,13 +1,15 @@
 use std::fmt;
 
 use base::ast::AstType;
+use base::error::Errors;
 use base::types::{self, TcType, Type, TypeVariable, TypeEnv, merge};
 use base::symbol::{Symbol, SymbolRef};
 use base::instantiate;
+use base::scoped_map::ScopedMap;
 
 use unify;
 use unify::{Error as UnifyError, Unifier, Unifiable};
-use substitution::{Variable, Substitutable};
+use substitution::{Variable, Substitutable, Substitution};
 
 pub type Error<I> = UnifyError<AstType<I>, TypeError<I>>;
 
@@ -32,6 +34,7 @@ pub enum TypeError<I> {
     UndefinedType(I),
     FieldMismatch(I, I),
     SelfRecursive(I),
+    UnableToGeneralize(I),
 }
 
 impl From<instantiate::Error> for Error<Symbol> {
@@ -58,6 +61,12 @@ impl<I> fmt::Display for TypeError<I>
             TypeError::SelfRecursive(ref id) => {
                 write!(f,
                        "The use of self recursion in type `{}` could not be unified.",
+                       id)
+            }
+            TypeError::UnableToGeneralize(ref id) => {
+                write!(f,
+                       "Could not generalize the variable bound to `{}` as the variable was used \
+                        outside its scope",
                        id)
             }
         }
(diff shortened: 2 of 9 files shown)


Marwes:
This still fails to compile the prelude as there are multiple places where a type signature is unified with a variable from a scope outside the type signature. Would like some feedback for how to fix this, see #77 for some ideas.

check/tests/fail.rs
```diff
@@ -294,3 +294,17 @@ let g x: A a -> () = x
...
+    let text = r#"
+let make m =
+    let m2: m = m
+    m
+
+make
+"#;
+    let result = support::typecheck(text);
+    assert!(result.is_err());
+}
```
brendanzab: Missing new line

check/tests/pass.rs
```diff
@@ -827,3 +826,18 @@ in r1"#;
...
+let any x = any x
+let make m: m -> { test: m, test2: m } =
+    let m2: m = any ()
+    { test = m, test2 = m2 }
+
+make
+"#;
+    let result = support::typecheck(text);
+    assert!(result.is_ok(), "{}", result.unwrap_err());
+}
```
brendanzab: Missing new line

brendanzab:
```
Line: 69, Column: 9: Expected the following types to be equal
Expected: m -> m -> m
Found: 88 -> 88 -> 88
3 errors were found during unification:
Could not generalize the variable bound to `m` as the variable was used outside its scope
Could not generalize the variable bound to `m` as the variable was used outside its scope
Could not generalize the variable bound to `m` as the variable was used outside its scope
    let (<>) : m -> m -> m = append
        ^~
```

Well, now that I think about it, the current definition of `make_Monoid` is quite odd:

```
let make_Monoid m =
    let { append, empty } = m

    let (<>) : m -> m -> m = append

    { append, empty, (<>) }
```

What `make_Monoid` should *really* look like is this ('scuse the straw man syntax for implicits #98):

```
let make_Monoid [t] M: [t : Type] -> Monoid t -> { Monoid t | (<>) : t -> t -> t }  =
    let { append, empty } = M

    let (<>) : t -> t -> t = append

    { append, empty, (<>) }
```

Since we have no current way of bringing the `t` binding into scope, we shouldn't be able to explicitly type `(<>)`. I'm not sure if this is the source of the errors we are getting, but there is certainly something unsound in our current typechecker that wasn't getting found.

Marwes:
Rebased. Fixed the nits and added some comments to the first commit, fixed a small mistake and commented out the failing type signatures in the other two commits

check/src/substitution.rs
```diff
@@ -294,6 +294,10 @@ impl<T: Substitutable + PartialEq + Clone> Substitution<T> {
     pub fn union(&self, id: &T::Variable, typ: &T) -> Result<(), ()>
         where T::Variable: Clone
     {
+        // Nothing needs to be done if both are the same variable already (also prevents the occurs check from failing)
```
brendanzab: > also prevents the occurs check from failing

What does this mean? Would `occurs` panic otherwise?

check/src/unify_type.rs
```diff
@@ -375,6 +384,109 @@ fn walk_move_types2<'a, I, F, T>(mut types: I, replaced: bool, output: &mut Vec<
...
+    subs: &'e Substitution<TcType>,
+    variables: &'e ScopedMap<Symbol, TcType>,
+    errors: Errors<Error<Symbol>>,
+    level: u32,
+}
+
+impl<'a, 'e> Unifier<State<'a>, TcType> for Merge<'e> {
+    fn report_error(unifier: &mut UnifierState<Self>,
+                    error: UnifyError<TcType, TypeError<Symbol>>) {
+        unifier.unifier.errors.error(error);
```
brendanzab: The stutteriness is strong with this one... :3

check/src/substitution.rs
Marwes: No it would fail Typecheck ingredients with `123 occurs in 123` which is obviously nonsense. Unifying with itself should always succeed

check/src/unify_type.rs
Marwes: And naming is hard :(

check/src/substitution.rs
brendanzab: Aha! Thanks!

PR #102, opened 2016-08-07
refactor(vm): Remove the need for the api::Userdata wrapper

Marwes:
By removing the blanket implementation on the `value::Userdata` trait. types can instead opt-in to use the `Userdata` representation in the vm. This means users only need to write `ìmpl Userdata for MyType { }` to allow them to define functions in gluon which create new values of `MyType` (`fn make_my_type(_: ()) -> MyType` works)

### DIFF
--- a/vm/src/api.rs
+++ b/vm/src/api.rs
@@ -3,7 +3,7 @@ use {Variants, Error};
 use gc::{DataDef, Gc, Traverseable, Move};
 use base::symbol::Symbol;
 use stack::{State, Stack, StackFrame};
-use vm::{Thread, Status, RootStr, RootedValue, Root};
+use vm::{self, Thread, Status, RootStr, RootedValue, Root};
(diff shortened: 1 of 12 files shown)


vm/src/api.rs
```diff
@@ -645,36 +656,14 @@ impl<'vm, T> Pushable<'vm> for Vec<T>
...
-    }
-}
-impl<'vm, T: Clone + ::vm::Userdata> Getable<'vm> for Userdata<T> {
-    fn from_value(_: &'vm Thread, value: Variants) -> Option<Userdata<T>> {
-        match value.as_ref() {
-            ValueRef::Userdata(data) => data.downcast_ref::<T>().map(|x| Userdata(x.clone())),
-            _ => None,
-        }
-    }
-}
```
Marwes: I left out this implementation as it requires a clone bound which could be unintentionally expensive.

Marwes:
Tests should be fixed for real this time (with a rather contrived solution).

PR #103, opened 2016-08-08
Rewrite parser using LALRPOP

### LINKED ISSUE #69: Improve compilation times of gluon_parser
This is causing timeout woes with travis and makes things hard for users on stable. As discussed on https://github.com/Marwes/gluon/issues/33#issuecomment-233157725 and https://github.com/Marwes/gluon/pull/66#issuecomment-234803447 we could possibly consider switching to a parser generator like rustpeg of LALRPOP.

brendanzab:
Fixes #69


brendanzab:
Very half done right now, but thought I'd post my progress anyway...

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,233 @@
+use std::iter;
```
brendanzab: This file as broken - was trying to see if I could adapt the one from `parser2` to work with the current lexer

brendanzab:
The most interesting part is probably the grammar definition: https://github.com/Marwes/gluon/pull/103/files#diff-1dbca9e85701699d5fce975a5912be7dR1

Still doesn't handle everything though.

parser/src/lexer.rs
```diff
@@ -65,8 +41,12 @@ pub enum Token<Id> {
...
     Then,
     Else,
-    Open(Delimiter),
-    Close(Delimiter),
+    OpenBrace,
+    CloseBrace,
+    OpenParen,
+    CloseParen,
+    OpenBracket,
+    CloseBracket,
```
Marwes: Why did you change these definitions?

parser2/src/grammar.lalrpop
```diff
@@ -0,0 +1,233 @@
...
+
+AtomicExpr: Expr = {
+    Ident =>
+        Expr::Ident(<>),
+
+    Literal =>
+        Expr::Literal(<>),
+
+    "(" "." <id: Ident> ")" =>
+        Expr::Getter(id),
```
Marwes: Is this intended to be sugar for `\x -> x.id`?

parser2/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+
+pub fn parse_literal(src: &str) -> Result<Literal, Error> {
+    grammar::parse_Literal(src, Tokenizer::new(src))
+}
+
+/// Arm of a match expression (see `Expr::Match`)
+#[derive(Clone, PartialEq, Debug)]
+pub struct MatchArm {
+    pattern: Pattern,
+    body: Box<Expr>,
```
Marwes: No need to use `Box<Expr>` since you already have `Vec<MatchArm>`

parser2/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+    pub fn tuple<Elems>(elems: Elems) -> Expr
+        where Elems: IntoIterator<Item = Expr>
+    {
+        Expr::Record(desugar_tuple(elems))
+    }
+}
+
+pub fn parse_expr(src: &str) -> Result<Expr, Error> {
+    grammar::parse_Expr(src, Tokenizer::new(src))
+}
```
Marwes: Why have you defined a new AST instead of using (and modifying the existing on)?

parser2/src/tokenize.rs
```diff
@@ -0,0 +1,585 @@
...
+    Then,
+    Type,
+    With,
+
+    Name(&'input str),
+    Operator(&'input str),
+    ByteLiteral(u8),
+    IntLiteral(i64),
+    FloatLiteral(f64),
+    StringLiteral(String),
```
Marwes: Might be a performance problem if `StringLiteral` gets cloned a lot and there is a large literal (not sure how LALRPOP handles it). The old parser does the same but since `&'input str` is already used maybe that should be used here as well? The processing of escapes could be done later on then.

parser2/src/tokenize.rs
```diff
@@ -0,0 +1,585 @@
...
+    RArrow,
+    RBrace,
+    RBracket,
+    RParen,
+
+    // Internal layout tokens
+    OpenBlock,
+    CloseBlock,
+    Separator,
+}
```
Marwes: No token for documentation comments (might be intentionally left out for now?)

parser2/src/tokenize.rs
```diff
@@ -0,0 +1,585 @@
...
+//         }
+
+//         Ok(())
+//     }
+// }
+
+pub struct Tokenizer<'input> {
+    src: &'input str,
+    chars: CharIndices<'input>,
+    lookahead: Option<(usize, char)>,
```
Marwes: Maybe use https://doc.rust-lang.org/std/iter/struct.Peekable.html instead ?

parser2/src/tokenize.rs
```diff
@@ -0,0 +1,585 @@
...
+fn is_operator_symbol(ch: char) -> bool {
+    "+-*/&|=<>:.".chars().any(|x| x == ch)
+}
+
+fn is_ident_start(ch: char) -> bool {
+    UnicodeXID::is_xid_start(ch)
+}
+
+fn is_ident_continue(ch: char) -> bool {
+    UnicodeXID::is_xid_continue(ch)
```
Marwes: Wasn't aware that unicode had specifications for identifiers, is this commonly used in other languages?

parser2/src/tokenize.rs
```diff
@@ -0,0 +1,585 @@
...
+    {
+        while let Some((i1, ch)) = self.lookahead {
+            if terminate(ch) {
+                return Some(i1);
+            } else {
+                self.bump();
+            }
+        }
+        None
+    }
```
Marwes: Is there a need for detecting the EOF case? If not it seems simpler to just return a `usize` with what is currently `None` returned as the length of the string.

parser2/src/grammar.lalrpop
```diff
@@ -0,0 +1,233 @@
...
+Comma<Rule>: Vec<Rule> =
+    <rules: (<Rule> ",")*> <last: Rule?> => {
+        let mut rules = rules;
+        rules.extend(last);
+        rules
+    };
+
+And<Rule>: Vec<Rule> =
+    <rules: (<Rule> "and")*> <last: Rule> =>
+        rules.into_iter().chain(iter::once(last)).collect();
```
Marwes: Avoid allocating a new `Vec` by pushing the last rule instead (as you do in the Comma macro).

parser2/src/grammar.lalrpop
```diff
@@ -0,0 +1,233 @@
...
+        Expr::IfElse(Box::new(pred), Box::new(if_true), Box::new(if_false)),
+
+    "match" <input: Expr> "with" <arms: MatchArm+> =>
+        Expr::Match(Box::new(input), arms),
+
+    "let" <bindings: And<ValueBinding>> "in" <body: Expr> =>
+        Expr::LetBindings(bindings, Box::new(body)),
+
+    "type" <bindings: And<TypeBinding>> "in" <body: Expr> =>
+        Expr::TypeBindings(bindings, Box::new(body)),
```
Marwes: Missing block expression

parser2/src/grammar.lalrpop
```diff
@@ -0,0 +1,233 @@
...
+};
+
+pub Expr: Expr =
+    "OpenBlock" <exprs: (<InnerExpr> "Separator")*> <last: InnerExpr?> "CloseBlock" =>
```
Marwes: Currently block expressions require at least one expression (just remove `?` on last?)

Marwes:
Left a bunch of comments, not all of them immediately important so feel free to ignore those for the moment. Looks like you are on the right track for the most part but if you want to do some refactorings on the current code then maybe split that into a separate commit (though it can be in the same PR), makes it a bit easier to review.

parser2/src/tokenize.rs
brendanzab: Gotta get onto that.

parser2/src/tokenize.rs
brendanzab: 👍

parser2/src/tokenize.rs
brendanzab: @nikomatsakis?

brendanzab:
Sorry for the massiveness - still pretty explaratory. Just wanted to check I wasn't strying too far off the deep-end. I'm thinking I might try to keep heading towards adapting the old lexer. What do you think? Is the rewrite a decent enough improvement to push forward, considering the pain of adapting the old lexer?

brendanzab:
I can start off by making PRs to the current data structures to make it easier to integrate it with LALRPOP, now that I know what is needed.

parser2/src/lib.rs
brendanzab: Was easier for experimentation in a separate crate. I'll probably move to amending the current AST.

Marwes:
> What do you think?

Looks good :+1:

> Is the rewrite a decent enough improvement to push forward, considering the pain of adapting the old lexer?

The only truly tricky part of the lexer is the layout function which I think should be kept (at least for the moment). If you want to rewrite the rest (and it looks like you have done maybe 80% of that?) that's great, I'd expect it to be a nice speed improvement as well.

parser2/src/tokenize.rs
brendanzab: Was just copying niko's stuff! ^_^

parser2/src/tokenize.rs
nikomatsakis: Yes, it does; I don't know how widespread it is in terms of usage though. I think many languages have their own tweaks.

brendanzab: I wonder if it might make more sense for us to specify our characters directly seeing as we never really generate rust code directly from our identifiers. I'm guessing the reason LALRPOP did it was for making it consistent for code gen purposes?

parser2/src/grammar.lalrpop
brendanzab: Yup. Is that ok? Elm uses it - I'm sure there are others. I can remove though - was experimenting!

parser2/src/tokenize.rs
brendanzab: @nikomatsakis: what kind of cloning does lalrpop do to tokens internally? Any?

parser2/src/grammar.lalrpop
Marwes: I think it's useful enough. Might want to add it in a follow up PR instead however

Marwes:
Something you may not have noticed is that the current parser have a bit of functionality built in to allow it to recover from errors and still produce an AST (even if its incomplete). It is currently only used to accept [empty field accesses](https://github.com/Marwes/gluon/blob/4a64c68d8d04a6788f1fea3d6f25471b873ee8e2/parser/src/lib.rs#L528-L537) like `record.` which is necessary for completion to work for that case.

Is it possible have that work with LALRPOP as well? Taking it more generally, is it possible to recover from errors in LALRPOP (for the purposes of detecting multiple errors and/or to do completion or other analysis on an incomplete AST)?

cc @nikomatsakis

brendanzab:
Not sure - I assumed that was why we had `Option<Result<...>>` at least for the tokeniser...

brendanzab:
Still haven't really run the parser yet to see what it produces.

Marwes:
`Option<Result<...>>` is just to detect EOF and is only in the lexer anyway, no? The errors I want to recover from is on the parser.

Marwes:
Looks like there is a concept of fallible rules in LALRPOP though it is currently undocumented https://github.com/nikomatsakis/lalrpop/issues/40.

Looks like [=>?](https://github.com/nikomatsakis/lalrpop/blob/324bbf0e9dec06d25c1b8b7dbf06ab16f48b26c7/lalrpop/src/parser/lrgrammar.lalrpop#L111) is the syntax for it but I am unsure how to actually use it. Hopefully Niko can respond and give an example.

brendanzab:
I think I'm going to try to get back to this one now!

Marwes:
Good luck! Hopefully there are no other refactoring sidetracks needed!

nikomatsakis:
Regarding `=>?` rules, I don't think they are quite what you are looking for: https://github.com/nikomatsakis/lalrpop/issues/40#issuecomment-244066767

brendanzab:
In order to do the interning in the parser I'm also figuring out a bug in lalrpop at nikomatsakis/lalrpop#142. Currently just using strings locally though :/

Marwes:
Managed to get the tests building as well as getting most of them to pass, still some tests to go though. Also, I think I forgot to fetch a new version from this branch before starting which caused some really messy merge commits 😅 . We probably want to rebase and rewrite the entire history of this PR anyway though.

```
failures:

---- associated_record stdout ----
        thread 'associated_record' panicked at '0: UnrecognizedToken { token: Some((BytePos(18), Comma, BytePos(19))), expected: [] }

', parser\tests\basic.rs:393

---- case_expr stdout ----
        thread 'case_expr' panicked at 'assertion failed: `(left == right)` (left: `Err((None, Errors { errors: [Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: ParseError { errors: [Message(Owned("UnrecognizedToken { token: Some((BytePos(39), Pipe, BytePos(40))), expected: [] }"))] } }] }))`, right: `Ok(Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Match(Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Ident(TypedIdent { typ: Hole, name: "None" }) }, [Alternative { pattern: Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Constructor(TypedIdent { typ: Hole, name: "Some" }, [TypedIdent { typ: Hole, name: "x" }]) }, expr: Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Ident(TypedIdent { typ: Hole, name: "x" }) } }, Alternative { pattern: Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Constructor(TypedIdent { typ: Hole, name: "None" }, []) }, expr: Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Literal(Int(0)) } }]) })`)', parser\tests\basic.rs:320

---- partial_field_access stdout ----
        thread 'partial_field_access' panicked at 'assertion failed: `(left == right)` (left: `None`, right: `Some(Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Projection(Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Ident(TypedIdent { typ: Hole, name: "test" }) }, "", Hole) })`)', parser\tests\basic.rs:614

---- operator_expr stdout ----
        thread 'operator_expr' panicked at 'assertion failed: `(left == right)` (left: `Spanned { span: Span { start: BytePos(0), end: BytePos(24) }, value: Infix(Spanned { span: Span { start: BytePos(0), end: BytePos(13) }, value: Infix(Spanned { span: Span { start: BytePos(0), end: BytePos(8) }, value: Infix(Spanned { span: Span { start: BytePos(0), end: BytePos(4) }, value: Ident(TypedIdent { typ: Hole, name: "test" }) }, TypedIdent { typ: Hole, name: "+" }, Spanned { span: Span { start: BytePos(7), end: BytePos(8) }, value: Literal(Int(1)) }) }, TypedIdent { typ: Hole, name: "*" }, Spanned { span: Span { start: BytePos(11), end: BytePos(13) }, value: Literal(Int(23)) }) }, TypedIdent { typ: Hole, name: "#Int-" }, Spanned { span: Span { start: BytePos(20), end: BytePos(24) }, value: Ident(TypedIdent { typ: Hole, name: "test" }) }) }`, right: `Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Infix(Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Infix(Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Ident(TypedIdent { typ: Hole, name: "test" }) }, TypedIdent { typ: Hole, name: "+" }, Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Infix(Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Literal(Int(1)) }, TypedIdent { typ: Hole, name: "*" }, Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Literal(Int(23)) }) }) }, TypedIdent { typ: Hole, name: "#Int-" }, Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Ident(TypedIdent { typ: Hole, name: "test" }) }) }`)', parser\tests\basic.rs:338

---- partial_field_access_in_block stdout ----
        thread 'partial_field_access_in_block' panicked at 'assertion failed: `(left == right)` (left: `None`, right: `Some(Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Block([Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Projection(Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Ident(TypedIdent { typ: Hole, name: "test" }) }, "", Hole) }, Spanned { span: Span { start: BytePos(0), end: BytePos(0) }, value: Ident(TypedIdent { typ: Hole, name: "test" }) }]) })`)', parser\tests\basic.rs:633

---- span_if_else stdout ----
        thread 'span_if_else' panicked at 'assertion failed: `(left == right)` (left: `Span { start: BytePos(1), end: BytePos(0) }`, right: `Span { start: BytePos(1), end: BytePos(35) }`)', parser\tests\basic.rs:487

---- span_match stdout ----
        thread 'span_match' panicked at '0: UnrecognizedToken { token: Some((BytePos(42), Pipe, BytePos(43))), expected: [] }

', parser\tests\basic.rs:465


failures:
    associated_record
    case_expr
    operator_expr
    partial_field_access
    partial_field_access_in_block
    span_if_else
    span_match
```

# Failures

- associated_record

Type fields are implemented with a `type` prefix instead of taking casing into account when deciding on a type or value field. Using a `type` prefix may actually be a good idea but I left this test alone for now. Thoughts?

- case_expr
Got the [LR1 not installed](https://github.com/nikomatsakis/lalrpop/issues/145) error when using multiple alternatives which I don't believe should be ambigous. Might be https://github.com/nikomatsakis/lalrpop/issues/144. Fixes seems to be heading to crates.io https://github.com/nikomatsakis/lalrpop/pull/151.

- operator_expr
Operator precedence is not taken into account

- partial_field_access
Did not implement "empty field access"

- partial_field_access_in_block
Ditto

- span_if_else
Haven't investigated this yet

- span_match
Ditto

That is it for tonight though, will probably continue tomorrow but feel free to jump on any of these failures if you like @brendanzab.

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,424 @@
...
+                                params.into_iter()
+                                    .map(|id| {
+                                        Generic {
+                                            kind: Kind::variable(0),
+                                            id: id,
+                                        }
+                                    })
+                                    .collect(),
+                                body),
+        }
```
brendanzab: WOAha! You really like making things dense! I'll try spacing some of these things out and making a separate pattern for the different return types ;)

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,424 @@
...
+            name: id.clone(),
+            alias: Alias::new(id,
+                                params.into_iter()
+                                    .map(|id| {
+                                        Generic {
+                                            kind: Kind::variable(0),
+                                            id: id,
+                                        }
+                                    })
+                                    .collect(),
```
brendanzab: If you split this into an args variable, it would look much happier. Eg:

```rust
    <id: Ident> <params: Ident*> "=" <body: Type> => {
        let args = params.into_iter()
                         .map(|id| Generic { kind: Kind::variable(0), id: id })
                         .collect(),

        TypeBinding {
            comment: None,
            name: id.clone(),
            alias: Alias::new(id, args, body),
        }
    }
```

parser/src/grammar.lalrpop
Marwes: I guess I don't really mind this sort of code when the complex code is just a parameter to a constructor (which has no complexity in itself). rustfmt lets me get away with this complexity as well to some extent 😛 . Will keep it in mind in any case (though this was just copy pasted from the current parser so I didn't even consider it).

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,409 @@
...
+        Expr::Record {
+            typ: Type::hole(),
+            types: types,
+            exprs: values,
+        }
+    },
+
+    "OpenBlock" <exprs: (<S<Expr>> "Semi")*> <last: S<Expr>> "CloseBlock" => {
+        if exprs.is_empty() {
+            last.value
```
Marwes: I think this may give the wrong spans in some cases as it will give a span between `OpenBlock` and `CloseBlock` instead of the actual expression (didn't think about that when I wrote it.

Marwes:
Looks like travis is having trouble with rustup :/

Marwes:
Add a few more fixes

parser/src/infix.rs
```diff
@@ -0,0 +1,305 @@
...
+                                        arg_stack.push(make_op(lhs, stack_op, rhs));
+                                    }
+                                    (Fixity::Right, Fixity::Right) => {
+                                        op_stack.push(stack_op);
+                                        op_stack.push(next_op);
+                                    }
+                                    // Conflicting fixities at the same precedence level -
+                                    // this is a user error!
+                                    (Fixity::Left, Fixity::Right) |
+                                    (Fixity::Right, Fixity::Left) => panic!(),
```
Marwes: Remember to remove this panic and return a result instead.

parser/src/infix.rs
```diff
@@ -0,0 +1,305 @@
...
+    fn resparse() {
+        //  1  + (2  * (6   -  8))
+        //  1  +  2  *  6   -  8
+        // (1  + (2  *  6)) -  8
+        let expr = op(int(1), "+", op(int(2), "*", op(int(6), "-", int(8))));
+        let result = Infixes::new(*expr, &MockEnv::new()).reparse();
+        let expected = *op(op(int(1), "+", op(int(2), "*", int(6))), "-", int(8));
+
+        assert_eq!(result, expected);
+    }
```
Marwes: Would be nice with a failing test (mixed precedence) as well.

parser/src/infix.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        assert_eq!(arg_stack.len(), 1);
+        assert_eq!(op_stack.len(), 0);
+
+        *arg_stack.pop().unwrap()
+    }
+}
+
+impl<'a, Id: 'a + Clone> Iterator for Infixes<'a, Id> {
```
Marwes: This `Clone` bound may not be necessary?

parser/src/infix.rs
```diff
@@ -0,0 +1,305 @@
...
+        self.remaining_expr.take().map(|expr| {
+            let expr = *expr; // Workaround for http://stackoverflow.com/questions/28466809/
+            match expr.value {
+                Expr::Infix(lhs, op, rhs) => {
+                    self.remaining_expr = Some(rhs);
+                    self.next_op = Some(op);
+                    InfixExpr::Arg(lhs)
+                }
+                _ => {
+                    assert!(self.next_op.is_none());
```
Marwes: Redundant assert (due to `next_op.take()` above)

parser/src/infix.rs
```diff
@@ -0,0 +1,305 @@
...
+                match self.next() {
+                    None => break,
+                    Some(Arg(next_expr)) => arg_stack.push(next_expr),
+                    Some(Op(next_op)) => {
+                        if op_stack.is_empty() {
+                            op_stack.push(next_op);
+                            continue;
+                        }
+
+                        let stack_op = op_stack.pop().unwrap();
```
Marwes: You could rewrite this and the above lines as `match op_stack.pop() { ... }`

parser/src/infix.rs
```diff
@@ -0,0 +1,305 @@
...
+
+                        match i32::cmp(&self.get_precedence(&next_op),
+                                       &self.get_precedence(&stack_op)) {
+                            Greater => {
+                                op_stack.push(stack_op);
+                                op_stack.push(next_op);
+                            }
+                            Less => {
+                                let lhs = arg_stack.pop().unwrap();
+                                let rhs = arg_stack.pop().unwrap();
```
Marwes: Could these `unwrap()`s ever fail? (probably not?)

Marwes:
I remember now that infix rewriting will require us to add an `Paren` variant to the `Expr` enum as parentheses are important during this pass.

brendanzab:
Why is that?

parser/src/infix.rs
brendanzab: Yeah - left over from having trouble with the `Expr::Infix(lhs, op, rhs)` destructure.

parser/src/infix.rs
brendanzab: Of course - I just was doing that because I wanted to get the tests to parse for now :(

Marwes:
Because without it how would you distinguish from `1 + 2 * 3` and `(1 + 2) * 3`?

parser/src/infix.rs
brendanzab: Oh, duhhh

parser/src/infix.rs
brendanzab: Not sure - was semi-copying [`Language.Haskell.Infix`](https://hackage.haskell.org/package/infix-0.1.1/docs/src/Language-Haskell-Infix.html), but will have to check. You are right to be concerned.

brendanzab:
Btw - this outputs the operators with a reversed grouping to what I was expecting. Ie. `8 - (1 + (2 * 6))` as opposed to `(1  + (2  *  6)) -  8`... I'm doing something wrong... 😒

brendanzab:
Squashed the commits into oblivion - sorry :(

brendanzab:
> Because without it how would you distinguish from `1 + 2 * 3` and `(1 + 2) * 3`?

You mean when a user explicitly adds parens?

brendanzab:
Could just reuse the block expr :)

Marwes:
> Btw - this outputs the operators with a reversed grouping to what I was expecting. Ie.

Guessing that is because parentheses are not reflected in the AST (?). Yeah, could just reuse `Block` for now, adding an extra constructor could always be added later on if that information is necessary.

parser/src/infix.rs
```diff
@@ -0,0 +1,413 @@
...
+                &self[name[1..].trim_left_matches(char::is_alphanumeric)]
+            } else {
+                &self.default
+            }
+        })
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum Error {
```
Marwes: :+1: Needs a `std::fmt::Display` and maybe `std::error::Error` impl as well. Adding the operators involved in the conflict to the error would be nice as well.

parser/src/lib.rs
```diff
@@ -63,845 +55,48 @@ impl StreamOnce for StreamType {
...
-         many(as_trait(&mut binding(Token::And))),
-         token(Token::In).expected("`in` or an expression in the same column as the `let`"))
-            .map(|(b, bindings, _)| {
-                let mut bindings: Vec<_> = bindings;
-                bindings.insert(0, b);
-                Bindings::LetBindings(bindings)
-            })
-            .parse_stream(input)
-    }
+pub fn parse_expr(symbols: &mut IdentEnv<Ident = Symbol>,
```
Marwes: Should we apply the precedence inside `parse_expr` or as a separate second pass? Make sure the normal parser tests use it anyway to get a few more tests passing.

parser/src/infix.rs
brendanzab: Was going to use `quick_error!`, but haven't got a chance yet

Marwes: Not sure how well it works with type parameters which may be needed if you want to store the operators that conflicted in the error. Might be possible to work around it.

Marwes:
@brendanzab Any part of the remaining errors you want me to look at?

brendanzab:
Adding more tests to narrow down the failure - was going to do some more, but feel free to look at it if you like. Also need to test that the spans are being recreated correctly...

Was using this to make it easier.... 😧 

```
    fn print_expr(expr: &SpannedExpr<String>) {
        use base::ast::Literal;

        match expr.value {
            Expr::Infix(ref lhs, ref op, ref rhs) => {
                print!("(");
                print_expr(lhs);
                print!(" {} ", op.name);
                print_expr(rhs);
                print!(")");
            }
            Expr::Literal(Literal::Int(x)) => print!("{}", x),
            _ => panic!(),
        }
    }
```

Marwes:
Is it only that ends up wrong `8 - ((6 * 2) + 1)`? What about `1 + 2 + 3`?

parser/src/infix.rs
```diff
@@ -0,0 +1,420 @@
...
+        let span = pos::span(lhs.span.start, rhs.span.end);
+        Box::new(pos::spanned(span, Expr::Infix(lhs, op, rhs)))
+    };
+
+    let mut infixes = Infixes::new(expr);
+    let mut arg_stack = Vec::new();
+    let mut op_stack = Vec::new();
+
+    loop {
+        if arg_stack.len() == 2 && op_stack.len() == 1 {
```
brendanzab: Added more tests this morning - haven't pushed it yet.

brendanzab: Added those tests. At least we have a good bunch now. :P

parser/src/infix.rs
```diff
@@ -0,0 +1,451 @@
...
+                            InfixExpr::Op(ident("*")),
+                            InfixExpr::Arg(int(6)),
+                            InfixExpr::Op(ident("-")),
+                            InfixExpr::Arg(int(8))];
+
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn resparse_single() {
```
Marwes: ``s/resparse/reparse/` (and in the other tests).

brendanzab: 😆

parser/src/infix.rs
brendanzab: This is what I was porting btw: https://hackage.haskell.org/package/infix-0.1.1/docs/src/Language-Haskell-Infix.html

Marwes: Did you mean to link to something more specific? That is the same link I posted (copied from infix.rs).

brendanzab: Yeah that's what it was. Used your fix, with some changes. Thanks! It's the weird flipperoony of vectors to linked lists that was messing me up :P

brendanzab:
Still need to actually use the infix reparser

brendanzab:
Any chance you could look at getting the op parser to work recursively over the AST?

brendanzab:
The partial parsing is also going to be a pain... (see nikomatsakis/lalrpop#156)

Marwes:
> Any chance you could look at getting the op parser to work recursively over the AST?

Sure, there is already a visitor for expressions so just walking the tree and reparsing all operator expressions should work.

Marwes:
> The partial parsing is also going to be a pain... (see nikomatsakis/lalrpop#156)

Still need to look partial parsing properly, only been making small innards so far.

Marwes:
Added reparsing on all nested infix expressions. Haven't had to time to remove the unwrap on error yet but this is a start.

Marwes:
I figured out why the `if_else_span` test fails. The `else` part is parsed as a single element block where the `CloseBlock` part is one byte after the float expression. While we do unwrap the single element block the `if` expression uses the (now removed) `CloseBlock` token as the ending position.

Unsure how to fix this at the moment, maybe we should just allow the current behaviour?

```rust

#[test]
fn span_if_else() {
    let _ = ::env_logger::init();

    let e = parse_new!(r#"
if True then
    1
else
    123.45
"#);
    assert_eq!(e.span,
               Span {
                   start: BytePos::from(1),
                   end: BytePos::from(35),
               });
}
```
```
---- span_if_else stdout ----
        1 OpenBlock 3
1 If 3
4 Identifier("True") 8
9 Then 13
18 OpenBlock 19
18 Int(1) 19
20 CloseBlock 24
20 Else 24
Spanned { span: Span { start: BytePos(18), end: BytePos(19) }, value: Literal(Int(1)) }
29 OpenBlock 35
29 Float(123.45) 35
36 CloseBlock 36
0 CloseBlock 0
Spanned { span: Span { start: BytePos(29), end: BytePos(35) }, value: Literal(Float(123.45)) }
Spanned { span: Span { start: BytePos(1), end: BytePos(36) }, value: IfElse(Spanned { span: Span { start: BytePos(4), end: BytePos(8) }, value: Ident(TypedIdent { typ: Hole, name: "True" }) }, Spanned { span: Span { start: BytePos(18), end: BytePos(19) }, value: Literal(Int(1)) }, Spanned { span: Span { start: BytePos(29), end: BytePos(35) }, value: Literal(Float(123.45)) }) }
thread 'span_if_else' panicked at 'assertion failed: `(left == right)` (left: `Span { start: BytePos(1), end: BytePos(36) }`, right: `Span { start: BytePos(1), end: BytePos(35) }`)', parser\tests\basic.rs:486


failures:
    span_if_else

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured
```

parser/src/lexer.rs
```diff
@@ -871,36 +865,6 @@ fn scan_for_next_block<'input, 'a, I>(lexer: &mut Lexer<'input, I>,
...
-    }
-
-    fn position(&self) -> Self::Position {
-        self.next_token
-            .as_ref()
-            .or_else(|| self.unprocessed_tokens.last())
-            .map(|token| token.span)
-            .unwrap_or_else(|| self.end_span.unwrap())
-    }
-}
```
brendanzab: 😍

parser/src/lib.rs
```diff
@@ -29,6 +28,51 @@ mod grammar;
...
 mod lexer;
 
+/// Shrink hidden spans to fit the visible expressions and flatten singleton blocks.
+fn shrink_hidden_spans<Id>(mut expr: SpannedExpr<Id>) -> SpannedExpr<Id> {
+    match expr.value {
+        Expr::IfElse(_, _, ref last) |
+        Expr::LetBindings(_, ref last) |
+        Expr::TypeBindings(_, ref last) => {
+            let end = last.span.end;
+            drop(last);
```
Marwes: Pretty sure this and the other drops does not do anythingdrop does not do anything.

parser/src/lib.rs
```diff
@@ -29,6 +28,51 @@ mod grammar;
 mod infix;
 mod lexer;
 
+/// Shrink hidden spans to fit the visible expressions and flatten singleton blocks.
+fn shrink_hidden_spans<Id>(mut expr: SpannedExpr<Id>) -> SpannedExpr<Id> {
+    match expr.value {
```
Marwes: Could write this match as

```
expr.span.end = match expr.value {
     Expr::TypeBindings(_, ref last) => last.span.end,
     // return on the expressions that do not change end
     ... => return;
```

parser/src/grammar.lalrpop
```diff
@@ -0,0 +1,408 @@
...
+pub Literal: Literal = {
+    "StringLiteral" => Literal::String(<>),
+    "CharLiteral" => Literal::Char(<>),
+    "IntLiteral" => Literal::Int(<>),
+    "ByteLiteral" => Literal::Byte(<>),
+    "FloatLiteral" => Literal::Float(<>),
+};
+
+Alternative: Alternative<Id> =
+    "|" <pat: Sp<Pattern>> "->" <expr: Sp<AtomicExpr>> => {
```
Marwes: `s/Sp<AtomicExpr>/SpannedExpr` should fix the spans of match

Marwes: Wait, does that cause the ambiguity again maybe? (Havenät checked yet)

parser/src/lib.rs
brendanzab: Ah - leftovers from some other borrow checker funniness

brendanzab:
We are now down to:

```
failures:
    associated_record
    case_expr
    partial_field_access
    partial_field_access_in_block
    type_mutually_recursive
    variant_type
```

Marwes:
Tried to solve the partial_* tests but got ambiguities. Might not be able to handle that without extending LALRPOP with error recovery :/.

Marwes:
Is it only the tests requiring `test.` that fails now? (There are also some in `check/tests/completion.rs`)

brendanzab:
Sorry - not sure what you mean. Could you rephrase? I think the indentation stuff is a problem - causing issues with parsing later stuff too.

Marwes:
Referred to tests which needs the parser to produce an AST for `record.` (projection with a missing field). What indentation stuff is causing problems?

brendanzab:
If I `#[ignore]` the partial access ones I get:

```
---- wrong_indent_expression stdout ----
    1 OpenBlock 4
1 Let 4
5 Identifier("y") 6
7 Equals 8
13 OpenBlock 16
13 Let 16
17 Identifier("x") 18
19 Equals 20
21 OpenBlock 22
21 Int(1) 22
27 CloseBlock 28
27 In 28
27 Identifier("x") 28
32 CloseBlock 33
32 Int(2) 33
34 In 35
34 Identifier("y") 35
36 CloseBlock 36
thread 'wrong_indent_expression' panicked at 'assertion failed: `(left == right)` (left: `Ok(Spanned { span: Span { start: BytePos(1), end: BytePos(35) }, value: LetBindings([ValueBinding { comment: None, name: Spanned { span: Span { start: BytePos(5), end: BytePos(6) }, value: Ident(TypedIdent { typ: Hole, name: "y" }) }, typ: Hole, args: [], expr: Spanned { span: Span { start: BytePos(13), end: BytePos(33) }, value: App(Spanned { span: Span { start: BytePos(13), end: BytePos(33) }, value: Block([Spanned { span: Span { start: BytePos(13), end: BytePos(28) }, value: LetBindings([ValueBinding { comment: None, name: Spanned { span: Span { start: BytePos(17), end: BytePos(18) }, value: Ident(TypedIdent { typ: Hole, name: "x" }) }, typ: Hole, args: [], expr: Spanned { span: Span { start: BytePos(21), end: BytePos(22) }, value: Literal(Int(1)) } }], Spanned { span: Span { start: BytePos(27), end: BytePos(28) }, value: Ident(TypedIdent { typ: Hole, name: "x" }) }) }]) }, [Spanned { span: Span { start: BytePos(32), end: BytePos(33) }, value: Literal(Int(2)) }]) } }], Spanned { span: Span { start: BytePos(34), end: BytePos(35) }, value: Ident(TypedIdent { typ: Hole, name: "y" }) }) })`, right: `Err(Errors { errors: [Parser(Spanned { span: Span { start: BytePos(32), end: BytePos(32) }, value: ParseError { errors: [Unexpected(Borrowed("Int")), Expected(Borrowed("`in` or an expression in the same column as the `let`"))] } })] })`)', parser/tests/indentation.rs:77
```

Marwes:
Meant to take a look at `wrong_indent_expression` but got caught up with some other stuff  (https://github.com/Marwes/json-schema, https://github.com/Marwes/debugserver-types). Will try to take a look tomorrow,

parser/src/lexer.rs
```diff
@@ -31,7 +31,9 @@ impl<I> StreamOnce for LocatedStream<I>
                 self.location.bump(ch);
                 // HACK: The layout algorithm expects `1` indexing for columns -
                 // this could be altered in the future though
-                self.location.column += Column::from(1);
+                if self.location.column == Column::from(0) {
+                    self.location.column = Column::from(1);
+                }
```
Marwes: @brendanzab  I may have misled you earlier here, just setting the column to 1 is the correct way. Incrementing would get executed on every uncons causing the column to increase twice as fast as it should have 😓

brendanzab: HAH! Easy fix then?

parser/src/grammar.lalrpop
```diff
@@ -400,6 +394,12 @@ Expr: Expr<Id> = {
 
         Expr::TypeBindings(bindings, Box::new(body))
     },
+
+    "OpenBlock" <exprs: (<SpExpr> "Semi")*> <last: SpExpr> "CloseBlock" => {
+        let mut exprs = exprs;
+        exprs.push(last);
+        Expr::Block(exprs)
+    },
```
brendanzab: I thought that seeing as this is atomic (does not require parens) it makes more sense lower down. How was it breaking it?

parser/src/lexer.rs
Marwes: It didn't actually have any effect as the column is only used in the lexer and its only compared with other columns which have been double incremented.

parser/src/grammar.lalrpop
Marwes:
```
let y =
    let x = 1
    x
   2
y
```
Would look like
```
let x = 1
x
```
applied on `2` which is not what one would expect.

Perhaps there is a better way of solving this but I am not aware of it at this time

brendanzab:
....and the build passes at last! _wipes brow_

brendanzab:
I dunno - I'm alright with `Errors` for now!

Marwes:
You are really going all in with refacatoring the lexer, looks good to me in any case :+1:

parser/Cargo.toml
```diff
@@ -13,18 +13,16 @@ documentation = "https://docs.rs/gluon"
...
 quick-error = "1.0.0"
-lalrpop = { git ="https://github.com/Marwes/lalrpop", path = "lalrpop", branch = "symbol_error" }
-lalrpop-util = "0.12.0"
+lalrpop-util = "0.12.2"
 log = "0.3.6"
 gluon_base = { path = "../base", version = "0.2.2" }
 collect-mac = "0.1.0"
 
 [build-dependencies]
-lalrpop = { git ="https://github.com/Marwes/lalrpop", path = "lalrpop", branch = "symbol_error" }
```
Marwes: Before we merge maybe we should rebase any commits with the symbol_error branch to point to crates.io (for easier bisecting later).

parser/src/grammar.lalrpop
```diff
@@ -300,8 +305,8 @@ AtomicExpr: Expr<Id> = {
     <expr: Sp<AtomicExpr>> "." <id: Ident> =>
         Expr::Projection(Box::new(expr), id, Type::hole()),
 
-    <expr: Sp<AtomicExpr>> "." <err: error> => {
-        errors.push(err);
+    <expr: Sp<AtomicExpr>> "." <err: !> => {
+        errors.push(err.error); // Make use of `ErrorRecovery`'s `dropped_tokens` field?
```
Marwes: Change to FIXME or remove?

parser/src/layout.rs
```diff
@@ -0,0 +1,404 @@
...
+            tokens: tokens,
+            unprocessed_tokens: Vec::new(),
+            indent_levels: Contexts::new(),
+        }
+    }
+
+    fn next_token(&mut self) -> SpannedToken<'input> {
+        self.unprocessed_tokens.pop().unwrap_or_else(|| {
+            self.tokens.next().unwrap_or_else(|| {
+                // Blegh
```
Marwes: Guessing this could be removed if we return `Option<SpannedToken>` everywhere.

parser/src/lib.rs
```diff
@@ -161,10 +162,19 @@ pub fn parse_partial_expr<Id>(symbols: &mut IdentEnv<Ident = Id>,
                               -> Result<SpannedExpr<Id>, (Option<SpannedExpr<Id>>, ParseErrors)>
     where Id: Clone,
 {
-    let lexer = Lexer::new(input);
+    let tokenizer = Tokenizer::new(input).map(|token| token.unwrap()); // FIXME
```
Marwes: Fix FIXME?

parser/src/token.rs
```diff
@@ -0,0 +1,669 @@
...
+pub type SpannedToken<'input> = Spanned<Token<'input>, Location>;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Error {
+    pub location: Location,
+    pub code: ErrorCode,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ErrorCode {
```
Marwes: I think we could use `quick_error` for this type? I don't think it is used in the result of parse_expr yet either?

parser/Cargo.toml
brendanzab: Unfortunately those commits use `error` instead of `!` :(

brendanzab: I think I might be able to amend some though!

Marwes: Remember to point it at 0.12.3 as 0.12.2 was yanked.

parser/src/layout.rs
Marwes: Might not be a problem (other than being ugly) as the span should not be used for the EOF token anyway, add a comment on that maybe if its true.

parser/src/token.rs
```diff
@@ -1,108 +1,669 @@
...
+
+        let token = match op {
+            "." => Token::Dot,
+            ":" => Token::Colon,
+            "=" => Token::Equals,
+            "|" => Token::Pipe,
+            "->" => Token::RArrow,
+            "#" => {
+                // Is this too permissive?
+                self.take_while(start, is_ident_start);
```
Marwes: I believe this is the same logic as before. It is not ideal but it works #204 or something similiar to make this more solid would be nice however.

brendanzab:
Currently doing some work on the fixmes at the moment

parser/src/lib.rs
```diff
@@ -134,32 +139,91 @@ type ErrorEnv<'err, 'input> = &'err mut Errors<LalrpopError<'input>>;
...
+{
+    fn new(tokens: Tokens) -> Lexer<'input, Tokens> {
+        Lexer {
+            tokens: tokens,
+            layout: LayoutContext::new(),
+        }
+    }
+
+    fn next_token(&mut self) -> Option<Result<SpannedToken<'input>, Error>> {
+        match self.tokens.next() {
```
Marwes: I think this is the cause of the failure, all tokens in `unprocessed_tokens` should be consumed before asking for another token

parser/src/layout.rs
```diff
@@ -318,10 +313,12 @@ impl<'input, Tokens> Layout<'input, Tokens>
                     // ```
                     // `let x = 1 in {{ a; b }}` and not `{{ (let x = 1 in a) ; b }}`
                     let offside = Offside::new(location, Context::Block { emit_semi: false });
-                    self.indent_levels.push(offside)?;
+                    if let Err(err) = self.indent_levels.push(offside) {
```
Marwes: Might be cleaner to have `Result<Option<_>, _>` instead of `Option<Result<_, _>>` internally and convert it in `fn next`?

brendanzab: Yeah, will try that - kind of a pain not to have `?` for `Option`... @nikomatsakis - where are our monads! :D

parser/src/lib.rs
brendanzab: Updated in https://github.com/gluon-lang/gluon/pull/103/commits/7586a0ff4b07be8c5c17134cfd586e072f79afac to be clearer

brendanzab:
Would it be possible to merge this with the token unwrap? We could make an issue to fix it before the next release.

Marwes:
As long as no tests fails I think that's fine. Got some stuff that is
blocked by this PR as well.

PR #105, opened 2016-08-08
feat(vm): Allow the heap size on each thread to be limited

Marwes:
This gives a way to control how much memory each thread is allowed to allocate. Attempting to allocate more memory than allowed causes an error to be returned.

Of note is that since this is a per-thread limit it is easily circumvented by just allocating more threads, maybe there should be an atomic counter instead shared by all threads?


Marwes:
This is a rebased branch from some time ago so it may be a bit rough as I really only made sure it compiled and worked on master. Added as a PR to get some discussion on whether it is useful or not.

vm/src/channel.rs
```diff
@@ -130,18 +130,16 @@ fn resume(vm: &Thread) -> Status {
...
-                    value.push(vm, &mut stack.stack);
-                    Status::Ok
+                    value.status_push(vm, &mut stack.stack)
                 }
                 Err(err) => {
                     let fmt = format!("{}", err);
-                    let result = Value::String(vm.alloc(&stack.stack, &fmt[..]));
-                    stack.push(result);
+                    let result = Value::String(vm.alloc_ignore_limit(&fmt[..]));
+                    let _ = stack.push(result);
```
brendanzab: Why is it ok to ignore the result here?

vm/src/gc.rs
```diff
@@ -540,14 +551,35 @@ impl Gc {
...
+        if needed >= self.memory_limit {
+            return Err(Error::OutOfMemory {
+                limit: self.memory_limit,
+                needed: needed,
+            });
+        }
+        Ok(self.alloc_ignore_limit_(size, def))
+    }
+
+    pub fn alloc_ignore_limit<D>(&mut self, def: D) -> GcPtr<D::Value>
```
brendanzab: Why is this public?

vm/src/lazy.rs
```diff
@@ -106,7 +106,8 @@ fn lazy(vm: &Thread) -> Status {
         value: Mutex::new(Lazy_::Thunk(f)),
         _marker: PhantomData,
     });
-    lazy.push(vm, &mut stack.stack)
+    let _ = lazy.push(vm, &mut stack.stack);
```
brendanzab: Could you also justify the ignores here?

vm/src/gc.rs
Marwes: It is so error messages can be allocated even if the thread would otherwise be out of memory

vm/src/channel.rs
Marwes: Should be an `unwrap`, since it pushes a `Value` it should never fail

vm/src/lazy.rs
Marwes: Nope, it should be returning an error if push fails

Marwes:
Could perhaps have held off a bit before making a PR, some obvious mistakes there if I had spent a bit more time cleaning it up :sweat_smile:

tests/vm.rs
```diff
@@ -759,6 +760,28 @@ string.slice s 1 (string.length s)
...
+        Err(Error::Macro(errors)) => {
+            match *errors.errors[0].downcast_ref::<Error>().unwrap() {
+                Error::VM(VMError::OutOfMemory { limit: 10, .. }) => (),
+                _ => panic!("Unexpected error {:?}", errors.errors[0]),
+            }
+        }
+        Err(err) => panic!("Unexpected error `{:?}`", err),
+        Ok(_) => panic!("Expected an error"),
+    }
+}
```
brendanzab: Why do we need such complex matching in a test? Is it possible we could make this more fine-grained?

Marwes: Yeah, it was made a bit more bulletproof than it needs to. Ideally I think that macro errors should be collapsed so that vm errors have less possible representations.

brendanzab:
No worries! That is why we review!

Marwes:
Rebased so I could use #102. I should have fixed all the issues now.

brendanzab:
Requires a rebase

PR #106, opened 2016-08-09
use fnv instead of SipHasher for HashMaps. add type FnvMap

### LINKED ISSUE #95: Use a non-cryptographic hasher for HashMaps
We have a bunch of `HashMap`s scattered all around the codebase, all of which use the default cryptographic `SipHasher`. We should switch to using a non-cryptographic hasher, for example [fnv](https://crates.io/crates/fnv).

leshow:
@brendanzab  mentioned putting the FnvMap alias elsewhere, the map appears to be most used in base/src though. I'm open to any suggestions. 

All tests pass.

Fixes #95


base/src/types.rs
```diff
@@ -5,14 +5,19 @@ use std::ops::Deref;
...
 use ast;
 use ast::{AstType, DisplayEnv};
 use symbol::{Symbol, SymbolRef};
 
+use fnv::FnvHasher;
+
 pub type TcType = ast::AstType<Symbol>;
 pub type TcIdent = ast::TcIdent<Symbol>;
 
+pub type FnvMap<K, V> = HashMap<K, V, BuildHasherDefault<FnvHasher>>;
```
brendanzab: I wonder if we could have a `gluon_base::fnv` module, with:

```rust
extern crate fnv;

...

pub use self::fnv::FnvHasher;

pub type FnvMap<K, V> = HashMap<K, V, BuildHasherDefault<FnvHasher>>;
```

leshow:
Sure, I'll move it out of `types` and add to the PR

brendanzab:
Just so you understand my reasoning, the name `types` refers to the fact that these are the data structures used for gloun's typechecking.  The naming might be initially confusing though as many Rust libraries use the same name to contain their common data structures.

leshow:
Yep, I follow, I appreciate your guidance, thanks!

I've run the tests again following the refactor, everything seems fine.

base/src/fixed.rs
```diff
@@ -2,12 +2,12 @@
 //! are alive. This is done by storing each value in a stable memory location and preventing an
 //! earlier inserted value to be overwritten.
 use std::cell::{RefCell, Ref};
-use std::collections::HashMap;
 use std::fmt;
 use std::hash::Hash;
 use std::iter::{FromIterator, IntoIterator};
 use std::ops::{Index, IndexMut};
 
+use fnv::FnvMap;
```
brendanzab: Nit: Could you add a space between this and the doc comment?

base/src/fnv.rs
```diff
@@ -0,0 +1,8 @@
+extern crate fnv;
+
+use std::collections::HashMap;
+use std::hash::BuildHasherDefault;
+
+pub use self::fnv::FnvHasher;
+
+pub type FnvMap<K, V> = HashMap<K, V, BuildHasherDefault<FnvHasher>>;
```
brendanzab: Maybe a doc comment here would be useful. Something along the lines of:

```rust
Non-cryptographic hash map using an FNV hasher

We don't need cryptographic security in the compiler and typechecker, so for performance reasons we use a faster hashing algorithm than Rust's default `SipHasher`.
```

brendanzab:
In your PR description, can you add 'Fixes #95'? That will mean that the issue will be auto-closed when we merge.

You also might be interested in joining our gitter channel (if you didn't know about it already) where we have more real-time discussions: https://gitter.im/Marwes/gluon

Marwes:
> In your PR description, can you add 'Fixes #95' at the top?

:+1:  Otherwise it is also possible to add this in the merge commit.

base/src/fnv.rs
```diff
@@ -1,3 +1,9 @@
+/// Non-crypto HashMap using Fnv Hasher
+///
+/// The default hashing implementation in std::collections uses `SipHasher`
+/// since gluon doesn't need the cryptographic quarantee provided by SipHasher,
+/// we've opted for the faster fnv hash.
```
brendanzab: Should be above the struct! ^_^

brendanzab: Woops, I mean 'type alias'

Marwes:
The results vary quite a lot but for typechecking I am seeing a 5-10% speedup. Probably a bit overshadowed by inneficiencies elsewhere but its still a nice speedup!

Before
```
test clone_prelude     ... bench:     248,752 ns/iter (+/- 142,182)
test typecheck_prelude ... bench:  12,221,382 ns/iter (+/- 3,601,499)
test prelude ... bench:  19,270,757 ns/iter (+/- 1,817,400)
```
After
```
test clone_prelude     ... bench:     252,140 ns/iter (+/- 159,846)
test typecheck_prelude ... bench:  11,597,207 ns/iter (+/- 2,093,485)
test prelude ... bench:  18,961,071 ns/iter (+/- 3,236,183)
```
(clone_prelude measure overhead due to cloning the AST every iteration in `typecheck_prelude`)

Marwes:
I seem to have caused caused some merge conflicts by merging #102, sorry!

leshow:
Is there anything you still want me to do on this? Do you want me to rebase or merge and resolve the conflicts?

Marwes:
Just rebase and move the comment as @brendanzab wrote and this should be good!

Marwes:
Merged! Thanks!

brendanzab:
Thanks again @leshow!

PR #109, opened 2016-08-10
[WIP] docs: Add more explanations about the type system

Marwes:


Marwes:
I have only added an explanation about overloading so for but I will try to add in more over the course of the week.

Note that while I try to explain it in an authoritative manner I do not know if 'intersection inference' is sound as it currently is nor is it certain that this implementation of overloading will stick around (implicit arguments may supersede it). In any case having an actual explanation is needed to be able to judge it properly.

TUTORIAL.md
```diff
@@ -265,53 +309,26 @@ Ref String
...
...s common in many programming languages. Instead the typechecker checks the type of the binding already in scope and calculates the 'intersection' of the two bindings. In the example above the binding which already exists has the type `Int -> String` while the added binding is `Float -> String`. By calculating the 'intersection' the typechecker calculates the most specialized type which both bindings can still successfully unify to which in this case is `a -> String`. Any subsequent use of `showType` will then be seen as `a -> String` which succeeds with both a `Int` and `Float` argument. It would however also succeed if a `String` (or any other) type were used as argument which is not acceptable as we have only implemented `showType` for `Int` and `Float`. To catch this case (and to figure out which overload should be use where) the typechecker does an extra pass after successfully typechecking expression. In this pass all uses of overloaded bindings are checked against the overload candidates until a match is found and as the commented out call has the type `String -> String` it would be unable to unify with one of the existing bindings allowing the typechecker to reject the code.
```
brendanzab: Could we break this up into two paragraphs somehow?

TUTORIAL.md
```diff
@@ -265,53 +309,26 @@ Ref String
...
-do_something2 () // `do_something1 ()` is run, then `do_something_2`. The result of `type ...` is the result of the expression
-type PrivateType = | Private Int
-let x = Private (do_something3 ())
-do_something3 ()
-match x with
-| Private y -> do_something4 x
+showType 0 // Returns "Int"
+showType 0.0 // Returns "Float"
+// showType "" // Would be a type error
 ```
```
brendanzab: What casing are we using? This is camel, but most are snake case.

Also, spacing between the type and colon. It would be nice to have consistency (I prefer a space between, but eh).

TUTORIAL.md
```diff
@@ -265,53 +309,26 @@ Ref String
...
-#### Higher-kinded types (TODO)
+### Higher-kinded types (TODO)
 
 Gluon has higher kinded types.
 
-### Indentation
+### Overloading
 
-If you have been following along this far you may be thinking think that the syntax so far is pretty limiting. In particular you wouldn't be wrong in thinking that the `let` and `type` syntax are clunky due to their need to be closed by the `in` keyword. Luckily gluon offers a more convenient way of writing bindings by relying on indentation.
+Sometimes there there is a need to overload a name with multiple differing implementations and let the compiler chose the correct implementation depending on which the inferred type is. If you have written any amount of Gluon code so far you are likely to have already encountered this with numeric operators such as `(+)` or comparison operators such as `(==)`. While these operators are core parts of gluon they are not special cased by the compiler which lets you define and use overloaded bindings yourself.
```
brendanzab: Could probably cut ` on which the inferred type is`.

TUTORIAL.md
Marwes: I am destroyed from writing C++ for work! Will fix the spacing as well.

Marwes:
Fleshed out the section on HKT a bit but I started to feel that a proper explanation would get to long and is in any case better explained in any number of other tutorials.

Are there any other topics I should expand on in addition to the current changes?

TUTORIAL.md
```diff
@@ -313,6 +268,78 @@ in
...
 
+## Typesystem
+
+### Function types
+
+```
+<type> -> <type>
+```
+
+Function types are written using the `->` operator which is right associative. This means that the type of `(+)` which is usually written as `Int -> Int -> Int` is parsed as `Int -> (Int -> Int)` (A function taking one argument of `Int` and returning a function of `Int -> Int`).
```
brendanzab: > Function types are written using the `(->)` operator which is right associative. This means that the funtion type `Int -> (Int -> Int)` (A function taking one argument of `Int` and returning a function of `Int -> Int`) can be written as `Int -> Int -> Int` .

TUTORIAL.md
```diff
@@ -313,6 +268,78 @@ in
...
+
+### Record type
+
+```
+{ (<identifier> <identifier>* = <type>,)* (<identifier> : <type>,)* }
+{ pi : Float, sin : Float -> Float }
+```
+
+Records are gluon's main way of creating associating related data and they should look quite familiar if you are familiar with dynamic languages such as javascript. Looks can be deceiving however as gluon's records are more similiar to a struct in Rust or C as the order of the fields are significant and fields can't be added, removed or modified.
```
brendanzab: > the order of the fields are significant

What does this mean?

Marwes: That  `{ x : Int, y : String } != { y : String, x : Int }`. It might be a bit cumbersome but it lets the interpreter access fields just by offset instead of by an hashmap. I can clarify it a bit.

brendanzab: Wow - I did not know this!

Marwes:
Merging this, might as well make a new PR later if there is more to be added.

brendanzab:
Sounds good :)

PR #110, opened 2016-08-12
chore: Enable workspaces in cargo

Marwes:
This will let nightly users to build directly in the root folder.


Marwes:
@brendanzab I remembered you removed this from #64 in case as a cautionary measure but skeptic is failing happily on its own so I think this should be a simple improvement to avoid building from the c-api directory.

Issue #108, opened 2016-08-10
Consider using byte positions for source locations internally


brendanzab:
Rather than keeping the row and column information from lexing onwards, rustc and LALRPOP use byte positions for source locations internally. They then reconstruct row and column information later. Not sure what all the the pros and cons are, but here are some:

- slightly less memory used in AST
- easier to integrate with LALRPOP

Maybe @nikomatsakis can weigh in on some of the reasoning behind this?

- `libsyntax_pos`: https://github.com/rust-lang/rust/blob/ae774103501337ed63b42b673c6c4fdbf369e80e/src/libsyntax_pos/lib.rs
- `lalrpop::file_text`: https://github.com/nikomatsakis/lalrpop/blob/125cf975e36cca84b18c7faf0b9187e7d46c5a6d/lalrpop/src/file_text.rs

brendanzab:
We might have to weigh up the fact the gluon is intended to be embeddable, but I'm not sure whether this would have an impact or not.

nikomatsakis:
If you are writing your own lexer, LALRPOP can handle custom location types, no problem. So I think that LALRPOP integration is not really an important factor.

However,  I do like using the byte-offset nonetheless. As you say, it's more compact in memory, which is the most important consideration. It's also perfectly general, simple, and it allows you to do things like easily find the original source text from a given span. The downside is that you need some context to convert to a line-column, but I tend to find that printing errors needs context anyhow.

Marwes:
With line + column there is still the reverse problem as then one needs to calculate the byte positions for each line https://github.com/Marwes/gluon/blob/4a64c68d8d04a6788f1fea3d6f25471b873ee8e2/base/src/error.rs#L74.

Thus the only real problem is that the [errors in the typechecker](https://github.com/Marwes/gluon/blob/4a64c68d8d04a6788f1fea3d6f25471b873ee8e2/check/src/typecheck.rs#L210 ) (as an example) contains line + columns spans so the typechecker needs to know how to translate byte positions to line + column. Alternativley this conversion could be postponed until after typechecking if the typechecker stored something like `Errors<ByteSpanned<Error>>` and then that is translated to the current `Spanned<_>` type.

brendanzab:
Yep, I think we can keep around the list of line beginning offsets then reconstruct them later as needed. Seems to be what happens in those links I posted.

PR #112, opened 2016-08-14
Use typed byte and character positions

brendanzab:
This should make it easier to implement #108 in a type-driven way.

### DIFF
--- a/base/src/pos.rs
+++ b/base/src/pos.rs
@@ -109,7 +109,7 @@ impl fmt::Display for CharPos {
 /// A location in a source file
 #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, Ord, PartialOrd)]
 pub struct Location {
-    pub row: u32,
+    pub line: u32,
     pub column: CharPos,
     pub absolute: BytePos,
 }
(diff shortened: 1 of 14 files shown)


base/src/pos.rs
```diff
@@ -0,0 +1,198 @@
...
+/// A location in a source file
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, Ord, PartialOrd)]
+pub struct Location {
+    pub row: u32,
+    pub column: CharPos,
+    pub absolute: BytePos,
+}
+
+impl Location {
+    pub fn line_offset(mut self, offset: CharPos) -> Location {
```
Marwes: Maybe this could be `fn line_offset<T: Into<CharPos>>(mut self, offset: T) -> Location`. Could improve ergonomics a bit (though probably mucking about with character positions should not be to common).

Marwes: I suppose with #108 `line_offset` may not even be used as it may just be a byte addition. So maybe not worth making the change.

brendanzab: Yeah, I was planning on removing `line_offset` later one. It's only seems to be used in the approximations of spans.

Marwes:
Just merge when you are done :+1: .

PR #113, opened 2016-08-14
feat(repl): Add completion to the repl

Marwes:
This has the usual caveat that completing other modules will not work and they must be explicitly imported first

```
let prelude = import "std/prelude.glu"
prelude.n // Yields prelude.not, prelude.num_Int, etc
```
```
prelude. // Nothing
std.prelude. // Nothing
```

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -12,7 +12,7 @@ use vm::api::{IO, Function, WithVM, VmType, Userdata};
 use vm::gc::{Gc, Traverseable};
 use vm::thread::{Thread, RootStr};
 
-use gluon::{Compiler, new_vm};
+use gluon::{Compiler, new_vm, RootedThread, Result as GluonResult};
 
 fn type_of_expr(args: WithVM<RootStr>) -> IO<Result<String, String>> {
     let WithVM { vm, value: args } = args;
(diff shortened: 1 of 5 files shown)


Marwes:
completion

src/repl.rs
```diff
@@ -84,9 +84,53 @@ fn find_info(args: WithVM<RootStr>) -> IO<Result<String, String>> {
...
+    use base::ast;
+    use check::completion::suggest;
+    use gluon::compiler_pipeline::*;
 
-impl Userdata for Editor { }
+    let location = ast::Location {
+        row: 1,
+        column: pos as i32,
+        absolute: 0,
+    };
```
brendanzab: Will need to fix this to use the new position types.

Marwes: Yup, fixed.

src/repl.rs
```diff
@@ -84,9 +84,54 @@ fn find_info(args: WithVM<RootStr>) -> IO<Result<String, String>> {
...
+    use base::pos::{BytePos, CharPos, Location};
+    use base::ast::EmptyEnv;
+    use check::completion::suggest;
+    use gluon::compiler_pipeline::*;
 
-impl Userdata for Editor { }
+    let location = Location {
+        line: 1,
+        column: CharPos(pos),
+        absolute: BytePos(pos as u32),
```
brendanzab: Could have used `BytePos::from`, but no matter.

PR #115, opened 2016-08-16
Use byte offsets instead of locations in AST

### LINKED ISSUE #108: Consider using byte positions for source locations internally

brendanzab:
Fixes #108

### DIFF
--- a/base/src/ast.rs
+++ b/base/src/ast.rs
@@ -3,7 +3,7 @@
 use std::fmt;
 use std::ops::Deref;
 
-use pos::Spanned;
+use pos::{BytePos, Spanned};
 use symbol::Symbol;
 use types::{self, Alias, AliasData, Kind, Type, TypeEnv, TypeVariable};
 
@@ -179,7 +179,7 @@ pub enum LiteralEnum {
 }
 
 /// Pattern which contains a location
-pub type SpannedPattern<Id> = Spanned<Pattern<Id>>;
+pub type SpannedPattern<Id> = Spanned<Pattern<Id>, BytePos>;
 
 #[derive(Clone, PartialEq, Debug)]
 pub enum Pattern<Id: AstId> {
@@ -214,7 +214,7 @@ pub struct Lambda<Id: AstId> {
 }
 
 /// Expression which contains a location
-pub type SpannedExpr<Id> = Spanned<Expr<Id>>;
+pub type SpannedExpr<Id> = Spanned<Expr<Id>, BytePos>;
 
 #[derive(Clone, PartialEq, Debug)]
 pub enum Expr<Id: AstId> {
@@ -417,7 +417,7 @@ impl<Id> Typed for Expr<Id>
     }
 }
 
-impl<T: Typed> Typed for Spanned<T> {
+impl<T: Typed> Typed for Spanned<T, BytePos> {
     type Id = T::Id;
 
     fn env_type_of(&self, env: &TypeEnv) -> AstType<T::Id> {
--- a/base/src/error.rs
+++ b/base/src/error.rs
@@ -5,7 +5,8 @@ use std::any::Any;
 use std::error::Error as StdError;
 use std::fmt;
 
-use pos::Spanned;
+use pos::{BytePos, Spanned};
+use source::Source;
 
 /// An error type which can represent multiple errors.
 #[derive(Debug, PartialEq)]
@@ -45,72 +46,54 @@ impl<T: fmt::Display + fmt::Debug + Any> StdError for Errors<T> {
     }
 }
 
-
-#[derive(Debug)]
-struct SourceContext<E> {
-    context: String,
-    error: Spanned<E>,
-}
-
-fn extract_context<E>(lines: &[&str], error: Spanned<E>) -> SourceContext<E> {
-    SourceContext {
-        context: lines.get((error.span.start.line - 1) as usize)
-            .cloned()
-            .unwrap_or("N/A")
-            .to_string(),
-        error: error,
-    }
-}
-
 /// Error type which contains information of which file and where in the file the error occured
 #[derive(Debug)]
-pub struct InFile<E> {
-    file: String,
-    error: Errors<SourceContext<E>>,
+pub struct InFile {
+    message: String,
 }
 
-impl<E> InFile<E> {
+impl InFile {
     /// Creates a new `InFile` error which states that the error occured in `file` using the file
-    /// contents in `contents` to provide a context to the span.
-    pub fn new(file: String, contents: &str, error: Errors<Spanned<E>>) -> InFile<E> {
-        let lines: Vec<_> = contents.lines().collect();
-        InFile {
-            file: file,
-            error: Errors {
-                errors: error.errors
-                    .into_iter()
-                    .map(|error| extract_context(&lines, error))
-                    .collect(),
-            },
+    /// contents in `source` to provide a context to the span.
+    pub fn new<E: fmt::Display>(file: &str,
+                                source: &Source,
+                                error: Errors<Spanned<E, BytePos>>)
+                                -> InFile {
+        use std::fmt::Write;
+
+        let mut message = String::new();
+
+        for error in &error.errors {
+            let start = source.location(error.span.start).unwrap();
+            let end = source.location(error.span.end).unwrap();
+            let (_, line) = source.line(start.line).unwrap();
+
+            write!(message, "{}:{}\n{}\n", file, error.value, line).unwrap();
+
+            for _ in 1..start.column.to_usize() {
+                write!(message, " ").unwrap();
+            }
+
+            write!(message, "^").unwrap();
+
+            for _ in start.column.to_usize()..(end.column.to_usize() - 1) {
+                write!(message, "~").unwrap();
+            }
+
+            writeln!(message, "").unwrap();
         }
-    }
-    pub fn errors(self) -> Errors<Spanned<E>> {
-        Errors { errors: self.error.errors.into_iter().map(|err| err.error).collect() }
+
+        InFile { message: message }
     }
 }
 
-impl<E: fmt::Display> fmt::Display for InFile<E> {
+impl fmt::Display for InFile {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        for error in &self.error.errors {
-            try!(write!(f, "{}:{}\n{}\n", self.file, error.error, error.context));
-            for _ in 1..error.error.span.start.column.to_usize() {
-                try!(write!(f, " "));
-            }
-            try!(write!(f, "^"));
-            for _ in error.error.span.start.column.to_usize()..(error.error
-                .span
-                .end
-                .column
-                .to_usize() - 1) {
-                try!(write!(f, "~"));
-            }
-            try!(writeln!(f, ""));
-        }
-        Ok(())
+        self.message.fmt(f)
     }
 }
 
-impl<T: fmt::Display + fmt::Debug + Any> StdError for InFile<T> {
+impl StdError for InFile {
     fn description(&self) -> &str {
         "Error in file"
     }
--- a/base/src/error.rs
+++ b/base/src/error.rs
@@ -5,7 +5,7 @@ use std::any::Any;
 use std::error::Error as StdError;
 use std::fmt;
 
-use pos::{BytePos, Spanned};
+use pos::{BytePos, Location, Span, Spanned};
 use source::Source;
 
 /// An error type which can represent multiple errors.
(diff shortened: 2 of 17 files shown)


base/src/ast.rs
```diff
@@ -260,16 +260,18 @@ impl<Id> LExpr<Id>
     /// Returns the an approximation of the span of the expression
     pub fn span<'a>(&self, env: &(DisplayEnv<Ident = Id> + 'a)) -> Span {
         use self::Expr::*;
+
+        let start = self.location.absolute;
```
Marwes: I suppose I already mentioned this on gitter but to document it here as well.

`absolute` is never set to a proper value in the parser so it can't be used :(

base/src/source.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// Path to the file that this source code coresponds to
+    FilePath(PathBuf),
+}
+
+impl fmt::Display for SourceName {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            SourceName::Placeholder(ref name) => write!(f, "<{}>", name),
+            SourceName::FilePath(ref name) => name.to_str().expect("Invalid filename").fmt(f),
+        }
```
Marwes: I think its better to use [display](https://doc.rust-lang.org/std/path/struct.Path.html#method.display) here (a method I never heard of before, wonder why it is needed)

base/src/source.rs
```diff
@@ -0,0 +1,189 @@
...
+
+impl Source {
+    pub fn new(name: SourceName, src: String) -> Source {
+        use std::iter;
+
+        let lines = {
+            let input_indices = src.as_bytes()
+                .iter()
+                .enumerate()
+                .filter(|&(_, &b)| b == ('\n' as u8))
```
Marwes: s/('\n' as u8)/b'\n'/

base/src/source.rs
```diff
@@ -0,0 +1,189 @@
...
+            .filter(|&i| self.lines[i] > byte)
+            .map(|i| i - 1)
+            .next()
+            .unwrap_or(num_lines - 1) + 1
+    }
+
+    pub fn location(&self, byte: BytePos) -> Option<Location> {
+        let line_number = self.line_number_at_byte(byte);
+
+        println!("line_number: {:?}", line_number);
```
Marwes: WIP I assume

base/src/source.rs
```diff
@@ -0,0 +1,185 @@
...
+
+                if curr_byte == byte {
+                    return Some(Location {
+                        line: line_number as u32,
+                        column: CharPos(col + 1),
+                        absolute: byte,
+                    });
+                }
+            }
+            println!("line overflowed!");
```
Marwes: Don't forget to remove this

base/src/source.rs
```diff
@@ -0,0 +1,185 @@
...
+            (start, line)
+        })
+    }
+
+    pub fn line_at_byte(&self, byte: BytePos) -> Option<(BytePos, &str)> {
+        let line_number = self.line_number_at_byte(byte);
+        self.line(line_number)
+    }
+
+    pub fn line_number_at_byte(&self, byte: BytePos) -> usize {
```
Marwes: Maybe it is better to return `None` if the byte is outside the source?

parser/src/lexer.rs
```diff
@@ -222,7 +222,17 @@ impl<Id> Token<Id> {
...
+pub struct Span {
+    pub start: Location,
+    pub end: Location,
+}
+
+#[derive(Clone, Debug)]
+pub struct SpannedToken<Id> {
+    pub span: Span,
+    pub value: Token<Id>,
+}
```
Marwes: I guess these may get removed sometime later since the lexer should just need byte position and column position? Otherwise `pos::Span` could be generic over the location

```
pub struct Span<P> {
    pub start: P,
    pub end: P
}
```

Marwes:
Taking a look at fixing the completion errors, I think I may have messed them up due to \r\n :/

brendanzab:
Perhaps in the test helper function we could replace `\n\r` with `\n`?

Marwes:
In rustc raw strings always use \n for newlines, but since they are checked out as \r\n so when I counted them manually I set the wrong offsets. They only worked previously because the line and column fields were the primary way the locations were determined.

Marwes:
Might add a .gitattributes which disables autocrlf so it always checks out as \n

brendanzab:
Not sure why this is failing!

```
     Running /home/travis/build/Marwes/gluon/target/debug/deps/repl-ef5bebea2058d017

running 4 tests

test repl::tests::compile_repl_test ... ok

test repl::tests::find_info ... ok

test repl::tests::find_kind ... ok
test repl::tests::type_of_expr ... FAILED

failures:

---- repl::tests::type_of_expr stdout ----

	thread 'repl::tests::type_of_expr' panicked at 'called `Option::unwrap()` on a `None` value', ../src/libcore/option.rs:326

```

Marwes:
Its not failing locally?

brendanzab:
It is as well :/

Marwes:
What says RUST_BACKTRACE=1?

brendanzab:
haharrr

```
   1:        0x10d48ca19 - std::sys::backtrace::tracing::imp::write::h482d45d91246faa2
   2:        0x10d48f530 - std::panicking::default_hook::_{{closure}}::h89158f66286b674e
   3:        0x10d48e6df - std::panicking::default_hook::h9e30d428ee3b0c43
   4:        0x10d48ede6 - std::panicking::rust_panic_with_hook::h2224f33fb7bf2f4c
   5:        0x10d48ec84 - std::panicking::begin_panic::hcb11a4dc6d779ae5
   6:        0x10d48eb92 - std::panicking::begin_panic_fmt::h310416c62f3935b3
   7:        0x10d48eaf7 - rust_begin_unwind
   8:        0x10d4bc2a0 - core::panicking::panic_fmt::hc5789f4e80194729
   9:        0x10d4bc1a4 - core::panicking::panic::h1953378f4b37b561
  10:        0x10cdff1d1 - _<core..option..Option<T>>::unwrap::hc34fbdaa73345c47
  11:        0x10cdebf4b - gluon_base::error::InFile::new::h870da4ab659f56f3
  12:        0x10ce83b3f - gluon::Compiler::typecheck_expr_expected::_{{closure}}::h437090b56b05f78b
  13:        0x10ce13101 - _<core..result..Result<T, E>>::map_err::hdb2bd6c907f4b667
  14:        0x10ce7ccd2 - gluon::Compiler::typecheck_expr_expected::hbfc9c938e7a5fc1c
  15:        0x10ce7d624 - gluon::Compiler::typecheck_str::h2699ffbf9d3bcc5d
  16:        0x10cd7b4d6 - repl::repl::type_of_expr::hbf28cdb0209e8c95
  17:        0x10cd6a953 - _<fn(A) .> R as gluon_vm..api..VmFunction<'vm>>::unpack_and_call::h847a3693a23fec2f
  18:        0x10cd7fad3 - repl::repl::compile_repl::wrapper::h249ebd72ac283f03
  19:        0x10cd82ae4 - fn(&vm..thread..Thread) .> vm..thread..Status::fn_pointer_shim.16439::h5a6dee19077ef451
  20:        0x10d140b86 - gluon_vm::thread::Context::execute_function::h23fda8b309141593
  21:        0x10d143aa0 - gluon_vm::thread::Context::execute::ha3a2979a19af173b
  22:        0x10d13cc81 - gluon_vm::thread::Thread::call_context::h4872b6d4f3c264da
  23:        0x10d13fbe8 - _<gluon_vm..thread..Thread as gluon_vm..thread..ThreadInternal>::call_function::h92c8ef59273de6d8
  24:        0x10cd5b0ff - _<gluon_vm..api..Function<T, fn(A) .> R>>::call::h045137cc7d467925
  25:        0x10cd7fefb - repl::repl::tests::type_of_expr::hb9919e556f63907e
  26:        0x10cdbcace - _<F as alloc..boxed..FnBox<A>>::call_box::h6d40ac6ff8180289
  27:        0x10cdb5c54 - std::panicking::try::do_call::hf4aa30d569353639
  28:        0x10d48fb1a - __rust_maybe_catch_panic
  29:        0x10cdbc4d7 - _<F as alloc..boxed..FnBox<A>>::call_box::h116d9064e2ce57af
  30:        0x10d48de95 - std::sys::thread::Thread::new::thread_start::he0bf102845911132
  31:     0x7fff8537999c - _pthread_body
  32:     0x7fff85379919 - _pthread_start
```

base/src/error.rs
```diff
@@ -45,72 +46,54 @@ impl<T: fmt::Display + fmt::Debug + Any> StdError for Errors<T> {
...
+                                error: Errors<Spanned<E, BytePos>>)
+                                -> InFile {
+        use std::fmt::Write;
+
+        let mut message = String::new();
+
+        for error in &error.errors {
+            let start = source.location(error.span.start).unwrap();
+            let end = source.location(error.span.end).unwrap();
+            let (_, line) = source.line(start.line).unwrap();
```
Marwes: Guessing its one of these that is panicking, either return an error or clamp the positions so they are within the file. I can investigate further after I am home from work if you like.

src/lib.rs
```diff
@@ -439,11 +446,12 @@ impl Compiler {
...
-                            vm: &Thread,
-                            file: &str,
-                            expr_str: &str)
-                            -> Result<(ast::SpannedExpr<ast::TcIdent<Symbol>>, TcType, Metadata)> {
+    pub fn extract_metadata
+        (&mut self,
+         vm: &Thread,
+         file: &str,
+         expr_str: &str)
+         -> Result<(ast::SpannedExpr<ast::TcIdent<Symbol>>, TcType, Metadata)> {
```
Marwes: Are you using a new version of rustfmt or a different config?

I don't have much of an opinion when it comes how code should be formatted as long as it is consistent (so if anything I just stick to the default since that is simpler).

brendanzab: Will check!

base/src/error.rs
Marwes: Its weird that this code is being run for `repl::tests::type_of_expr` as well. That test should not error in the first place.

base/src/error.rs
```diff
@@ -45,72 +46,54 @@ impl<T: fmt::Display + fmt::Debug + Any> StdError for Errors<T> {
...
+                write!(message, " ").unwrap();
+            }
+
+            write!(message, "^").unwrap();
+
+            for _ in start.column.to_usize()..(end.column.to_usize() - 1) {
+                write!(message, "~").unwrap();
+            }
+
+            writeln!(message, "").unwrap();
```
Marwes: Any reason why you write all errors to a string straight away? I find it quite nice to keep errors typed as far as possible so it is possible to match on them and maybe have multiple ways of displaying them.

base/src/pos.rs
```diff
@@ -139,35 +139,35 @@ impl fmt::Display for Location {
 
 /// A span between two locations in a source file
 #[derive(Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Debug)]
-pub struct Span {
-    pub start: Location,
-    pub end: Location,
+pub struct Span<Pos> {
+    pub start: Pos,
+    pub end: Pos,
```
Marwes: :+1:

base/src/source.rs
```diff
@@ -0,0 +1,148 @@
...
+        assert_eq!(source.line(5), Some((BytePos(20), "bloop")));
+        assert_eq!(source.line(6), Some((BytePos(26), "")));
+        assert_eq!(source.line(7), None);
+    }
+
+    #[test]
+    fn source_line_number_at_byte() {
+        assert_eq!(test_source().line_number_at_byte(BytePos(0)), 1);
+        assert_eq!(test_source().line_number_at_byte(BytePos(3)), 1);
+        assert_eq!(test_source().line_number_at_byte(BytePos(13)), 3);
```
Marwes: Nit. Add asserts for 13,14,15 (or any successive positions around a newline) to catch off by one errors

base/src/error.rs
brendanzab: I think you're right on this one! Not sure why I did...

base/src/error.rs
brendanzab: Feel free to investigate - a bit tired to do anything right now.

src/lib.rs
brendanzab: I upgraded to the latest one and it formats the same :/

brendanzab: I think it's just because the line is longer now.

Marwes: You may be right. Looked at the default settings for rustfmt and I think this is the default. Not sure why I haven't seen it before.

Marwes: (It seems to have been the default for some time as well https://github.com/rust-lang-nursery/rustfmt/blame/fe69e975e3032d5aa6f73ecdf9e35288d09e4dc1/src/config.rs#L362)

base/src/error.rs
Marwes: Doh. `type_of_expr` tests on the wrong thing and have always returned a typechecking error (`std.prelude.Option` is not a value that can be accssed...). It still succeeded though as it only checked for `Ok` on the outer result (ie that the call executed without panics or other errors, which it did).

The test should pass alright if `you change the test to take an expression which actually is actually valid like (Notice the double `Ok` applications...).
```
assert_eq!(type_of.call("123"), Ok(IO::Value(Ok("Int".into()))));
```

brendanzab:
Wooo, it passes!

PR #116, opened 2016-08-16
[WIP] Use spans instead of single locations in AST

brendanzab:


base/src/ast.rs
```diff
@@ -194,48 +194,48 @@ pub enum Pattern<Id: AstId> {
...
-        exprs: Vec<(Id::Untyped, Option<LExpr<Id>>)>,
+        exprs: Vec<(Id::Untyped, Option<SExpr<Id>>)>,
     },
     Lambda(Lambda<Id>),
-    Tuple(Vec<LExpr<Id>>),
-    Type(Vec<TypeBinding<Id::Untyped>>, Box<LExpr<Id>>),
-    Block(Vec<LExpr<Id>>),
+    Tuple(Vec<SExpr<Id>>),
+    Type(Vec<TypeBinding<Id::Untyped>>, Box<SExpr<Id>>),
+    Block(Vec<SExpr<Id>>),
```
Marwes: I think this is a bit to many constructors.

```rust
pub enum Expr<Id> {
    Atom(Id),
    Cons(SExpr<Id>, SExpr<Id>),
}
```
Should be enough no? :stuck_out_tongue:

base/src/pos.rs
```diff
@@ -135,64 +135,62 @@ pub struct Span {
...
-#[derive(Copy, Clone, PartialEq, Debug)]
-pub struct Spanned<T> {
-    pub span: Span,
-    pub value: T,
-}
-
-impl<T: fmt::Display> fmt::Display for Spanned<T> {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}: {}", self.span.start, self.value)
+    pub fn merge(self, other: Span) -> Option<Span> {
```
Marwes: Didn't see this used anywhere (yet)? Do you have a usecase in mind?

base/src/ast.rs
brendanzab: Yes - I was wondering about that! `SpExpr`? :P

Marwes: Nah, `SExpr` is fine

base/src/pos.rs
brendanzab: I was thinking that this would allow us to expand the spans during parsing, but we are only emitting located tokens right now....

brendanzab:
Updated - I am not expanding the spans in the parser though, so I need to figure that out... :/

PR #117, opened 2016-08-18
fix(completion): Give completion for local variables when pointing to whitespace

Marwes:
Before the closest identifier to the left of the location would be completed which meant
```f#
test   abc
//   ^ completion here would complete `test` instead of listing local variables
```
This fix ensures that local variables are enumerated instead.

### DIFF
--- a/check/src/completion.rs
+++ b/check/src/completion.rs
@@ -174,7 +174,7 @@ impl<F> FindVisitor<F>
         
         match current.value {
             Identifier(_) | Literal(_) => {
-                if current.span(self.env).containment(&self.location) == Ordering::Equal {
+                if current.span.containment(&self.location) == Ordering::Equal {
                     self.on_found.expr(current)
                 } else {
                     self.on_found.nothing()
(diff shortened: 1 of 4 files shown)


Marwes:
While testing this I found that operators will be enumerated without parentheses. That is perfectly legal but it might not be what is expected.

```f#
test 
//    ^ completion here will enumerate '+', '-' etc 
// Is it better if it is '(+)', (-), etc ?
```

base/src/ast.rs
```diff
@@ -261,7 +261,12 @@ impl<Id> LExpr<Id>
     pub fn span<'a>(&self, env: &(DisplayEnv<Ident = Id> + 'a)) -> Span {
         use self::Expr::*;
         let end = match self.value {
-            Identifier(ref id) => self.location.line_offset(CharPos(env.string(id).len())),
+            Identifier(ref id) => {
+                let name = env.string(id);
+                // Remove the `:<location>` suffix generated when renaming creates a unique name
+                let offset = name.split(':').next().unwrap_or(name).len();
+                self.location.line_offset(CharPos::from(offset))
+            }
```
Marwes: This is a workaround due to the current renaming behaviour. With #116 this is obviously not nececssary

Marwes:
Copied the byte positions in the compleition tests from #115 to get the tests to pass again so I guess one of these is going to get a merge conflict.

brendanzab:
You'll need to rebase this now - the completion API now uses byte positions, not locations.

Marwes:
Only after #115 is merged, we can wait with merging this until after that as I only need to delete b576cf4d7f78241bbe751aa3bac2577dd170e6f1

brendanzab:
Oh, durp, for some reason I thought I'd finished it!

PR #120, opened 2016-08-19
chore: Enable building with stable rustc again on travis

Marwes:
With 1.11 released building on stable should no longer trigger timeouts.


Marwes:
Superseeded by #122

PR #123, opened 2016-08-20
Use spans instead of single locations in AST

### LINKED ISSUE #121: Use Spans in the AST
This will mean that we can get rid of the `LExpr::span` and `LPattern::span` methods, making our locations much less error prone. The removal of those methods will also make it easier to switch to byte positions in #108, and make the switch to LALRPOP easier, as discussed in #69.

I have started this in #116, but it seemed to be more than I could chew at this stage.

Marwes:
Continuation of #116 

Closes #121

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -30,10 +30,7 @@ use combine_language::{Assoc, Fixity, expression_parser};
 
 use lexer::{Lexer, Delimiter, Token, IdentType};
 
-pub type Error = ParseError<BufferedStream<'static,
-                                           Lexer<'static,
-                                                 &'static str,
-                                                 &'static mut IdentEnv<Ident = String>>>>;
+pub type Error = ParseError<Wrapper<'static, 'static, 'static, String>>;
 
 /// Parser passes the environment to each parser function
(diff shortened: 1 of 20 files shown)


parser/src/lib.rs
```diff
@@ -401,16 +398,16 @@ impl<'s, I, Id, F> ParserEnv<I, F>
...
-            located(Location {
-                        column: CharPos(position.column as usize),
-                        line: position.line as u32,
-                        absolute: BytePos(0),
-                    },
-                    expr)
+            pos::spanned(Span {
+                             start: position,
+                             // ARGH - TODO
+                             end: position,
```
Marwes: Haven't fixed this yet so spans are not working yet

brendanzab:
Might want to rebase onto master so that beta passes as well

Marwes:
Oh yeah, forgot that.

brendanzab:
Hmm - did you rebase onto the right master branch?

Marwes:
Now it should be fixed, accidentally had the addition of stable commit to my local master and I forgot to pull in the master from github.

PR #124, opened 2016-08-21
Delay string interning until parse time

brendanzab:
The strings are already allocated by the parser, so it is easier just to delay this until we hit the parser. This greatle simplifies the type signatures in the lexer, and will make it much easier to transition to LALRPOP. For efficiency's sake, we should eventually transition to using slices in the lexer though.

### DIFF
--- a/parser/tests/basic.rs
+++ b/parser/tests/basic.rs
@@ -188,8 +188,6 @@ y
 #[test]
 fn expression() {
     let _ = ::env_logger::init();
-    let e = parse("2 * 3 + 4");
-    assert_eq!(e, Ok(binop(binop(int(2), "*", int(3)), "+", int(4))));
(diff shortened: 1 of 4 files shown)


Marwes:
I took the evil path and force pushed to your branch so if you want to add something to this make sure you pull down the remote branch

parser/tests/basic.rs
```diff
@@ -188,8 +188,6 @@ y
 #[test]
 fn expression() {
     let _ = ::env_logger::init();
-    let e = parse("2 * 3 + 4");
-    assert_eq!(e, Ok(binop(binop(int(2), "*", int(3)), "+", int(4))));
```
brendanzab: Seems this test was lost in the rebase - I restored it in #125

PR #125, opened 2016-08-21
Add holes to the type syntax, and use them when building the AST

brendanzab:


brendanzab:
I tried to do this in `substitution.rs`, but ran into bounds check errors:

```diff
 impl<T: Substitutable> Substitution<T> {
     pub fn new() -> Substitution<T> {
-        let variables = FixedVec::new();
-        // 0 is the default, uninitialized variable so we add it to make sure that the first
-        // variable we create through `new_var` is 1
-        variables.push(T::new(0));
         Substitution {
             union: RefCell::new(QuickFindUf::new(1)),
-            variables: variables,
+            variables: FixedVec::new(),
             types: FixedMap::new(),
         }
     }

     pub fn clear(&mut self) {
         self.types.clear();
         self.variables.clear();
-        self.variables.push(T::new(0));
     }
```

Marwes:
Change
```
 union: RefCell::new(QuickFindUf::new(0)),
```

brendanzab:
I swear I tried that - will do again

Marwes:
Probably an (unrelated) bug in `clear` as since that does not clear `union` (I hink `clear` is not used though)

brendanzab:
There was an underflow on `(self.variables.borrow().len() - 1) as u32` - I just removed the `-1`

brendanzab:
Now I get:

```
	thread 'function_call' panicked at 'assertion failed: `(left == right)` (left: `Ok(App(Builtin(Function), [Variable(TypeVariable { kind: Type, id: 4 }), Variable(TypeVariable { kind: Type, id: 5 })]))`, right: `Ok(App(Builtin(Function), [Generic(Generic { kind: Type, id: 0x10c40c0b8:a0 }), Generic(Generic { kind: Type, id: 0x10c40c098:a1 })]))`)', /Users/brendan/dev/rust/gluon/check/tests/completion.rs:129
```

Marwes:
```
    fn typecheck_bindings(&mut self, bindings: &mut [ast::Binding<TcIdent>]) -> TcResult<()> {
        self.enter_scope();
        self.type_variables.enter_scope();
        let level = self.subs.var_id(); // Add + 1 here
```

In typecheck.rs: finish_type

```
                Type::Variable(ref var) if self.subs.get_level(var.id) > level => { // Change to >=
```

Marwes:
The number which represent a type variable also represent 'levels' (or how deep in the expression the variable was created). Variables in types of let bindings only get generalized if the variable does not exist outside the current level. (Bad explanation but maybe you sort of know how generalizing work, otherwise I can expand on it)

Marwes:
If that does not work then keep the behaviour of starting from 1 :P.

brendanzab:
Ah, right - are these debruijn indices?

brendanzab:
Still same error unfortunately. I would prefer not to leave this dead code in there though to trip somebody up...

Marwes:
Not sure, have only heard about de bruijin indexes with no context. The generalization is taken from my haskell compiler, which I no longer remember from what it can from, either from GHC, a tutorial or I came up on it on my own.

It is similiar/identical to this http://okmij.org/ftp/ML/generalization.html (which I found a 1-2 years ago).

Marwes:
Partly just guessing what is wrong since I haven't been able to test and take a proper look yet

brendanzab:
Ok now in `unify::test::intersection_test` at `/check/src/unify.rs:374`:

- left: `TType(Arrow(TType(Variable(1)), TType(Variable(1))))`
- right: `TType(Arrow(TType(Variable(2)), TType(Variable(2))))`

brendanzab:
oh wait, that is right? right?

Marwes:
Yup that is correct

Marwes:
"There are only 2 things hard in programming. Naming things, cache invalidation and off by one errors."

(Found a off by one error in combine as well which was the cause of the spans being wrong after changing the tokens to contain slices).

check/src/kindcheck.rs
```diff
@@ -151,6 +151,7 @@ impl<'a> KindCheck<'a> {
 
     fn kindcheck(&mut self, typ: &TcType) -> Result<(RcKind, TcType)> {
         match **typ {
+            Type::Hole => Ok((self.subs.new_var(), Type::hole())),
```
Marwes: s/Type::hole()/typ.clone/

For better sharing

brendanzab: Done!

Marwes: Any reason why you are not just calling typ.clone()? clarity?

Thought that may be better as with a cache it will still allocate a `Type::Hole` even if a hole is not used (and with clone() we are sharing a hole which definitely exists)

brendanzab: Ohhh - forgot that

parser/tests/basic.rs
```diff
@@ -188,6 +188,8 @@ y
 #[test]
 fn expression() {
     let _ = ::env_logger::init();
+    let e = parse("2 * 3 + 4");
+    assert_eq!(e, Ok(binop(binop(int(2), "*", int(3)), "+", int(4))));
```
Marwes: Oops, were testing the next test case in this function so I removed this and forgot to add it back. :sweat:

Marwes:
:+1: Merge when travis is happy.

parser/tests/basic.rs
brendanzab: Probably a sign it should be broken into two test cases in that case 😉

Marwes: Also true. You don't expect tests as simple as these to break though...

PR #129, opened 2016-08-24
perf: Use SmallVec in the Type::App variant

Marwes:
Since applications are usually done on just a few argument it is usually unnecessary to allocate a `Vec`. Using `SmallVec` this heap allocation can be avoided for applications with only a few arguments which should cover most of the cases.
#50


Marwes:
Not as happy as I expected with this PR as `SmallVec` currently has two drop flags which causes `Type` to grow by 8 bytes. Once https://github.com/rust-lang/rust/pull/35764 gets merged it should be possible to actually increase the number of arguments to 3 (from the current 2) while still keeping the same size for `Type` (thought naturally this would be worse for anything else than nightly until the trains catch up).

Atm I am only seeing a minor improvement  in the `typecheck_prelude` benchmark so maybe hold of on merging this until it can be tested with the aforementioned rustc PR.

lfairy:
Heads up, rust-lang/rust#35764 was merged a while ago. Might like to revisit this change.

Marwes:
I think I redid the benchmarks after rust-lang/rust#35764 was merged but didn't see any significant improvements (allocation does not seem to be a large part of the runtime currently, but admittedly it might be shadowed by other things). Going to redo it anyway to verify (and remember to update this PR).

I believe drop-flags will only be removed when 1.13 is released however so In any case it might be best waiting for 1.13 before merging.

Marwes:
Rebased and removed the intermediate `Vecs` used when constructing `AppVecs` (which gave an additional ~5% speedup). Used the master branch of https://github.com/servo/rust-smallvec so we should wait with merging this until at least a new version on cargo is released. May also want to wait for rust 1.13 as this will likely be a regression in performance on stable otherwise.

# SmallVec
```
test typecheck_prelude ... bench:  13,559,631 ns/iter (+/- 569,445)
```

# Vec
```
test typecheck_prelude ... bench:  15,035,660 ns/iter (+/- 570,014)
```

check/src/unify_type.rs
```diff
@@ -180,16 +182,22 @@ fn do_zip_match<'a, U>(self_: &ArcType,
                 }
                 Less => {
                     let offset = r_args.len() - l_args.len();
-                    let new_type =
-                        unifier.try_match(l, &Type::app(r.clone(), r_args[..offset].into()));
+
+                    let reduced_r = Type::app(r.clone(),
+                                              r_args[..offset].iter().cloned().collect());
```
brendanzab: Could you make this an intermediate? like:

```rust
let offset = r_args.len() - l_args.len();

let reduced_r_args = r_args[..offset].iter().cloned();
let reduced_r = Type::app(r.clone(), reduced_r_args.collect());
```

Just find the weird indentation kinda ugly.

check/tests/pass.rs
```diff
@@ -346,10 +348,10 @@ test
     let result = support::typecheck(text);
     assert!(result.is_ok(), "{}", result.unwrap_err());
 
-    let variants = Type::variants(vec![(support::intern_unscoped("T"),
+    let variants = Type::variants(collect![(support::intern_unscoped("T"),
                                         Type::function(vec![typ("a")],
                                                        support::typ_a("Test", vec![typ("a")])))]);
```
brendanzab: Indentation is now out of alignment.

base/src/types.rs
```diff
@@ -1212,7 +1225,7 @@ fn walk_move_types2<'a, I, F, T>(mut types: I, replaced: bool, output: &mut Vec<
             Some(typ) => {
                 output.push(typ);
             }
-            None if replaced || !output.is_empty() => {
+            None if replaced || output.len() != 0 => {
```
brendanzab: Reasoning?

base/src/types.rs
```diff
@@ -1186,24 +1196,27 @@ pub fn walk_move_type_opt<F: ?Sized, I, T>(typ: &Type<I, T>, f: &mut F) -> Optio
...
     where I: IntoIterator<Item = &'a T>,
           F: FnMut(&'a T) -> Option<T>,
           T: Clone + 'a,
+          R: Default + VecLike<T>,
 {
-    let mut out = Vec::new();
+    let mut out = R::default();
     walk_move_types2(types.into_iter(), false, &mut out, &mut f);
-    if out.is_empty() {
+    if out.len() == 0 {
```
brendanzab: Ditto

check/tests/pass.rs
Marwes: Ihave got rustfmt set to format automatically but I think `vec!` is special cased, other macros do not get formatted...

base/src/types.rs
brendanzab: Dang - yeah, a PR would probably be useful there.

check/src/unify_type.rs
Marwes: I find it hard to read with to many superfluous intermediates as well (neither of the names `reduced_r` or `reduced_r_args` says anything). I will see what I can do, may be able to simply the code overall.

check/tests/pass.rs
brendanzab: Yep, that's right :(

check/src/unify_type.rs
brendanzab: Agreed...  no real 'best way' here. Not a big deal - I agree it might be a tad superfluous here.

_shakes fist at Rust's ugly syntax and rightward drift_

😆

Marwes:
Rebase. Since rust 1.13 is release there is no reason to delay merging this anymore I think. Disabled osx builds since they take forever to start.

base/src/types.rs
```diff
@@ -1077,12 +1085,14 @@ pub fn walk_move_type_opt<F: ?Sized, I, T>(typ: &Type<I, T>, f: &mut F) -> Optio
     }
 }
 
-pub fn walk_move_types<'a, I, F, T>(types: I, mut f: F) -> Option<Vec<T>>
+// FIXME Add R: Default and remove the `out` parameter
+pub fn walk_move_types<'a, I, F, T, R>(types: I, mut f: F) -> Option<R>
```
brendanzab: Could you elaborate on this?

check/src/unify_type.rs
```diff
@@ -277,6 +255,53 @@ fn do_zip_match<'a, U>(self_: &ArcType,
...
+
+            let reduced_l = Type::app(l.clone(), l_args[..offset].iter().cloned().collect());
+            let new_type = unifier.try_match(&reduced_l, r);
+
+            let new_args = walk_move_types(l_args[offset..].iter().zip(r_args),
+                                           |l, r| unifier.try_match(l, r));
+            Ok(merge(r, new_type, r_args, new_args, Type::app))
+        }
+    }
+}
```
brendanzab: Love this - great idea breaking it out into another function ❤️

base/src/types.rs
Marwes: Just forgot to remove that comment when removing the out parameter, fixed.

brendanzab: :D

PR #130, opened 2016-08-24
refactor(base): Remove the duplicate walk_move_type functions from in…

Marwes:
…stantiate.rs

After TypeVisitor was implemented instantiate can use the normal walk_move_type implementation which now avoids recursing with a `FnMut` argument.


Marwes:
Might as well remove this from #129 since we may want to hold of on that.

Marwes:
Add the newline but does having a newline matter in any way? I never really paid attention to it.

brendanzab:
Causes messy diffs. Say if you change the `n-1` line, the diff will also show the `n` line as having been changed, even though it stayed the same.

PR #131, opened 2016-08-25
[WIP] refactor: Remove the Kind enum by unifying it with Type

Marwes:
By removing the `Kind` enum we cut out a good amount of duplication as
kind unification just work through type unification. There is a slight
increase in complexity through the addition of `Type` as a builtin type
but this otherwise looks like a nice improvement.

### DIFF
--- a/base/src/ast.rs
+++ b/base/src/ast.rs
@@ -5,7 +5,8 @@ use std::ops::Deref;
 
 use pos::Spanned;
 use symbol::Symbol;
-use types::{self, Alias, AliasData, Kind, Type, TypeEnv, TypeVariable};
+use types::{Alias, AliasData, BuiltinType, Type, TypeEnv, TypeVariable};
+use instantiate::instantiate;
 
 pub type AstType<Id> = ::types::ArcType<Id>;
 
@@ -110,7 +111,7 @@ impl<Id> TcIdent<Id> {
         TcIdent {
             typ: Type::variable(TypeVariable {
                 id: 0,
-                kind: Kind::variable(0),
+                kind: Type::builtin(BuiltinType::Type),
             }),
             name: name,
         }
@@ -467,7 +468,7 @@ fn get_return_type(env: &TypeEnv,
                             None => panic!("Unexpected type {:?} is not a function", alias_type),
                         };
 
-                        let typ = types::instantiate(typ.clone(), |gen| {
+                        let typ = instantiate(typ.clone(), |gen| {
                             // Replace the generic variable with the type from the list
                             // or if it is not found the make a fresh variable
                             args.iter()
--- a/base/src/types.rs
+++ b/base/src/types.rs
@@ -18,11 +18,11 @@ pub type TcIdent = ast::TcIdent<Symbol>;
 /// Trait for values which contains kinded values which can be refered by name
 pub trait KindEnv {
     /// Returns the kind of the type `type_name`
-    fn find_kind(&self, type_name: &SymbolRef) -> Option<RcKind>;
+    fn find_kind(&self, type_name: &SymbolRef) -> Option<TcType>;
 }
 
 impl<'a, T: ?Sized + KindEnv> KindEnv for &'a T {
-    fn find_kind(&self, id: &SymbolRef) -> Option<RcKind> {
+    fn find_kind(&self, id: &SymbolRef) -> Option<TcType> {
         (**self).find_kind(id)
     }
 }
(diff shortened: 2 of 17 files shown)


Marwes:
There is a slight increase in runtime in the `typecheck_prelude` benchmark and there may be slightly higher memory use due to `Type` having a larger size than `Kind`. These two regressions can probably be recovered however (though maybe not in this PR). Still looking to cut out the `KindEnv` trait which may also require the `TypeEnv` trait to be changed.

Marwes:
@brendanzab No need to go through nits on this just yet (though feedback on the overall idea/design would be nice) as this current commit is just about ripping out `Kind`.

base/src/ast.rs
```diff
@@ -110,7 +111,7 @@ impl<Id> TcIdent<Id> {
         TcIdent {
             typ: Type::variable(TypeVariable {
                 id: 0,
-                kind: Kind::variable(0),
+                kind: Type::builtin(BuiltinType::Type),
```
brendanzab: Should this not be generic at this stage?

Marwes: Oh yeah that should even be `TcIdent { typ: Type::hole(), .. }`

base/src/types.rs
```diff
@@ -229,6 +217,7 @@ pub enum BuiltinType {
     Array,
     /// Type constructor for functions, `(->) a b : Type -> Type -> Type`
     Function,
+    Type,
```
brendanzab: So you decided not to use a universe variable here? Or will that come later?

check/src/kindcheck.rs
```diff
@@ -73,15 +50,15 @@ impl<'a> KindCheck<'a> {
             subs: subs,
             type_kind: typ.clone(),
             function1_kind: function1_kind.clone(),
```
brendanzab: You could just instantiate `function1_kind` in place here now.

check/src/kindcheck.rs
```diff
@@ -143,7 +120,7 @@ impl<'a> KindCheck<'a> {
     fn builtin_kind(&self, typ: BuiltinType) -> RcKind {
         match typ {
             BuiltinType::String | BuiltinType::Byte | BuiltinType::Char | BuiltinType::Int |
-            BuiltinType::Float | BuiltinType::Unit => self.type_kind(),
+            BuiltinType::Float | BuiltinType::Unit | BuiltinType::Type => self.type_kind(),
```
brendanzab: This is incorrect - should be the next universe up.

base/src/types.rs
brendanzab: I'm also thinking it might make sense to put it on the main enum, seeing as it is core to our language. Still umming over `Function` too...

Marwes: I must admit I am a bit out of my depth when it comes to this stuff. That being said, as far as I understand does not universes come into play when there may be mixing of types of multiple universes. As long as types and kinds (and values) are separate as they are now universes should be implicitly defined simply from the context the type are in. Might be wrong though :worried:

check/src/kindcheck.rs
```diff
@@ -143,7 +120,7 @@ impl<'a> KindCheck<'a> {
...
             BuiltinType::Array => self.function1_kind(),
             BuiltinType::Function => self.function2_kind(),
```
brendanzab: If we did have universes, I'm not sure how we would handle them here and in other places. We'd need to figure out structurally where the largest universe is within a type... I think?

check/src/kindcheck.rs
Marwes: That would allocate one extra type, not a big deal but still...

brendanzab:
We could probably remove `:k` from the repl now too :)

brendanzab:
I hope I wasn't too nit-picky!

base/src/ast.rs
brendanzab: Yep, was just thinking that!

Marwes:
> I hope I wasn't too nit-picky!

I don't mind, just figured I would likely find a lot of nits myself as I go through and fix the aforementioned unfinished stuff.

check/src/kindcheck.rs
Marwes: See https://github.com/Marwes/gluon/pull/131#discussion_r76334248. I think at least as inference works now (types, kinds etc are all separate, only the representation is unified), this should be fine. If that is true maybe it is fine to omit universe variables for now.

check/src/kindcheck.rs
brendanzab: Ohhh. But you could do `function2_kind: Kind::function(vec![typ.clone(), typ.clone()], typ)`. Not sure which is better...

check/src/kindcheck.rs
brendanzab: I think we should at least make a new issue to introduce cumulative universes after this is merged.

Marwes:
Without mixing of types from different universes, is this change a good idea? With 'kinds' represented by `Type` there are a lot of constructors which 'kinds' never use while previously it was explicit which representations a kind has.

brendanzab:
Lets be cautious and leave this open and have a think, rather than jumping in too fast.

check/src/kindcheck.rs
lfairy: [GHC Haskell does something similar](https://ghc.haskell.org/trac/ghc/wiki/DependentHaskell/Phase1#Kindsandtypesarethesame), and they claim that (in their own case at least) it is consistent.

brendanzab: We could just have a `Type1` variant that should not be definable. No need for the complexity of full universes.

Issue #114, opened 2016-08-15
Discuss type equality based on record field order


brendanzab:
Currently records with identical fields but with different ordering don't unify, for example: `{ x : Int, y : String } != { y : String, x : Int }`. As commented at https://github.com/Marwes/gluon/pull/109#discussion_r74745341:

> It might be a bit cumbersome but it lets the interpreter access fields just by offset instead of by an hashmap.

Do we want to continue using these semantics going forward? I personally find it a bit surprising. Perhaps we could have some deterministic ordering of fields in the VM to allow us to still use offsets? I dunno.

Marwes:
I agree that it can be pretty confusing behavior but I also think that forfeiting the ability to only use offsets for field accesses is rather unfortunate as well. Having a deterministic order may overall be the best solution. It is something I have considered before but as I wasn't sufficiently annoyed by it I refrained from implementing it.

There is another solution which might work and that is if row polymorphism #92 is implemented. Then it might be possible to consider two distinct types of records `{ x : Int, y : String | r }` and `{ x : Int, y : String }` where the former works with no regard to the order of the fields while the latter would take order into account letting field accesses be done as just offsets. I only say it might work however because there essentially needs to be some sort of subtyping relationship between polymorphic and and non-polymorphic records (non-polymorphic records must be 'widened' to polymorphic if they are unufied) which may be tricky/impossible to properly to typecheck.

Marwes:
To get deterministic ordering it is enough to sort the fields by name when emitting code, it will make record construction slightly more complicated however.

brendanzab:
I guess it just feels very surprising from a user point of view. Maybe a bit of complexity in the implementation is worth it?

Marwes:
I don't want to put to much emphasis on the complexity of implementing it as it as omitting it for now has been more due to unwillingness to commit to fixing the order that way.

Thought of some other (potential) drawbacks from enforcing a specific order to solve this problem.

* Low level languages can use order to pack fields more tightly (probably not relevant for gluon but its worth mentioning).
* More specific for gluon maybe it can be used to group garbage collected and non garbage collected values to let mark phase proceed quicker. So if fields containing garbage collected values were put first then the record could have some information to tell the collector that only the first n values need to be scanned.

Probably avoiding surprises is better than potential performance improvements but still... It is something to consider at least.

Marwes:
[...]

PR #132, opened 2016-08-27
feat: Add row polymorphic records

### LINKED ISSUE #92: Add row polymorphism (aka. extensible records) to type system
This could be useful for adding a more powerful effect system, or perhaps for adding default implementations for interfaces/traits/modules/typeclasses (whatever we call them?).

## Resources

Feel free to add more if you like!

- [willtim/row-polymorphism](https://github.com/willtim/row-polymorphism) - Example row-polymorphism implementations 
- [Extensible records with scoped labels](https://www.microsoft.com/en-us/research/publication/extensible-records-with-scoped-labels/)
- [Koka: Programming with Row Polymorphic Effect Types](https://www.microsoft.com/en-us/research/publication/koka-programming-with-row-polymorphic-effect-types-2/)
- [Elm: Records](http://elm-lang.org/docs/records)
- [Elm: Simplified Records](http://elm-lang.org/blog/compilers-as-assistants#simplified-records)
- [PureScript: Handling Native Effects with the Eff Monad](http://www.purescript.org/learn/eff/)
- [PureScript Language Guide: Kind System](https://github.com/purescript/purescript/wiki/Language-Guide#kind-system) 
- [PureScript Language Guide: Kinds](https://github.com/purescript/purescript/wiki/Language-Guide#kinds)

Marwes:
This PR implements row polymorphic records which should make records much more flexible at the expense of some performance.

Left to do:

- [x] Add comments
- [x] Handle polymorphic records in the vm
- [x] Fix pretty printing of polymorphic records
- [x] Add tests to test row kinds specifically
- [x] Improve row polymorphism related error messages

Fixes #92


base/src/types.rs
```diff
@@ -106,8 +106,13 @@ pub enum Type<Id, T = AstType<Id>> {
     Record {
         /// The associated types of this record type
         types: Vec<Field<Id, Alias<Id, T>>>,
+        row: T,
+    },
+    EmptyRow,
```
brendanzab: Why do we need this? Is it not just an empty record?

brendanzab: Oh I see - `Record` has a `row` field now. I recon call these `EmptyRecord` and `ExtendRecord`

Marwes: `{ } : Type` but `EmptyRow : Row`. Slide 20 is good at untangling this https://www.google.se/url?q=https://www.cs.cmu.edu/~neelk/rows.pdf&sa=U&ved=0ahUKEwi_mIb61-LOAhXIkCwKHf-9BxAQFggOMAE&usg=AFQjCNHgZPdKT__UjF9Zr9Yr3Ax5FF2hog

Marwes: I am not to sold on the current names but I think they should not be called `Record` + something. I initially thought it odd that rows and records were separate in https://github.com/willtim/row-polymorphism but in the end it makes sense as rows are a 'part' of a record but is not a record itself (it even has a different kind).

brendanzab: I wonder if in the future we could have a `Record : row -> type` constructor?

brendanzab: [Elm's AST](https://github.com/elm-lang/elm-compiler/blob/master/src/Type/Type.hs#L43-L54):

```hs
data Term1 a
    = App1 a a
    | Fun1 a a
    | EmptyRecord1
    | Record1 (Map.Map String a) a


data TermN a
    = PlaceHolder String
    | AliasN Var.Canonical [(String, TermN a)] (TermN a)
    | VarN a
    | TermN (Term1 (TermN a))
```

brendanzab: I think there is a `TRecord Type` variant in [`AlgorithmW_Records`](https://github.com/willtim/row-polymorphism/blob/master/src/AlgorithmW_Records.hs#L48) because you do need a record constructor...

brendanzab: Another idea would be to have record construction be implicit, [like Elm does](https://github.com/elm-lang/elm-compiler/blob/master/src/Type/Type.hs#L46-L47).

Marwes: > I wonder if in the future we could have a Record : Row -> Type constructor?

I am not sure I follow, what does that give us over
```f#
Show = {
    t : Type,
    show : t -> String,
}
```

Marwes: > I think there is a TRecord Type variant in AlgorithmW_Records because you do need a record constructor. So we would need to switch to Type::Record(T).

That's already how the current representation is, only with 'type fields' packed into `Type::Record` as well since I didn't want to deal with polymorphic 'type fields' just yet.

> Another idea would be to have record construction be implicit, like Elm does.
Iit might be that the [TermN](https://github.com/elm-lang/elm-compiler/blob/master/src/Type/Type.hs#L54) constructor serves as `TRecord Type`. Without `TRecord : Row -> Type` I am not sure how to reconcile kind checking as with just say `EmptyRecord` and `ExtendRecord` they may have the kind `Row` or `Type` depending on context.

brendanzab: Ah, makes sense that it is temporary that there were two variants with fields on them - that was what was confusing me! Not sure if there is a hole in Elm's implementation there, but I think it would make sense to stick with the more traveled route of having a separate variant for `Row -> Type`.

base/tests/types.rs
```diff
@@ -173,3 +173,30 @@ fn show_kind() {
...
+                                                        typ: Alias::new("Test",
+                                                                        vec![Generic {
+                                                                                 kind: Kind::typ(),
+                                                                                 id: "a",
+                                                                             }],
+                                                                        Type::id("a")),
+                                                    }],
+                                               vec![],
+                                               Type::id("r"));
+    assert_eq!(format!("{}", typ), "{ Test a = a | r }");
```
brendanzab: I'd like to bikeshed this syntax at some stage! ^_^

tests/support.rs
```diff
@@ -0,0 +1,92 @@
+extern crate env_logger;
```
brendanzab: The idiom is to put this file at `support/mod.rs`. That way it doesn't try to compile it as a test.

base/tests/types.rs
Marwes: Definitely, just wanted something that was readable for now. Just copied elms syntax for now.

tests/support.rs
Marwes: So that's why the support module for `check` is in its own folder! I didn't know that, will fix!

base/tests/types.rs
brendanzab: Elm's is actually `{ r | Test a = a }`

brendanzab:
What semantics are we using for field shadowing?

Marwes:
Shadowing like?
```f#
{ x = 1, x = "" }
```
I have just assumed that should be an error.

Marwes:
This compiles without error which is pretty bad (at least in how I think it may interact with the current typechecker and vm which assume fields are unique). :sweat_smile: 
```f#
#[test]
fn double_field_declared() {
    let _ = ::env_logger::init();
    let text = r#"
{ x = 1, x = "" }
"#;
    assert!(result.is_err());
}
```

brendanzab:
I think in Elm the semantics are that if you re-specify a field it is the same as removing and adding a field. Not sure if that's what we want though.

Marwes:
That may make sense as well. I think 1ML has the same semantics given how it even desugars let bindings into modules/records (Page 3 bottom left https://www.mpi-sws.org/~rossberg/1ml/1ml-extended.pdf)

base/tests/types.rs
Marwes: Guess I misremembered, I will just leave it as is for the moment as it just needs to display the information for now.

Marwes:
The problem I have had with the vm part is that I wanted to keep that same compact representation as it is currently and only add a header which then points to a map off the field offsets.

```
+------------------------+
| Map<FieldName, Offset> |
+------------------------+
     ^
     |
+----+------------+------------------+----------------+
|                 |                  |                |
|  Header         |  Field 1         |     Field 2    |
+-----------------+------------------+----------------+
```
Unfortunately that map needs to be owned by something and I couldn't decide where. I think however that with #135 each thread can intern all necessary maps which will let all values owned by the thread take a pointer to each map.

check/src/typecheck.rs
```diff
@@ -1312,13 +1313,15 @@ impl<'a> Typecheck<'a> {
...
-                f(&field.0, &field.1, &associated_type.typ);
+    if let Type::Record { ref row, .. } = **typ {
+        if let Type::ExtendRow { fields: ref field_types, .. } = **row {
+            for field in fields {
+                // If the field in the pattern does not exist (undefined field error) then skip it as
+                // the error itself will already have been reported
+                if let Some(associated_type) = field_types.iter()
+                    .find(|type_field| type_field.name.name_eq(&field.0)) {
+                    f(&field.0, &field.1, &associated_type.typ);
+                }
```
brendanzab: Should we also check that `rest` is an empty row?

Marwes: I don't think we need to check that exactly but there is definitely a bug here since there could be a nested `ExtendRow { rest: ExtendRow { .. } }` and I only check the outermost row. Need to change to to use `Type::field_iter` as I use in most places.

check/src/unify_type.rs
```diff
@@ -318,7 +318,11 @@ fn find_alias_<'a, U>(unifier: &mut UnifierState<'a, U>,
                 if l_id == r_id {
                     // If the aliases matched before going through an alias there is no need to
                     // return a replacement type
-                    return Ok(if did_alias { Some(l.clone()) } else { None });
+                    return Ok(if did_alias {
+                        Some(l.clone())
+                    } else {
+                        None
+                    });
```
brendanzab: Odd... seems like rustfmt is formatting this differently for you :/

Marwes: Yeah, I am only on 0.5 because there is a bug in Rustycode preventing me from using it https://github.com/saviorisdead/RustyCode/issues/164. Hoped I would still be fine on 0.5 but I guess not. Going to see if I can setup rustfmt through the test.sh script instead

Marwes: Applying the patch and installing manually was easier than I thought so I can now use 0.6. Will reformat tomorrow.

Marwes:
Other than adding some more comments (let me know if there is something I should explain better) and running rustfmt I think this should be done. Let me know if you have any questions @brendanzab or thoughts on it!

brendanzab:
As usual I find your code pretty dense and monolithic, but I can't really give much constructive advice yet. I will try to think about it... I would like our codebase to be as nice at rustc is these days! :D Their comments are also incredibly helpful - I would look to them for inspiration!

Some tests for the new errors would be great. I understand it might be hard though given that we do not have syntax for extensible records yet. I would also look at Elm's errors for records - they are very helpful, showing a 'diff' of the expected and found record types. Perhaps for another time.

check/src/rename.rs
```diff
@@ -323,8 +321,11 @@ pub fn rename(symbols: &mut SymbolModule,
 }
 
 pub fn equivalent(env: &TypeEnv, actual: &ArcType, inferred: &ArcType) -> bool {
+    use substitution::Substitution;
+    // FIXME Inneficient and possible wrong
```
Marwes: Fix this FIXME

brendanzab: spelling! :D

Marwes: That as well :) But first I will try and actually fix it.

Marwes: Since nothing is actually allocated in `Substitution::new()` and I don't believe there should not be any bugs from having it there I just fixed the spelling. Its not in a hot path of the code so trying to optimize won't have an impact anyway.

Marwes:
Tried to expand a bit more in the comments but I am afraid I was not able clean up the code itself much. If you could point out some specifics that are hard to follow that would be helpful since I am afraid I am rather blind towards the problems :/.

This would be ready except I found a problem as I added the last doc commit.

```rust
#[test]
fn different_order() {
    let _ = env_logger::init();

    let text = r#"
if True then
    { x = 1, y = "" }
else
    { y = "", x = 1 }
"#;

    assert!(result.is_ok());
}
```

This passes (which is fine if #114 is fixed) but currently the records does have a different representation which will lead to runtime errors. I guess normalizing the field order is the best idea here? I could maybe add a special case in typechecking to still care about order for non-polymorphic records but maybe it is better to do it properly right away (though that complicates code generation grumble grumble...).

base/src/types.rs
```diff
@@ -533,6 +566,17 @@ impl<Id, T> Type<Id, T>
     }
 }
 
+pub trait TypeRef: Sized {
```
brendanzab: Why do we need an extra trait for this? Could we not just implement this method directly on `Type`?

base/tests/types.rs
```diff
@@ -171,3 +171,30 @@ fn show_kind() {
...
     assert_eq!(format!("{}", function_arg), "(Type -> Type) -> Type");
 }
+
+#[test]
+fn show_polymorphic_record() {
+    let typ: ArcType<&str> = Type::poly_record(vec![],
+                                               vec![Field {
+                                                        name: "x",
+                                                        typ: Type::string(),
+                                                    }],
```
brendanzab: Might be cleaner to assign these fields to intermediate variables.

check/src/unify_type.rs
```diff
@@ -68,6 +74,21 @@ impl<I> fmt::Display for TypeError<I>
...
+            TypeError::MissingFields(ref typ, ref fields) => {
+                try!(write!(f, "The type `{}` lacks the following fields: ", typ));
+                for (i, field) in fields.iter().enumerate() {
+                    let sep = if i == 0 {
+                        ""
+                    } else if i == fields.len() - 1 {
+                        " and "
+                    } else {
+                        ", "
+                    };
```
brendanzab: How about:

```rust
match i {
    0 => "",
    i if i == fields.len() - 1 => " and ",
    _ => ", ",
};
```

check/src/unify_type.rs
```diff
@@ -284,14 +411,14 @@ fn find_alias_<'a, U>(unifier: &mut UnifierState<'a, U>,
...
-/// // try_zip_alias(Float, Test 0) => Ok((Float, Test 0))
+/// // find_common_alias(Test2, Test 0) => Ok((Test String, Test 0))
+/// // find_common_alias(Float, Test 0) => Ok((Float, Test 0))
 /// ```
-fn try_zip_alias<'a, U>(unifier: &mut UnifierState<'a, U>,
-                        expected: &ArcType,
-                        actual: &ArcType,
-                        through_alias: &mut bool)
-                        -> Result<(ArcType, ArcType), ()>
+fn find_common_alias<'a, U>(unifier: &mut UnifierState<'a, U>,
```
brendanzab: Much better name!

brendanzab:
Added some comments, but will have to look at it again after work.

check/src/unify_type.rs
Marwes: I can never decide on the best way to write these lists. Might be useful to have some abstraction over 'listing things in a human readable way'.

base/src/types.rs
Marwes: It would need to be implemented on `ArcType` so that the sharing can be preserved. I suppose implementing it on `ArcType` is fine now but if we add more type representations a trait is the way to go to prevent duplication. Should I change it now?

#143

brendanzab: You could also just implement it as a free function. I guess I am just wary of traitifying too early if it is not yet needed, and the right abstractions are not yet clear.

check/src/unify_type.rs
brendanzab: I think itertools might have something for that, from memory.

brendanzab: Yep! [`Itertools::format_default`](https://docs.rs/itertools/0.4.18/itertools/trait.Itertools.html#method.format_default)

Marwes: `Itertools` does not seem to allow changing the separator however. Its nice to have `and` before the last element.

brendanzab: You could do: `let sep = if foo.len() == 1 { " and " } else { ", " };

Marwes:
Took the quick and easy way out and just enforced that when unifying two non-polymorphic records the field order matters. Its easy enough to remove and lets me skip out on deciding on and implementing #114.

r?

check/src/typecheck.rs
```diff
@@ -992,8 +997,7 @@ impl<'a> Typecheck<'a> {
     }
 
     fn kindcheck(&self, typ: &mut ArcType) -> TcResult<()> {
-        let subs = Substitution::new();
-        let mut check = super::kindcheck::KindCheck::new(&self.environment, &self.symbols, subs);
+        let mut check = super::kindcheck::KindCheck::new(&self.environment, &self.symbols);
```
brendanzab: Is `KindCheck` not already in scope?

check/src/typecheck.rs
```diff
@@ -1433,21 +1435,21 @@ fn primitive_type(op_type: &str) -> ArcType {
 /// extern crate gluon_check;
```
brendanzab: The 'Removes layers of `Type::App`.' comment is stale now.

check/src/completion.rs
```diff
@@ -98,14 +100,16 @@ impl<E: TypeEnv> OnFound for Suggest<E> {
...
-                for field in fields {
-                    if field.name.as_ref().starts_with(id) {
-                        self.result.push(Suggestion {
-                            name: field.name.declared_name().into(),
-                            typ: field.typ.clone(),
-                        });
+            if let Type::Record { ref row, .. } = *typ {
+                if let Type::ExtendRow { ref fields, .. } = **row {
+                    let id = ident.as_ref();
+                    for field in fields {
```
Marwes: Should change this to field_iter

check/src/typecheck.rs
```diff
@@ -727,9 +741,9 @@ impl<'a> Typecheck<'a> {
                 let mut match_type = self.remove_alias(match_type);
                 let mut types = Vec::new();
                 let new_type = match *match_type {
```
Marwes: Pretty sure this match block can be simplified now that row polymorphism is implemented. The reason for the split is just because omitting fields from patterns was allowed previously (and still will be) but since `let { x } = .. ` would look like ``{ x : a }` it could not simply be unified with, say `{ x : Int, y : Int }` but now it can simply be unified with a polymorphic record.

vm/src/compiler.rs
```diff
@@ -23,6 +23,12 @@ pub enum Variable<G> {
     UpVar(VmIndex),
 }
 
+/// Field accesses on records can either be by offset on a non-polymorphic record, or
```
Marwes: Unfinished

check/src/typecheck.rs
Marwes: For a reason have I have not been able to figure out (yet) I got a type error from the use of `make_Applicative` in the prelude. Added a minimized version of the issue to `pass.rs` but I have not been able to figure out a solution yet. Without the simplification it still works however.

Marwes:
Fixed the nits and threw in polymorphism over type fields as well which as it didn't seem to tricky. There is probably some more simplifications possible but that could be done later (assuming no other nits appear from these changes :) ).

base/src/types.rs
```diff
@@ -156,6 +154,13 @@ impl<Id> ArcType<Id> {
         ArcType { typ: Arc::new(typ) }
     }
 
+    pub fn type_field_iter(&self) -> TypeFieldIterator<Self> {
```
brendanzab: `fields` is more rustic :3

brendanzab:
Let's just merge this! :)

Marwes:
:tada:

PR #135, opened 2016-08-31
perf: Use a single mutex for both the stack and gc

Marwes:
This gives a noticeable performance improvement for code which frequently enters and leaves the vm itself as only one mutex need to be locked. It does complicate some of the surrounding code slightly but this will be a necessary change regardless since the now added `Contex` type can be extended with additional fields which would otherwise involve adding more locks.

Before
```
test factorial                    ... bench:     249,725 ns/iter (+/- 12,150)
test factorial_tail_call          ... bench:     242,908 ns/iter (+/- 13,685)
test gluon_rust_boundary_overhead ... bench:     396,684 ns/iter (+/- 34,313)
```

After
```
test factorial                    ... bench:     128,571 ns/iter (+/- 7,537)
test factorial_tail_call          ... bench:     125,914 ns/iter (+/- 6,374)
test gluon_rust_boundary_overhead ... bench:     237,173 ns/iter (+/- 11,306)
```

### DIFF
--- a/benches/function_call.rs
+++ b/benches/function_call.rs
@@ -5,7 +5,8 @@ extern crate test;
 extern crate gluon;
 
 use gluon::{Compiler, new_vm};
-use gluon::vm::api::FunctionRef;
+use gluon::vm::thread::{Status, Thread};
+use gluon::vm::api::{FunctionRef, primitive};
 
 // Benchmarks function calls
 #[bench]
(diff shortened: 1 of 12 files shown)


Marwes:
I am actually not entirely sure where what locking I removed to speedup the factorial benchmarks so much...

Marwes:
Going to go ahead and merge this after travis passes since I have already based work in #132 on this. It can always be prettied up later.

brendanzab:
:+1:

Marwes:
Heh, actually got an error message saying that I could not merge this... because it was already merged :P

PR #136, opened 2016-09-01
Normalize identifier variant names

brendanzab:


brendanzab:
derpy skeptic is derpy

PR #137, opened 2016-09-03
Ast cleanups

brendanzab:


rustfmt.toml
```diff
@@ -1 +1,3 @@
+match_block_trailing_comma = true
 newline_style = "Native"
+where_trailing_comma = true
```
brendanzab: These are some of my preferences - you might disagree though...

brendanzab:
Sorry for the enormity of this - just wanted to get it all out of the way!

rustfmt.toml
Marwes: I am a bit wary about the trailing comma on match arms. When writing code I omit it but since rustfmt takes care of adding it anyway I guess I don't really mind it. It does make things more consistent.

Marwes:
Looks good. I guess you probably should merge this sooner rather than later since it is likely to conflict with most other PRs

PR #140, opened 2016-09-03
Switch ids to types in AST where possible

brendanzab:


base/src/ast.rs
```diff
@@ -274,12 +272,17 @@ pub struct ValueBinding<Id: AstId> {
...
+
     fn visit_expr(&mut self, e: &mut SpannedExpr<Self::T>) {
         walk_mut_expr(self, e);
     }
+
     fn visit_pattern(&mut self, e: &mut SpannedPattern<Self::T>) {
         walk_mut_pattern(self, &mut e.value);
     }
+
+    fn visit_typ(&mut self, _: &mut AstType<<Self::T as AstId>::Untyped>) {}
```
brendanzab: Not sure if this is the best solution :/

parser/src/lib.rs
```diff
@@ -517,7 +517,7 @@ impl<'input, I, Id, F> ParserEnv<I, F>
                                   sep_end_by(self.expr(), token(Token::Comma)))
                          .map(|exprs| {
                              loc(Expr::Array(Array {
-                                 id: self.empty_id.clone(),
+                                 typ: Type::hole(),
```
brendanzab: Haven't figured out how to cache these yet. I needed to add a `hole_typ: AstType<<F::Ident as AstId>::Untyped>` field to `ParserEnv`, but `F::Ident` is not known to be an `AstId`. Any suggestions?

vm/src/compiler.rs
```diff
@@ -664,7 +665,9 @@ impl<'a> Compiler<'a> {
                 function.emit(ConstructArray(a.expressions.len() as VmIndex));
             }
             Expr::Lambda(ref lambda) => {
-                let (function_index, vars, cf) = try!(self.compile_lambda(&lambda.id,
+                let name = self.symbols.symbol("lambda");
```
brendanzab: We lose the line number here... not sure how much of a problem that is.

parser/src/lib.rs
Marwes: Store a `Type::hole()` in the parser instance instead and clone that (just like `empty_id`).

vm/src/compiler.rs
```diff
@@ -531,13 +531,14 @@ impl<'a> Compiler<'a> {
                             ast::Pattern::Ident(ref name) => name,
                             _ => panic!("Lambda binds to non identifer pattern"),
                         };
-                        let (function_index, vars, cf) = try!(self.compile_lambda(name,
-                                                                             &bind.arguments,
-                                                                             &bind.expression,
-                                                                             function));
+                        let (fn_index, vars, cf) = try!(self.compile_lambda(&name.id(),
```
Marwes: `s/name/typ`? (or `function_type` or something)

base/src/ast.rs
Marwes: Yah this seems a bit wonky with the current visitor. Probably the best way of doing it though.

base/src/ast.rs
```diff
@@ -274,12 +272,17 @@ pub struct ValueBinding<Id: AstId> {
...
+
     fn visit_identifier(&mut self, _: &mut Self::T) {}
```
Marwes: Maybe forward `visit_identifier` to `visit_typ`? Should simplify the use in `typecheck.rs`

base/src/ast.rs
Marwes: Should we name it `visit_type` instead? `typ` is only needed because `type` is a keyword so when it is part of a larger identifier it is not needed.

vm/src/compiler.rs
Marwes: Maybe keep the `Id` for `Lambda` as it actually serves a purpose there. Having a name for each function is useful in stacktraces.

parser/src/lib.rs
Marwes: Tack on the bound maybe? Or does that not work for some reason?

base/src/ast.rs
brendanzab: I kind of prefer being consistent between casing. I think in rustc they just use `ty` everywhere if it's lower case... dunno.

base/src/ast.rs
brendanzab: How would I get the `AstType` from here?

parser/src/lib.rs
brendanzab: What do you mean by 'parser instance' here?

base/src/ast.rs
Marwes: Yes you are right that wont work. Never mind then.

brendanzab: 😝

base/src/ast.rs
Marwes: Fair enough. As an aside, I use `typ` in part because its `type` in swedish so it seemed rather fitting. Its also nice that it is one letter more than `ty` making it stand out a bit more.

parser/src/lib.rs
Marwes: The `Self` type ie `ParserEnv` which now has `empty_id`

PR #141, opened 2016-09-04
Use holes instead of option in value bindings

brendanzab:


brendanzab:
So this breaks typechecking. Which is interesting. Evidently we aren't inferring all of our holes eagerly enough...

Marwes:
I think its just that holes are not handled in unification at the moment. Check unify_type.rs, I think there just need to be a case handling `Type::hole` there (which returns a fresh type variable).

brendanzab:
So there is `try_match` and `do_zip_match` in `unify_type.rs`. Which one would I be looking at?

brendanzab:
I tried adding these cases in `try_match`:

```rust
            (&Type::Hole, &Type::Hole) => Ok(Some(subs.new_var())),
            (_, &Type::Hole) => Ok(Some(l.clone())),
            (&Type::Hole, _) => Ok(Some(r.clone())),
```

Didn't seem to work :/

Marwes:
That is what I was going to suggest... :( Though that shoud work

brendanzab:
Woops - left the generated `grammar.rs` from my lalrpop experiments in there. Had to rebase.

Marwes:
Pretty sure the remaining errors is because after the call to `merge_signature` in typecheck.rs the resulting type is not assigned back to the type stored in the `AST`. So then later on `unify` is called on a `Type::Hole` and that does not have any handling of holes (nor should it). If you get what change is necessary then please implemeent it, otherwise I can send a PR to this branch :)

brendanzab:
Yeah, I think send a PR or push directly to this branch.

Marwes:
Oh yeah, I can just push directly, forgot that!

Marwes:
Turns out I was on the wrong track with adding `Type::Hole` cases to unification. The correct way (I believe) is to eagerly instantiate ``Type::Hole` to ensure that holes never enter unification in the first place. I also needed to write the type back into the AST to ensure that the type after typechecking is correct.

check/src/kindcheck.rs
```diff
@@ -157,7 +157,7 @@ impl<'a> KindCheck<'a> {
                 gen.kind = try!(self.find(&gen.id));
                 Ok((gen.kind.clone(), Type::generic(gen)))
             }
-            Type::Variable(_) => panic!("kindcheck called on variable"),
+            Type::Variable(_) => Ok((self.subs.new_var(), typ.clone())),
```
Marwes: This was just a panic before since variables should never have been kindchecked on. Now however holes will have been replaced by variables before kindchecking so these must be of 'any' kind just as holes are.

check/src/typecheck.rs
```diff
@@ -1165,7 +1168,11 @@ impl<'a> Typecheck<'a> {
         typ
     }
 
-    fn refresh_symbols_in_type(&mut self, typ: TcType) -> TcType {
+    // Replaces `Type::Id` types with the actual `Type::Alias` type it refers to
+    // Replaces variant names with the actual symbol they should refer to
+    // Instantiates Type::Hole with a fresh type variable to ensure the hole only ever refers to a
+    // single type variable
+    fn create_unifiable_signature(&mut self, typ: TcType) -> TcType {
```
brendanzab: Yay, docs :)

brendanzab:
I can handle merging master back into this branch if you like

Marwes:
Yeah. Go ahead.

PR #142, opened 2016-09-04
Flatten type aliases

brendanzab:


Marwes:
There is some motivation to keeping an alias over types which is that it hides the implementation at least a bit which could let use switch it out with less churn. Been thinking that maybe it could be faster to avoid using an `Arc` based type when doing typechecking and only use `Arc` for types stored in the vm. On the other hand that makes things more complex and forces some additional overhead to typechecking since `ArcType` would have to be translated to the type used in typechecking...

I am rambling a bit I guess. `AstType` and `TcType` is not exactly good names anyway so I'd say this is a change for the better regardless.

brendanzab:
Heh, I have a tendency to hack and slash at layers of stuff as much as I can so that I can understand things better. :D I would advocate for holding off on a new name until we actually need it, but feel free to disagree!

Switching from `Arc` would be quite nice actually. Perhaps make an issue for that?

Marwes:
There should be an issue for that but in the vm it is required or else it cannot be `Send + Sync`.

Marwes:
Types used to be `Rc` but I changed it when ensuring the vm implemented `Send + Sync`. Benchmarked it and didn't see a noticeable change at least.

brendanzab:
cc. #143

Marwes:
Yeah, guess we can merge this, unless there is anything else?

PR #146, opened 2016-09-05
Remove AstId trait

brendanzab:

PR #147, opened 2016-09-06
fix(parser): Instantiate a specialized version of parse in the parser crate

Marwes:
Due to the long compile times of the current parser a specialized version used by the rest of the crates is compiled into the parser crate itself letting the other crates remain unaffected.


Marwes:
Unfortunately when I skimmed #146 I didn't notice that it removed a specialized version of parse which was needed to keep the compile times in check. This is just a quick fix to bring them back to normal.

PR #148, opened 2016-09-06
refactor: Rename MaybeError to RuntimeResult

### LINKED ISSUE #104: Rename MaybeError to something better
[MaybeError](https://github.com/Marwes/gluon/blob/0987a3a3d11d8ff6f61f7160541311dbe8b86278/vm/src/api.rs#L775) is used to represent results for functions which may 'panic' (in the context of gluon not Rust). `MaybeError` is a really terrible name for this however so renaming it is a good idea.

Possible names:

```rust
enum Panicking<T, E> { }
enum Panic<T, E> { }
enum Diverging<T, E> { }
// More?
```

Let the bikeshedding comence!

Marwes:
Fixes #104

### DIFF
--- a/vm/src/api.rs
+++ b/vm/src/api.rs
@@ -776,22 +776,22 @@ impl<'vm, T: Getable<'vm>, E: Getable<'vm>> Getable<'vm> for StdResult<T, E> {
     }
 }
 
-pub enum MaybeError<T, E> {
-    Ok(T),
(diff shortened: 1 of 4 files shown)


Marwes:
Since `Err` was already called `Panic` I think it is a bit better to call `Ok`, `Return` to keep it completely distinct from `std::result::Result`.

PR #150, opened 2016-09-07
Assorted cleanups to base crate

brendanzab:


base/src/types.rs
```diff
@@ -183,6 +173,7 @@ impl<Id: AsRef<str>> fmt::Display for RcType<Id> {
 
 impl<Id> Deref for RcType<Id> {
```
Marwes: `RcType` is not used anymore so I think we can just remove it completely.

Marwes:
Looks good to me

PR #152, opened 2016-09-09
chore: Only test using skeptic on stable when running in travis

### LINKED ISSUE #149: Fix skeptic build errors
My best guess towards the errors caused by skeptic is that it is due to some conflicts between libraries compiled in stable/beta/nightly which end up getting linked in somehow so I want to try only running skeptic tests on travis with stable (probably using cargo feature flags).

Marwes:
This solution feels a bit heavy handed but at least it seems like skeptic no longer fails on travis from the 5 or so times I reran it.

This also enables beta testing again on travis since the rustc fixes has been backported.

Fixes #149

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
     - libdw-dev
 rust:
 - nightly
-# - beta
+- beta
 - stable
 before_script:
 - |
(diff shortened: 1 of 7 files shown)


Marwes:
Don't merge this yet, want to reorder the dependencies in Cargo.toml before this can be merged.

Marwes:
r?

PR #153, opened 2016-09-09
fix: Don't return None from Source::location when byte is at end of file

Marwes:
Since spans are exclusive they may refer to indexes one past the end.

### DIFF
--- a/base/src/error.rs
+++ b/base/src/error.rs
@@ -54,9 +54,9 @@ struct SourceContext<E> {
 
 impl<E> SourceContext<E> {
     fn new(source: &Source, error: Spanned<E, BytePos>) -> SourceContext<E> {
-        let start = source.location(error.span.start).unwrap();
-        let end = source.location(error.span.end).unwrap();
-        let (_, line) = source.line_at_byte(error.span.start).unwrap();
+        let start = source.location(error.span.start).expect("SourceContext: start location");
+        let end = source.location(error.span.end).expect("SourceContext: end location");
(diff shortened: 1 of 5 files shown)


Marwes:
Only changed the `unwrap`s into `expect` to give some context at least. Is there a better way of handling this instead of panicking? Even if we return `Option`/`Result` I am not sure what we can actually do with that information with the current use of `Infile`.

cc @brendanzab

brendanzab:
The solution here seems kind of ugly... will have to think :/

brendanzab:
Why would we ever get a location of of `src.len()`? Can we fix that instead?

brendanzab:
When exactly does this bug show up?

Marwes:
You can see it by just typing any variable into the repl `foo`. That will give an undefined variable error and panic. We get a position at `src.len()` because the spans are exclusive.

Marwes:
https://gitter.im/Marwes/gluon?at=57d29742c475164703c0e9d3

brendanzab:
Hmm, I wonder how Rust handles this case :/

Marwes:
Best I can tell from https://github.com/rust-lang/rust/blob/master/src/libsyntax/codemap.rs, Rust uses exclusive spans and so allows spans that end at `src.len()`. I think that this method will return a line + row position which may be at `src.len()` as well https://github.com/rust-lang/rust/blob/c87ba3f1222ba20d491e8ed76a04977283280742/src/libsyntax/codemap.rs#L315. Might be wrong though.

Marwes:
@brendanzab Have you been able to come up with a better solution for this?

PR #154, opened 2016-09-10
fix: Issues 60 and 134

Marwes:


base/src/types.rs
```diff
@@ -800,6 +800,14 @@ impl<'a, I, T, E> DisplayType<'a, I, T, E>
     fn pretty(&self, arena: &'a Arena<'a>) -> DocBuilder<'a, Arena<'a>>
         where I: AsRef<str>,
     {
+        fn ident<'b>(arena: &'b Arena<'b>, name: &'b str) -> DocBuilder<'b, Arena<'b>> {
+            if name.starts_with(is_operator_char) {
```
brendanzab: Would prefer `ast::is_operator_char` here

check/src/typecheck.rs
```diff
@@ -42,6 +43,8 @@ pub enum TypeError<I> {
     Rename(::rename::RenameError),
     /// Multiple types were declared with the same name in the same expression
     DuplicateTypeDefinition(I),
+    /// A field were defined more than once in a record constructor or pattern match
```
brendanzab: "A field that was defined"

check/src/typecheck.rs
```diff
@@ -642,46 +649,53 @@ impl<'a> Typecheck<'a> {
...
+                        });
+                    }
+                }
+
+                let mut new_fields: Vec<Field<_, _>> = Vec::with_capacity(fields.len());
+                for field in fields {
+                    let typ = try!(match field.1 {
+                        Some(ref mut expr) => Ok(self.typecheck(expr)),
+                        None => self.find(&field.0),
+                    });
```
brendanzab: Would this work?

```rust
let typ = match field.1 {
    Some(ref mut expr) => self.typecheck(expr),
    None => try!(self.find(&field.0)),
};
```

check/src/typecheck.rs
Marwes: `"A field that was defined more than once in a record constructor or pattern match" does not seem quite right?

brendanzab: Oh woops: "A field was defined more than once in a record constructor or pattern match"

Marwes: Aha, 'were' vs' was. damn you proper English

PR #155, opened 2016-09-12
Identifier cleanups 

brendanzab:
Just a bunch of stuff I found while muddling around on the parser. I also reordered the type module somewhat because I kept getting lost due to things not being grouped together correctly.

### DIFF
--- a/base/src/types.rs
+++ b/base/src/types.rs
@@ -13,11 +13,11 @@ use symbol::{Symbol, SymbolRef};
 /// Trait for values which contains kinded values which can be refered by name
 pub trait KindEnv {
     /// Returns the kind of the type `type_name`
-    fn find_kind(&self, type_name: &SymbolRef) -> Option<RcKind>;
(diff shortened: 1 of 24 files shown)


base/src/types.rs
```diff
@@ -386,6 +289,45 @@ pub struct Field<Id, T = ArcType<Id>> {
...
+    Variants(Vec<(Id, T)>),
+    /// Representation for type variables
+    Variable(TypeVariable),
+    /// Variant for "generic" variables. These occur in signatures as lowercase identifers `a`, `b`
+    /// etc and are what unbound type variables are eventually made into.
+    Generic(Generic<Id>),
+    /// A builtin type
+    Builtin(BuiltinType),
+    /// A record type
+    Record(T),
```
Marwes: Now you aren't even cleaning up legacy code, just my laziness/unawareness (after factoring out the types field) :stuck_out_tongue:.

PR #159, opened 2016-09-17
Line and column numbering improvements

brendanzab:
Improving the the implementation of this by moving to 0-based indexing. Closes #153

### DIFF
--- a/base/src/pos.rs
+++ b/base/src/pos.rs
@@ -66,11 +66,22 @@ impl fmt::Display for BytePos {
 }
 
 pos_struct! {
-    /// A character position - usually used for column offsets
-    pub struct CharPos(usize);
+    /// A column number, indexed from `1`
+    pub struct Column(usize);
 }
 
-impl fmt::Display for CharPos {
+impl fmt::Display for Column {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+pos_struct! {
+    /// A line number, indexed from `1`
+    pub struct Line(usize);
+}
+
+impl fmt::Display for Line {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         self.0.fmt(f)
     }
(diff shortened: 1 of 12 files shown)


Marwes:
Pretty sure the errors is because the locations in `lexer.rs` assume that columns start at 1, so probably it will work if you change them to start at 0.

Marwes:
The problem is that the layout algorithm does not view EOF as a token which is always further to the left than any other token https://github.com/brendanzab/gluon/blob/f2f02881be6ce761779ea4781f72446126912e5c/parser/src/lexer.rs#L693. It has to be viewed as always further to the left since it needs to close the top-most block in the expression. Probably changing the column position to be signed and changing the linked position above to be -1 instead is the simplest and cleanest solution.

base/src/pos.rs
```diff
@@ -56,76 +65,48 @@ impl fmt::Display for BytePos {
...
     }
 }
 
-impl SubAssign for CharPos {
-    fn sub_assign(&mut self, rhs: CharPos) {
-        self.0 -= rhs.0;
-    }
+pos_struct! {
+    /// A `0`-indexed line number, displayed externally as if it were offset from `1`.
+    pub struct Line(usize);
```
Marwes: No reason to use `usize` here if `BytePos` is represented by a `u32`.

base/src/pos.rs
```diff
@@ -56,76 +65,48 @@ impl fmt::Display for BytePos {
...
-    }
-}
-
-impl AddAssign for CharPos {
-    fn add_assign(&mut self, rhs: CharPos) {
-        self.0 += rhs.0;
-    }
+pos_struct! {
+    /// A `0`-indexed column number, displayed externally as if it were offset from `1`.
+    pub struct Column(usize);
```
Marwes: This could also be a `u32`

Marwes: Or an i32 to make it signed

brendanzab:
Updated!

parser/src/lexer.rs
```diff
@@ -29,6 +29,11 @@ impl<I> StreamOnce for LocatedStream<I>
             .uncons()
             .map(|ch| {
                 self.location.bump(ch);
+                // HACK: The layout algorithm expects `1` indexing for columns -
+                // this could be altered in the future though
+                if self.location.column == Column::from(0) {
+                    self.location.column = Column::from(1);
+                }
```
Marwes: Should this not add `+1` instead?

Might be an issue if the indentation is only a by a single space since the first and second column will look like they are the same.
```f#
f 1
 2
// Is parsed like
f 1
2
// instead of
f 1 2
```

brendanzab: Good catch!

brendanzab: Fixed

Marwes:
Great work! Merged!

Marwes:
Huh, you can close PRs as well use `Closes/Fixes #x`, got confused about where #153 went.

brendanzab:
Yep :)

PR #160, opened 2016-09-17
feat(vm): Add a hook function which is executed on every closure call

Marwes:
This is added to work in a similiar way to the debug hook in Lua but is currently very basic. In its current form it is still usable as a means to prevent infinite loops by adding a timeout which is checked in the hook


Marwes:
Adding this  because I need a way to prevent infinite loops in https://github.com/Marwes/gluon-webserver which should be the last piece missing for sandboxing

Marwes:
I should probably clarify that the local changes are to https://github.com/Marwes/gluon-webserver, more specifically to block loading of functions such as `io.readfile`

@brendanzab Any thoughts on this? It is really basic for now but I think we want something like this in any case and it should be easy to add finer granularity on top of this.

PR #163, opened 2016-09-21
feat: Extend the macro system to support import erroring on cyclic dependencies

### LINKED ISSUE #93: Make the compiler_pipeline module easier to use
The [compiler_pipeline](https://github.com/Marwes/gluon/blob/3ab3ab46688673c92fe5957752927c36e5938988/src/lib.rs#L89) module is intended to make it easy to run only part of the entire compilation procedure without having to worry about missing a step. So if a `&str` should be typechecked the module takes care of running `parse -> macro expand -> typecheck` but for an already macro expanded expression the same overloaded ``typecheck` method could be called on a `MacroValue` and only the typechecking part would run.

While the traits up to and including `Typecheckable` make sense the `Compilable` and `Executable` traits are not as easy to use due to them being parameterised which makes them much trickier to use which should be fixed so that these traits are more useable.

### LINKED ISSUE #99: Rewrite the macro system
The current macro system is quite limited and the only macro which currently exist is `import`. Even just for that macro the current implementation is inadequate. Thus a rewrite (or even removal with `import` moved to be known at the language level).

### Issues with import

* Does not handle cyclic imports anymore (not that hard to fix but might as well be done in the rewrite)

* An imported module will be loaded and executed even if the module importing it fails typechecking. This behavior may actually be fine but I think it is worth mentioning, the alternative is to only typecheck all modules before loading and executing them.

* Buggy when multiple threads try to load modules simultaneously

* Cannot support parallel loading (from a single thread) of modules if that is desired

### Sketch

If the macro system is rewritten this is my current sketch for the idea.

```rust
pub trait Macro: ::mopa::Any + Send + Sync {
    fn expand(&self,
              /// `&mut MacroExpander` to allow macros to store state between calls
              env: &mut MacroExpander,
              arguments: &mut [ast::LExpr<TcIdent>])
              -> Result<ast::LExpr<TcIdent>, Error>;
    fn clone(&self) -> Box<Macro>;
}

pub struct MacroExpander<'a> {
    pub env: &'a Thread,
    pub macros: &'a MacroEnv,
    /// Allow macros to be defined locally so that they can be overridden temporarily
    pub local_macros: HashMap<String, Box<Macro>>,
    errors: Errors<Error>,
    /// Some sort of generic state storage to let each macro store some state
}
```
With something like this it should (at least) support the `import` macro since by using the same `MacroExpander` for resolving all recursive imports the `import` macro can store the names of each recursively imported module to resolve cyclic references.

Marwes:
This implements the change to the macro system in #99 (with a few tweaks). The main change is that macros can now store some state which in the case of the import macro, allows it to store a stack of modules currently being imported. In addition to this #93 closed as I found it necessary to have more control over how imported modules are compiled (in this case it was necessary to reuse the `MacroExpander` for all recursive expansions of `import`).

While this makes the `compiler_pipeline` module at least usable, there is still some work to make it easy to use. I don't think that is of immediate concern though and I don't have any concrete ideas how so I think it is fine to close that issue as well as that can probably only be discovered through further use of it.

Closes #99
Closes #93


Marwes:
This could probably do with a few more comments. Will add that sometime tonight.

check/src/typecheck.rs
```diff
@@ -1045,9 +1045,10 @@ impl<'a> Typecheck<'a> {
                 self.environment.stack.get_all(symbol).expect("Symbol is not in scope");
             if existing_types.len() >= 2 {
                 let existing_type = &existing_types[existing_types.len() - 2];
-                debug!("Intersect\n{} <> {}",
-                       types::display_type(&self.symbols, existing_type),
-                       types::display_type(&self.symbols, symbol_type));
+                debug!("Intersect `{}`\n{} <> {}",
```
brendanzab: `{} ∩ {}`! 😄

src/import.rs
```diff
@@ -77,10 +89,16 @@ impl CheckImporter {
...
 impl Importer for CheckImporter {
-    fn import(&self, vm: &Thread, modulename: &str, input: &str) -> Result<(), MacroError> {
+    fn import(&self,
+              compiler: &mut Compiler,
+              vm: &Thread,
+              modulename: &str,
+              input: &str,
+              expr: SpannedExpr<Symbol>)
+              -> Result<(), MacroError> {
         use compiler_pipeline::*;
```
brendanzab: Add a space after the import to make the code a little airier.

src/import.rs
```diff
@@ -77,10 +89,16 @@ impl CheckImporter {
     }
 }
 impl Importer for CheckImporter {
-    fn import(&self, vm: &Thread, modulename: &str, input: &str) -> Result<(), MacroError> {
+    fn import(&self,
+              compiler: &mut Compiler,
+              vm: &Thread,
+              modulename: &str,
```
brendanzab: snake_casify this

src/import.rs
```diff
@@ -77,10 +89,16 @@ impl CheckImporter {
...
+              input: &str,
+              expr: SpannedExpr<Symbol>)
+              -> Result<(), MacroError> {
         use compiler_pipeline::*;
-        let mut compiler = Compiler::new().implicit_prelude(modulename != "std.types");
-        let TypecheckValue(expr, typ) = try!(input.typecheck(&mut compiler, vm, modulename, input));
+        let TypecheckValue { expr, typ } = try!(MacroValue { expr: expr }
+            .typecheck(compiler, vm, modulename, input));
```
brendanzab: Perhaps break this up?

```rust
let macro_value = MacroValue { expr: expr };
let typecheck_value = try!(macro_value.typecheck(compiler, vm, modulename, input));
```

Marwes:
Don't worry, I am in no rush.

src/lib.rs
```diff
@@ -125,64 +124,104 @@ pub struct Compiler {
...
 /// Advanced compiler pipeline which ensures that the compilation phases are run in order even if
-/// not the entire compilation procedure is needed
+/// not the entire compilation procedure is needed.
+///
+/// Each trait in this module represents a stage in a full compilation so to only run compilation
+/// up until and including typechecking the `Typecheckable` trait can be used. Furthermore, if
+/// compilation should continue at some point after typechecking has succeeded, the result of
+/// typechecking (`TypecheckValue`) can be used as input to the next stage, ensuring that it is
+/// difficult to forget a stage.
 pub mod compiler_pipeline {
```
brendanzab: Wondering if we should break this out into a submodule. Not needed in this commit though.

src/lib.rs
```diff
@@ -191,83 +230,133 @@ pub mod compiler_pipeline {
...
-    pub struct CompileValue(pub SpannedExpr<Symbol>, pub ArcType, pub CompiledFunction);
+    /// Result of successful compilation
+    pub struct CompileValue<E> {
+        pub expr: E,
+        pub typ: ArcType,
+        pub function: CompiledFunction,
+    }
 
     pub trait Compileable<Extra> {
+        type Expr;
```
brendanzab: Inconsistent new-line spacing here. I would add one between the type and the following function.

src/lib.rs
Marwes: Yeah its grown a bit, might as well break it out now.

PR #166, opened 2016-09-24
feat(vm): Fix and cleanup the run_expr method

Marwes:
See individual commits.

### DIFF
--- a/tests/io.rs
+++ b/tests/io.rs
@@ -2,6 +2,7 @@ extern crate gluon;
 extern crate env_logger;
 
 use gluon::{Compiler, new_vm};
+use gluon::vm::api::IO;
 
 #[test]
 fn read_file() {
@@ -21,10 +22,11 @@ fn read_file() {
             assert (array.index bytes 1 #Byte== 112b) // p
             pure (array.index bytes 8)
         "#;
-    let result = Compiler::new().run_io_expr::<u8>(&thread, "<top>", text);
+    let result = Compiler::new().run_io_expr::<IO<u8>>(&thread, "<top>", text);
 
     match result {
-        Ok((value, _)) => assert_eq!(value, b']'),
+        Ok((IO::Value(value), _)) => assert_eq!(value, b']'),
+        Ok((IO::Exception(err), _)) => assert!(false, "{}", err),
         Err(err) => assert!(false, "{}", err),
     }
 }
--- a/tests/vm.rs
+++ b/tests/vm.rs
@@ -6,8 +6,7 @@ mod support;
 
 use support::*;
 
-use gluon::vm::api::generic::A;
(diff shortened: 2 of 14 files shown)


vm/src/thread.rs
```diff
@@ -560,7 +564,11 @@ impl ThreadInternal for Thread {
 
 
     /// Calls a module, allowed to to run IO expressions
-    fn call_module(&self, typ: &ArcType, closure: GcPtr<ClosureData>) -> Result<Value> {
+    fn call_module(&self,
+                   typ: &ArcType,
+                   closure: GcPtr<ClosureData>,
+                   run_io: bool)
+                   -> Result<Value> {
```
brendanzab: It might be better to have two functions, rather than the boolean flag. `call_module` and `call_io_module`?

Marwes: You are right, I was a bit sloppy there since I figured it is just an internal function, not much more work to do it properly though. 😓

Marwes:
What in particular do you not understand? (so I can add some comments to elaborate)

tests/io.rs
```diff
@@ -21,10 +22,11 @@ fn read_file() {
             assert (array.index bytes 1 #Byte== 112b) // p
             pure (array.index bytes 8)
         "#;
-    let result = Compiler::new().run_io_expr::<u8>(&thread, "<top>", text);
+    let result = Compiler::new().run_io_expr::<IO<u8>>(&thread, "<top>", text);
```
brendanzab: :+1:

tests/vm.rs
```diff
@@ -557,10 +559,14 @@ fn run_expr_int() {
 
     let text = r#"io.run_expr "123" "#;
     let mut vm = make_vm();
-    let (result, _) = Compiler::new().run_io_expr::<String>(&mut vm, "<top>", text).unwrap();
-    let expected = "123 : Int";
-
-    assert_eq!(result, expected);
+    let (result, _) = Compiler::new().run_io_expr::<IO<String>>(&mut vm, "<top>", text).unwrap();
```
brendanzab: Isn't the `IO` redundant here? What else could an `io_expr` return?

Marwes: Should make this clearer, it can still run non `IO` expressions, but if it is `IO` then the action is run as well. Might be a better API than this though

brendanzab: Is there a reason why we need to do that kind of internal dispatch as opposed to just having two functions?

Marwes: Not really, its just a legacy quirk. Never had a reason not to run `IO` before and the current way of inspecting and running it is more re less a hack I guess. I think I will try and just have `run_expr` and then a secondary function which can run the `IO` action that is returned.

PR #168, opened 2016-09-26
chore: Disable skeptic tests on travis

Marwes:
Getting failures every single PR now so until skeptic can be made to work on travis I think we need to disable it completely.


brendanzab:
:(

PR #176, opened 2016-10-06
[WIP] feat: Add source maps

Marwes:

### DIFF
--- a/src/compiler_pipeline.rs
+++ b/src/compiler_pipeline.rs
@@ -13,6 +13,7 @@ use std::borrow::{Borrow, BorrowMut};
 use base::ast::SpannedExpr;
 use base::error::InFile;
 use base::types::ArcType;
+use base::source::Source;
 use base::symbol::{Name, NameBuf, Symbol, SymbolModule};
 
 use vm::compiler::CompiledFunction;
@@ -169,10 +170,11 @@ pub trait Compileable<Extra> {
                compiler: &mut Compiler,
                thread: &Thread,
                file: &str,
+               expr_str: &str,
                arg: Extra)
                -> Result<CompileValue<Self::Expr>>;
 }
-impl<'a, 'b, T> Compileable<(&'a str, Option<&'b ArcType>)> for T
+impl<'a, 'b, T> Compileable<Option<&'b ArcType>> for T
     where T: Typecheckable,
 {
     type Expr = T::Expr;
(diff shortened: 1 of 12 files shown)


vm/src/compiler.rs
```diff
@@ -312,14 +322,16 @@ impl<'a, T: CompilerEnv> CompilerEnv for &'a T {
 impl<'a> Compiler<'a> {
     pub fn new(globals: &'a (CompilerEnv + 'a),
                vm: &'a GlobalVmState,
-               symbols: SymbolModule<'a>)
+               symbols: SymbolModule<'a>,
+               source_map: &'a Source<'a>)
```
brendanzab: This isn't really a `SourceMap` at this stage - right?

src/compiler_pipeline.rs
```diff
@@ -196,6 +199,7 @@ impl<E, Extra> Compileable<Extra> for TypecheckValue<E>
                compiler: &mut Compiler,
                thread: &Thread,
                filename: &str,
+               expr_str: &str,
```
brendanzab: Wondering if we could be passing in a `Source` at this stage. Ditto with the other methods.

src/compiler_pipeline.rs
```diff
@@ -181,10 +183,11 @@ impl<'a, 'b, T> Compileable<(&'a str, Option<&'b ArcType>)> for T
                compiler: &mut Compiler,
                thread: &Thread,
                file: &str,
-               (expr_str, expected_type): (&'a str, Option<&'b ArcType>))
+               expr_str: &str,
```
brendanzab: :+1:

src/compiler_pipeline.rs
```diff
@@ -181,10 +183,11 @@ impl<'a, 'b, T> Compileable<(&'a str, Option<&'b ArcType>)> for T
...
+               expected_type: Option<&'b ArcType>)
                -> Result<CompileValue<Self::Expr>> {
         self.typecheck_expected(compiler, thread, file, expr_str, expected_type)
-            .and_then(|tc_value| tc_value.compile(compiler, thread, file, ()))
+            .and_then(|tc_value| tc_value.compile(compiler, thread, file, expr_str, ()))
     }
 }
 impl<E, Extra> Compileable<Extra> for TypecheckValue<E>
```
brendanzab: Could ad a space here while you are in this file. Same with the other `impl`s.

vm/src/source_map.rs
```diff
@@ -0,0 +1,37 @@
+use base::pos::Line;
+
+#[derive(Debug)]
+pub struct SourceMap {
+    /// `map` stores the index of the first instruction for each line.
```
brendanzab:
```rust
/// The index of the first instruction for each line
```

vm/src/compiler.rs
Marwes: It is some sort of source mapping however :P I should rename it now however.

src/compiler_pipeline.rs
Marwes: Its still pretty terrible however :/

vm/src/compiler.rs
brendanzab: Yeah... it's interesting that they are similar in function. Let's see how it pans out - maybe in the future we can find some common idea in `Source` and `SourceMap` to make this stuff simpler.

Marwes:
Superseeded by #182

PR #181, opened 2016-10-10
[WIP] feat: Add source maps

Marwes:


Marwes:
Oops, apparently this defaulted to a PR to the source_map branch.

brendanzab:
If you edit the title of a PR, you can re-target the branch. Just for future reference!

Marwes:
Oh, hadn't notice that, thanks (I'd say its not an especially intuitive way of doing it though).

brendanzab:
Yep - it's quite odd! I only know because I stumbled upon it one day...

PR #182, opened 2016-10-10
feat: Source map

Marwes:


Marwes:
Lost push access to this repo after moving it which is for the better since I can just keep my own branches separate from this repo. I did have to resubmit #176 however. Fixed the nits and added some miscellaneous stuff needed for the language-server that should perhaps be in a different PR.

Marwes:
Rebased this and since I believe this to be in decent shape I will merge this. I may to do some more iterations on the source map but as it already gives some value through line numbers on the stacktrace.

PR #183, opened 2016-10-13
feat: Tweaks to enable the language server to work with 0.2

Marwes:
Extracted from #182 to first be able to use the language server with 0.2 letting #182 wait until they have been tested with the language server.

### DIFF
--- a/base/src/source.rs
+++ b/base/src/source.rs
@@ -1,12 +1,18 @@
+//! Module containing types and functions for mapping between byte indexes and line and column
+//! locations
+
 use pos::{BytePos, Column, Line, Location};
 
+/// Type which provides a bidirectional mapping between byte offsets and line and column locations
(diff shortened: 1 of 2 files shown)


base/src/source.rs
```diff
@@ -1,31 +1,74 @@
 use pos::{BytePos, Column, Line, Location};
 
 #[derive(Clone, Debug)]
-pub struct Source<'a> {
-    src: &'a str,
-    /// The starting byte position of each line in `src`
+pub struct Lines {
     lines: Vec<BytePos>,
```
brendanzab: Maybe call this `starting_bytes` to reduce stuttering?

Also might make sense to give `Source` and `Lines` doc comments seeing as they are public APIs

Marwes:
Added some docs

PR #185, opened 2016-10-15
Move abstract types out of AliasData

brendanzab:
This cleans up a bunch of `Option` juggling and makes things a bit easier to understand!

### DIFF
--- a/base/src/ast.rs
+++ b/base/src/ast.rs
@@ -345,8 +345,7 @@ fn get_return_type(env: &TypeEnv, alias_type: &ArcType, arg_count: usize) -> Arc
                 match alias_type.as_alias() {
                     Some((id, alias_args)) => {
                         let (args, typ) = match env.find_type_info(&id).map(Alias::deref) {
-                            Some(&AliasData { ref args, typ: Some(ref typ), .. }) => (args, typ),
-                            Some(&AliasData { .. }) => panic!("ICE: '{:?}' does not exist", id),
+                            Some(&AliasData { ref args, ref typ, .. }) => (args, typ),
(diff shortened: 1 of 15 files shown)


Marwes:
The option is actually not there to indicate an unknown type but it is actually an indicator of an abstract type  (such as `Array a`). Having that hidden under `Alias` is perhaps not best so refactoring it may be a good idea. The notion of an abstract type should remain though as it is used for all native types.

brendanzab:
Ohhh, ok. Sorry for misunderstanding!

brendanzab:
What about having an `Abstract` variant under `Type`?

brendanzab:
Or perhaps this is a job for existentials?

Marwes:
Yeah, was thinking something like that. Should be clearer than than how it is right now.

brendanzab:
Ok, I made a new variant, `Opaque`. It seems that `abstract` is a reserved word in Rust 😉 

I don't fully understand how to implement existentials, or whether they would be suitable for this purpose, but this works for now.

Marwes:
`abtstract` many be reserved but `Abstract` should not be no? `opaque` is fine as well though, any reason I am missing to not go for abstract?

brendanzab:
What would I use for the constructor?

brendanzab:
`abs` is for 'abstraction'. 😆

Marwes:
Ah, forgot about it being lower case there. I guess opaque is fine then, it should not confuse anyone I hope.

brendanzab:
`abstrakt`? 😆

Marwes:
No? :p

Marwes:
Lets just go with opaque

PR #188, opened 2016-10-16
Use rows for variant constructors

brendanzab:
Like extensible records, you can't actually do much with these at the moment, but it _does_ simplify the internal representation somewhat.

### DIFF
--- a/base/src/types.rs
+++ b/base/src/types.rs
@@ -242,12 +242,18 @@ pub enum Type<Id, T = ArcType<Id>> {
         /// The rest of the row
         rest: T,
     },
-    /// An identifier type. Anything that is not a builtin type.
+    /// An identifier type. These are created during parsing, but should all be
+    ///resolved into `Type::Alias`es during type checking.
+    ///
+    /// Identifiers are also sometimes used inside aliased types to avoid cycles
+    /// in reference counted pointers. This is a bit of a wart at the moment and
+    /// _may_ cause spurious unification failures.
     Ident(Id),
-    /// Representation for type variables
+    /// An unbound type variable that may be unified with other types. These
+    /// will eventually be converted into `Type::Generic`s during generalization.
     Variable(TypeVariable),
-    /// Variant for "generic" variables. These occur in signatures as lowercase identifers `a`, `b`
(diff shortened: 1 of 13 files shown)


base/src/types.rs
```diff
@@ -750,23 +760,31 @@ impl<'a, I, T, E> DisplayType<'a, I, T, E>
...
+                            for arg in arg_iter(&field.typ) {
+                                doc = chain![arena;
+                                            doc,
+                                            " ",
+                                            dt(self.env, Prec::Constructor, &arg).pretty(arena)];
+                            }
+                        }
                     }
-                }
+                    _ => panic!("Polymorphic variant syntax not yet decided"),
```
Marwes: "Unexpected type {} in variant" or something? May be a bit clearer about what happened if this is actually reached.

base/src/types.rs
```diff
@@ -903,6 +921,7 @@ pub fn walk_type_<I, T, F: ?Sized>(typ: &T, f: &mut F)
             }
         }
         Type::Record(ref row) => f.walk(row),
+        Type::Variant(ref row) => f.walk(row),
```
Marwes: Could use `Type::Record(ref row) | Type::Variant(ref row) =>`

check/src/rename.rs
```diff
@@ -143,9 +143,9 @@ pub fn rename(symbols: &mut SymbolModule,
 
         fn stack_type(&mut self, id: Symbol, span: Span<BytePos>, alias: &Alias<Symbol, ArcType>) {
             // Insert variant constructors into the local scope
-            if let Type::Variants(ref variants) = *alias.typ {
-                for &(ref name, ref typ) in variants {
-                    self.env.stack.insert(name.clone(), (name.clone(), span, typ.clone()));
+            if let Type::Variant(ref row) = *alias.typ {
+                for field in row.field_iter().cloned() {
```
Marwes: Maybe we need to rename `field_iter` now? `row_iter` ?

brendanzab: Want me to change `type_field_iter` as well?

Marwes: That one should not apply to variants so maybe that one is fine.

PR #190, opened 2016-10-19
fix: Handle completion requests with an implicit prelude

Marwes:
The last commit (fe041e5) may have a bit of an ugly solution but I don't have a better idea at this time unfortunately. Happy to do some other solution if one exists.

### DIFF
--- a/base/src/pos.rs
+++ b/base/src/pos.rs
@@ -56,7 +56,13 @@ macro_rules! pos_struct {
 
 pos_struct! {
     /// A byte offset in a source string
-    pub struct BytePos(u32);
+    pub struct BytePos(i32);
+}
+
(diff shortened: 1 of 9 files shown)


Marwes:
So this really was two bugs, the first was that let and type did not get their spans set correctly which would cause lookups to fail.

The second is more insidious. When searching through an `Expr` we assume that we can always look at the current expression to figure out which sub-expression to traverse into. But when an implicit prelude is inject we get something like this

```f#
// Pos 0
let abc =
let bcd = 
// Pos 123
//End of implicit prelude
// Pos 0
real_expr
```

So when looking for position 56 the searcher would just look in the implicit prelude which we really just want to ignore.

PR #195, opened 2016-10-29
fix: Track from where a part of the AST originated

Marwes:
This is an attempt of a less hacky alternative to #190. By tracking the source of an expression the completion requests can skip over expanded expressions (like the implicit prelude) and only search in the non expanded expressions.

This increases the size of `Span<BytePos>` by 50% which, being a frequently used type, is rather unfortunate. We could maybe do some trickery to reduce this but this could be done later if it is necessary.

(Supersedes)
Closes #190

### DIFF
--- a/base/src/ast.rs
+++ b/base/src/ast.rs
@@ -171,14 +171,14 @@ pub trait MutVisitor {
 pub fn walk_mut_expr<V: ?Sized + MutVisitor>(v: &mut V, e: &mut SpannedExpr<V::Ident>) {
     match e.value {
         Expr::IfElse(ref mut pred, ref mut if_true, ref mut if_false) => {
-            v.visit_expr(&mut **pred);
-            v.visit_expr(&mut **if_true);
-            v.visit_expr(&mut **if_false);
+            v.visit_expr(pred);
+            v.visit_expr(if_true);
+            v.visit_expr(if_false);
         }
         Expr::Infix(ref mut lhs, ref mut id, ref mut rhs) => {
-            v.visit_expr(&mut **lhs);
+            v.visit_expr(lhs);
             v.visit_typ(&mut id.typ);
-            v.visit_expr(&mut **rhs);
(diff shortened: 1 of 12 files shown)


base/src/ast.rs
```diff
@@ -171,36 +171,36 @@ pub trait MutVisitor {
 pub fn walk_mut_expr<V: ?Sized + MutVisitor>(v: &mut V, e: &mut SpannedExpr<V::Ident>) {
     match e.value {
         Expr::IfElse(ref mut pred, ref mut if_true, ref mut if_false) => {
-            v.visit_expr(&mut **pred);
-            v.visit_expr(&mut **if_true);
-            v.visit_expr(&mut **if_false);
+            v.visit_expr(pred);
+            v.visit_expr(if_true);
+            v.visit_expr(if_false);
```
brendanzab: Neat!

base/src/pos.rs
```diff
@@ -118,14 +118,69 @@ impl fmt::Display for Location {
     }
 }
 
+/// An expansion identifier tracks wheter a span originated from a macro expansion or not
+#[derive(Copy, Clone, Default, Ord, PartialOrd, Eq, PartialEq, Debug)]
+pub struct ExpansionId(pub u32);
+
+pub const NO_EXPANSION: ExpansionId = ExpansionId(0);
+
+pub const UNKNOWN_EXPANSION: ExpansionId = ExpansionId(1);
```
brendanzab: Why did you decide not to use an `enum` here?

Marwes: Mimicked how Rust uses this field, e could later on generate `ExpansionId`s for each expanded macro so

```
import "asd" // ExpansionId(2)
import "asd" // ExpansionId(3)
some_other_macro "asd" // ExpansionId(4)
```

brendanzab:
Got some merge conflicts that need to be resolved

PR #196, opened 2016-10-29
fix(vm): Only store generation-0 values in the globals table

### LINKED ISSUE #193: Segfault in the garbage collector in try_gluon
The link to try_gluon server in the readme is dead:
http://ec2-52-29-1-213.eu-central-1.compute.amazonaws.com/

screen shot 2016-10-24 at 02 54 17

Marwes:
So the issue in #193 was that a module would be loaded from a child thread which means the objects were allocated from that thread as well. After the child thread was destroyed the values allocated from that thread still existed in the global table being potentially referenced again. To fix this we deep clone any values stored in the global table into the `GlobalVmState` gc which ensures they live long enough.

Fixes #193

### DIFF
--- a/benches/function_call.rs
+++ b/benches/function_call.rs
@@ -53,7 +53,7 @@ fn factorial_tail_call(b: &mut ::test::Bencher) {
 fn gluon_rust_boundary_overhead(b: &mut ::test::Bencher) {
     let vm = new_vm();
 
-    fn test_fn(_: &Thread) -> Status {
+    extern "C" fn test_fn(_: &Thread) -> Status {
         Status::Ok
     }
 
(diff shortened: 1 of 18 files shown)


Marwes:
Relevant https://github.com/gluon-lang/gluon/blob/8a3e9f5efd8016c77506530ebf4c8b510100d957/vm/src/gc.rs#L98-L131

brendanzab:
This is a bit magic to me, but I trust you!

Marwes:
Sorry! I can try to explain some parts a bit more thoroughly if you like but this kind of fix is going to be rather involved in the nitty gritty no matter how you look at it :).

brendanzab:
Yeah - more about my inexperience with VMs and stuff

Issue #194, opened 2016-10-24
Add binaries to github releases


dgellow:

Marwes:
Seems like a good idea, not sure what the easiest way of doing this is. I think appveyor produces build artifacts, unsure if travis can do it.

dgellow:
I think you'll need this https://docs.travis-ci.com/user/deployment/releases/

Marwes:
Thanks! Will have a look.

saifikhan:
What are the target platforms ?

Perhaps, 64-bit FreeBSD, Linux and Windows.

Marwes:
linux should be first since that is what travis builds on, Appveyor should be possible with windows. Dunno how to get FreeBSD builds, I don't believe traivs has FreeBSD build machines in any case. Mac could maybe work, travis makes builds on it but they are unfortunately extremely slow to start (not enough machines to run builds) which is problematic.

Unfortunately I ran into problems last time I tried and had to remove the binary deployment https://github.com/gluon-lang/gluon/commit/969d66852c3463c4909e99cdcc39293b347eda60 would be happy to accept PRs for this though from anyone that knows how to get it working.

Marwes:
Linux builds from travis is working now. Haven't managed to get Appveyor (windows) working yet however.

https://github.com/gluon-lang/gluon/releases

Still compress the binary (which should let me give the artifact a more informative name as well).

Marwes:
Releases should now be provided for every release. Currently we build the following platforms. 

* gluon-v0.7.1-x86_64-apple-darwin.tar.gz
* gluon-v0.7.1-x86_64-pc-windows-gnu.tar.gz
* gluon-v0.7.1-x86_64-unknown-freebsd.tar.gz
* gluon-v0.7.1-x86_64-unknown-linux-gnu.tar.gz 

If builds for additional platforms are desired just leave a comment or issue a PR adding it like the other deploy only builds https://github.com/gluon-lang/gluon/blob/8a0be0956960ef79217883f821c1b1f26b454ec6/.travis.yml#L16-L30 !

https://github.com/gluon-lang/gluon/releases/tag/v0.7.1

PR #197, opened 2016-10-30
feat: Add the repl binary to github releases

Marwes:
I think this should solve #194. The repl executable should be the only relevant file to upload

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,14 @@ env:
 notifications:
   webhooks:
     urls:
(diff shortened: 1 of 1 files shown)


brendanzab:
I'm guessing this only does linux?

Marwes:
Yep, I think travis can build for mac as well but not for windows. Never had any reason to test other platforms since there is not anything platform specific to test. Could maybe add an appveyor build as well for windows

PR #200, opened 2016-11-11
Use try shorthand

brendanzab:
So this is a thing now?

### DIFF
--- a/check/src/kindcheck.rs
+++ b/check/src/kindcheck.rs
@@ -1,4 +1,5 @@
 use std::fmt;
+use std::result::Result as StdResult;
 
 use base::ast;
 use base::kind::{self, ArcKind, Kind, KindEnv};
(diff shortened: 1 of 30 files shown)


parser/src/lexer.rs
```diff
@@ -513,15 +512,15 @@ impl<'input, I> Lexer<'input, I>
                    -> ParseResult<Token<&'input str>, LocatedStream<I>> {
         loop {
             // Skip all whitespace before the token
-            let (_, new_input) = try!(spaces().parse_lazy(input).into());
+            let (_, new_input) = spaces().parse_lazy(input).into()?;
             input = new_input.into_inner();
```
Marwes: Looks like inference fails here because the preceding `into()` call cannot figure out that it should be turned into a `Result` :/

brendanzab:
Had to step around rustfmt in some places.... :(

check/src/kindcheck.rs
```diff
@@ -192,41 +192,46 @@ impl<'a> KindCheck<'a> {
...
                             name: field.name.clone(),
                             typ: typ,
                         })
                     })
-                    .collect());
-                Ok((self.type_kind(), Type::variant(row)))
+                    .collect();
+
+                #[cfg_attr(rustfmt, rustfmt_skip)]
+                Ok((self.type_kind(), Type::variant(try!(row))))
```
Marwes: So rustfmt automatically rewrites `try!` now? Maybe it is better to add type annotations to help out `.collect::<Result<_, _>>()`?

PR #204, opened 2016-11-20
[WIP] Remove primitive operators

brendanzab:
This simplifies the internals by removing the special-cased primitive operators

### DIFF
--- a/check/tests/completion.rs
+++ b/check/tests/completion.rs
@@ -119,8 +119,8 @@ fn binop() {
 
     let (mut expr, result) = support::typecheck_expr(r#"
 let (++) l r =
-    l #Int+ 1
-    r #Float+ 1.0
+    prim.add_Int l 1
+    prim.add_Float r 1.0
     l
 1 ++ 2.0
 "#);
@@ -217,7 +217,7 @@ fn suggest_after_unrelated_type_error() {
 
     let result = suggest(r#"
 let record = { aa = 1, ab = 2, c = "" }
-1.0 #Int+ 2
+prim.add_Int 1.0 2
 record.a
 "#,
                          BytePos::from(104));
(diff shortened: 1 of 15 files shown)


brendanzab:
Not sure about how you feel about this. Have some failing tests though because all the special cases have been remove from the type checker, etc...

Marwes:
Could you add some benchmark for calling these operations and compare it before and after?

While I think that removing opcodes for addition, subtraction etc is a non-starter due to how slow it is going to be, the way `(+)`, `(-)` work now is they call a gluon function which runs the opcode so it is already slow now. If there is not a noticeable speed difference compared to the current implementation I'd be on board on doing this, however it will make adding basic inlining much more important in my mind as it just shows how slow the current way of calling a function for basic operations is simply far to slow.

check/tests/completion.rs
```diff
@@ -119,8 +119,8 @@ fn binop() {
 
     let (mut expr, result) = support::typecheck_expr(r#"
 let (++) l r =
-    l #Int+ 1
-    r #Float+ 1.0
+    prim.add_Int l 1
```
Marwes: Either add dummy functions with the correct type signature and call them
```
let add_Int x : Int -> Int -> Int = add_Int x
```
or add the basic functions we need to the `MockEnv`

or add a way to import modules in just the typechecker. That would be nice but the macro system is currently tied into the vm module so probably not feasible atm.

vm/src/primitives.rs
```diff
@@ -284,6 +364,26 @@ pub fn load(vm: &Thread) -> Result<()> {
         show_Int => primitive!(1 prim::show_int),
         show_Float => primitive!(1 prim::show_float),
         show_Char => primitive!(1 prim::show_char)
+        add_Int => primitive!(2 add_int),
```
Marwes: Why not `s/add_int/VMInt::add`? I guess the name in the vm would be different but maybe we should overload `primitive!` to allow a different name?

PR #206, opened 2016-11-26
fix: Don't panic when compiling partially compiled constructors

### LINKED ISSUE #202: Partially applying variant constructors panics in the compiler
Partially applied variant constructors will panic with Constructor x is not fully applied' in the compiler. This should either be an error during typechecking (easy) or be rewritten into a lambda (hard)

### Reproduction

Type `Some` into the repl.

Marwes:
Ideally partially applied constructors should be compiled into an automatically generated lambda but until then we should at least not panic when encountering one.

Fixes #202

### DIFF
--- a/tests/vm.rs
+++ b/tests/vm.rs
@@ -718,6 +718,15 @@ pure 123
 }
 
 #[test]
+fn dont_panic_on_partially_applied_constructor() {
+    let _ = ::env_logger::init();
+    let vm = make_vm();
+
+    let result = Compiler::new().run_expr::<OpaqueValue<&Thread, Hole>>(&vm, "test", "Some");
+    assert!(result.is_err());
+}
+
+#[test]
 fn stacktrace() {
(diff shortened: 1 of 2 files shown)


brendanzab:
How does this actually work?

Marwes:
What do you mean? I changed the panic in returning an error. Just changed the part in the compiler instead of doing it in the typechecker since I think we want to rewrite `Some` -> `\x -> Some x` later on so this was the easiest fix.

brendanzab:
Ah - yeah just seemed a bit of magic to me that changing something so simple could fix the problem!

Marwes:
A while back the compiler didn't return errors so it needed to panic but since that was changed it was trivial to hook into that. I suppose there is an argument for not panicking at all, but int theory all panics should only be reachable if there is a bug in the typechecker so I prefer to keep them as panics since those are easier to debug.

Issue #208, opened 2016-11-28
Add examples explaining how to convert data between rust and gluon


ethanpailes:
The tutorial makes the right way to get primitive data types back from a call to run_expr fairly obvious, but I've been struggling to figure out the best way to get back more complex data types. In particular I would like to evaluate a gluon expression to return a record which contains a list of strings among other things. I have faith that enough time bashing my head against the rustdoc would get me something that works, but I'm not sure that a rust noob like me could get the correct solution. By correct solution I mean something which has the following attributes:

[] Typesafe. Any possible impedance mismatch between the gluon value and the expected rust value should be described by some sort of Result type. 
[] Minimal boilerplate. Getting to a rust value from a gluon value should require the least amount of typing possible. Maybe there is already a magic macro to two to help with this.
[] Minimal overhead. Obviously it would be nice to be able to avoid copying memory unless I have to.

If someone with more experience could explain how to do that stuff to me I would be thrilled to open a PR adding documentation to the tutorial or README.

Also, even though I'm not trying to convert rust values to gluon values just yet, it would be nice to document both directions of the conversion for completeness. Again I would be happy to write something up.

Marwes:
Depending on exactly what you want to do there are things that can be done, that should be possible and some that probably will never work (safely :). I think what you want is some way of retrieving a gluon record `{ x: Int, y : String }`?

For pushing records to gluon the [record!](https://github.com/gluon-lang/gluon/blob/c87c733c519eb25f8ae513bf737cd21b1aaa8ab8/vm/src/primitives.rs#L269-L281) macro can be used but there is no equivalent for retrieving a record atm. I added a macro for unpacking in #209 and I should probably add a macro for defining record types as well since the current way is rather cumbersome ([used here](https://github.com/gluon-lang/gluon/blob/c87c733c519eb25f8ae513bf737cd21b1aaa8ab8/vm/src/channel.rs#L98-L113)).

That should solve point 1 and 2 I think, maybe some way of holding a gluon record in rust without unpacking it could be useful as well but I think that will require some more advanced macro magic.
```rust
let gluon_record = thread.run_expr("{ x = 1, y = "a" }");
gluon_record.get_x()// 1
gluon_record.get_y()// "a"
```

For point 3 that really depends on what you are specifically trying to do, say if you want to retrieve a string, the easiest way is to just specify that the function returns `std::string::String` and the string will be copied out, but it is also possible to use `RootStr`  in which case the string in gluon is reference counted.

[...]

PR #209, opened 2016-11-28
feat: Add a macro for describing gluon record patterns

Marwes:
cc #208

### DIFF
--- a/vm/src/api.rs
+++ b/vm/src/api.rs
@@ -1322,6 +1322,31 @@ macro_rules! record {
 
 /// Creates a pattern which matches on
 /// ```rust
+/// // Fields used in `record_type!` needs to be forward declared
+/// field_decl! {x, y}
+/// fn new_vec(x: f64, y: f64) -> record_type!(x => f64, y f64) {
+///     record_no_decl!(x => y, y => y)
+/// }
+/// ```
+#[macro_export]
+macro_rules! record_type {
+    (impl) => { () };
+    (impl $field: ident => $value: ty) => {
+        $crate::api::record::HList<(_field::$field, $value), ()>
+    };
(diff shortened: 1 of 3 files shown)


vm/src/api.rs
```diff
@@ -1320,6 +1320,36 @@ macro_rules! record {
     }
 }
 
+#[macro_export]
+macro_rules! hlist_p {
```
brendanzab: No docs? :3

Marwes: That is an internal macro but since it is used by the exported `record_p` macro it needs to be exported as well. Maybe I can hide it inside `record_p` instead.

brendanzab: At least a comment saying 'Used internally by `record_p!`'... or something

brendanzab: ohhh, there is also `#[no_doc]` I think?

Marwes: Decided to hide the internal macros inside the macros themselves since they are only used in one macro each.

brendanzab: Maybe check the generated docs and see if it looks fine - macro scoping is weird...

vm/src/api.rs
```diff
@@ -1320,6 +1361,74 @@ macro_rules! record {
...
+    ($field: ident => $value: ty) => {
+        $crate::api::record::HList<(_field::$field, $value), ()>
+    };
+    ($field: ident => $value: ty, $($field_tail: ident => $value_tail: ty),*) => {
+        $crate::api::record::HList<(_field::$field, $value),
+                                record_type_inner!( $($field_tail => $value_tail),*)>
+    }
+}
+
+/// Creates a pattern which matches on
```
brendanzab: Unfinished?

Marwes: I am sure I fixed that...

PR #214, opened 2016-12-03
[WIP] Remove unwrap on tokenizer errors

### LINKED ISSUE #210: Don't panic on tokenizer errors
At the moment (as of #103), [we `unwrap` token results when iterating through them](https://github.com/gluon-lang/gluon/blob/c05428e2d98e859781a70a144daf8b8fcf4835f7/parser/src/lib.rs#L142). We should instead break the iteration and return the token error. This would also involve adding another variant to `gluon_parser::Error`.

brendanzab:
Fixes #210

### DIFF
--- a/parser/src/layout.rs
+++ b/parser/src/layout.rs
@@ -1,4 +1,4 @@
-use base::pos::{self, BytePos, Column, Line, Location, Spanned};
+use base::pos::{self, Column, Location};
 
 use token::{SpannedToken, Token};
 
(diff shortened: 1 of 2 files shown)


parser/src/layout.rs
```diff
@@ -359,7 +353,10 @@ impl<'input, Tokens> Layout<'input, Tokens>
                 (&Token::With, _) => self.scan_for_next_block(Context::MatchClause)?,
 
                 (&Token::Else, _) => {
-                    let next = self.next_token();
+                    let next = match self.unprocessed_tokens.pop() {
+                        Some(token) => token,
+                        None => return Ok(None),
```
Marwes: I think this may cause the initial token (from line 148) to be forgotten? Maybe unprocessed_tokens need to be filled with enough tokens to ensure it always has tokens available (unless its at EOF).

PR #215, opened 2016-12-03
[WIP] Implement universal quantification

### LINKED ISSUE #184: Add universally quantified types
It would be nice to add universally quantified types to our abstract syntax:

``` rust
pub enum Type<Id, T = ArcType<Id>> {
    ...
    /// Universally quantified type
    Forall(Vec<Generic<Id>>, T),
    ...
}
```

We can figure out concrete syntax later.

brendanzab:
Fixes #184

### DIFF
--- a/base/src/types.rs
+++ b/base/src/types.rs
@@ -172,9 +172,9 @@ impl<Id, T> AsRef<T> for Alias<Id, T> {
 impl<Id, T> Alias<Id, T>
     where T: From<Type<Id, T>>,
 {
-    pub fn new(name: Id, args: Vec<Generic<Id>>, typ: T) -> Alias<Id, T> {
+    pub fn new(name: Id, typ: T) -> Alias<Id, T> {
         Alias {
-            _typ: Type::alias(name, args, typ),
+            _typ: Type::alias(name, typ),
             _marker: PhantomData,
         }
     }
@@ -201,8 +201,6 @@ impl<Id> Alias<Id, ArcType<Id>>
 pub struct AliasData<Id, T> {
     /// Name of the Alias
     pub name: Id,
-    /// Arguments to the alias
-    pub args: Vec<Generic<Id>>,
     /// The type that is being aliased
     pub typ: T,
 }
@@ -242,6 +240,8 @@ pub enum Type<Id, T = ArcType<Id>> {
     Opaque,
     /// A builtin type
     Builtin(BuiltinType),
+    /// Universally quantified types
+    Forall(AppVec<Generic<Id>>, T),
     /// A type application with multiple arguments. For example,
     /// `Map String Int` would be represented as `App(Map, [String, Int])`.
     App(T, AppVec<T>),
@@ -285,6 +285,18 @@ impl<Id, T> Type<Id, T>
 
     pub fn opaque() -> T {
         T::from(Type::Opaque)
+    }
+
+    pub fn builtin(typ: BuiltinType) -> T {
+        T::from(Type::Builtin(typ))
+    }
+
+    pub fn forall(params: AppVec<Generic<Id>>, typ: T) -> T {
+        if params.is_empty() {
+            typ
+        } else {
+            T::from(Type::Forall(params, typ))
+        }
     }
 
     pub fn array(typ: T) -> T {
@@ -351,18 +363,13 @@ impl<Id, T> Type<Id, T>
         T::from(Type::Generic(typ))
     }
 
-    pub fn builtin(typ: BuiltinType) -> T {
-        T::from(Type::Builtin(typ))
-    }
-
     pub fn variable(typ: TypeVariable) -> T {
         T::from(Type::Variable(typ))
     }
 
-    pub fn alias(name: Id, args: Vec<Generic<Id>>, typ: T) -> T {
+    pub fn alias(name: Id, typ: T) -> T {
         T::from(Type::Alias(AliasData {
             name: name,
-            args: args,
             typ: typ,
         }))
     }
@@ -423,6 +430,14 @@ impl<Id, T> Type<Id, T>
             Type::Ident(ref id) => Some(id),
             Type::Alias(ref alias) => Some(&alias.name),
             _ => None,
+        }
+    }
+
+    pub fn params(&self) -> &[Generic<Id>] {
+        match *self {
+            Type::Alias(ref alias) => alias.typ.params(),
+            Type::Forall(ref params, _) => params,
+            _ => &[],
         }
     }
 }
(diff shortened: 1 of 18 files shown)


brendanzab:
Just the syntax right now, using the straw-man `forall <args> . <type>` syntax...

base/src/types.rs
```diff
@@ -760,6 +766,16 @@ impl<'a, I, T, E> DisplayType<'a, I, T, E>
...
+            Type::Forall(ref args, ref typ) => {
+                chain![arena;
+                       "forall",
+                       arena.concat(args.iter().map(|arg| {
+                           arena.text(self.env.string(&arg.id)).append(" ").into()
+                       })),
+                       " .",
+                       arena.newline(),
+                       top(self.env, typ).pretty(arena)]
+            }
```
brendanzab: Need to add a test for this!

parser/src/grammar.lalrpop
```diff
@@ -167,6 +168,12 @@ AppType = {
 Type = {
     AppType,
 
+    "forall" <args: Ident+> "." <ty: Type> =>
+        Type::forall(args.into_iter()
+                         .map(|id| Generic { id: id, kind: Kind::variable(0) })
```
brendanzab: Should we have a `Kind::Hole`?

Marwes: Maybe, but it this variable will never get read from in any case so I don't think its necessary (for types `Hole` is at least used other than a dumy value).

parser/src/grammar.lalrpop
```diff
@@ -108,19 +109,19 @@ TypeBinding: TypeBinding<Id> = {
         TypeBinding {
             comment: None,
             name: id.clone(),
-            alias: Alias::new(id, params, Type::variant(row)),
+            typ: Type::forall(params, Type::variant(row)),
```
Marwes: I'd rather not lose the information of `id` here. During typechecking, when `id` is refereed to, the `typ/alias` field is loaded and for efficiency and the sake of user-friendliness returning a `Type::Alias` is going to give more information than `Type::Forall` (though they should typecheck the same),

check/src/kindcheck.rs
```diff
@@ -171,6 +171,7 @@ impl<'a> KindCheck<'a> {
             }
             Type::Variable(_) => Ok((self.subs.new_var(), typ.clone())),
             Type::Builtin(builtin_typ) => Ok((self.builtin_kind(builtin_typ), typ.clone())),
+            Type::Forall(_, _) => unimplemented!(),
```
brendanzab: Not sure about what to do here... :/

Marwes: Shouldn't it be the kind of the instantiated type? `ForAll(vars, typ) => instantiate(vars, typ).kind()`?

Marwes: Oh, but what would one instantiate with...

check/src/typecheck.rs
```diff
@@ -146,7 +146,7 @@ impl<'a> KindEnv for Environment<'a> {
             .get(type_name)
             .map(|&(_, ref alias)| {
                 let mut kind = Kind::typ();
-                for arg in alias.args.iter().rev() {
+                for arg in alias.typ.params().iter().rev() {
```
brendanzab: Not sure if we should be working off aliases here now. Might we be missing `forall`s?

check/src/kindcheck.rs
Marwes: No, it should be fine to just instantiate the type with `vars` as the variables themselves. As long as their kinds are correct the resulting kind will be correct.

Marwes:
Found a bug in base which caused forall to be forgotten.

The second commit does some changes which unpacks foralls in type bindings, I believe this needs to be done in a few more places as well so we probably need some helper to make it less error prone.

brendanzab:
Thanks for having a look! Will try to fix the other places :)

base/src/types.rs
```diff
@@ -1161,7 +1161,7 @@ pub fn walk_move_type_opt<F: ?Sized, I, T>(typ: &Type<I, T>, f: &mut F) -> Optio
           I: Clone,
 {
     match *typ {
-        Type::Forall(_, ref typ) => f.visit(typ),
+        Type::Forall(ref args, ref typ) => f.visit(typ).map(|typ| Type::forall(args.clone(), typ)),
```
brendanzab: :facepalm:

check/src/typecheck.rs
```diff
@@ -748,7 +750,10 @@ impl<'a> Typecheck<'a> {
 
                 let mut duplicated_fields = FnvSet::default();
                 for field in associated_types.iter().chain(fields.iter()) {
-                    if self.error_on_duplicated_field(&mut duplicated_fields, span, field.0.clone()) {
+                    if 
+                        self.error_on_duplicated_field(&mut duplicated_fields,
+                                                       span,
+                                                       field.0.clone()) {
```
brendanzab: That is some weird rustfmting!

check/src/typecheck.rs
```diff
@@ -289,11 +289,13 @@ impl<'a> Typecheck<'a> {
 
     fn stack_type(&mut self, id: Symbol, alias: &Alias<Symbol, ArcType>) {
         // Insert variant constructors into the local scope
-        if let Type::Variant(ref row) = *alias.typ {
-            for field in row.row_iter().cloned() {
-                let symbol = self.symbols.symbol(field.name.as_ref());
-                self.original_symbols.insert(symbol, field.name.clone());
-                self.stack_var(field.name, field.typ);
+        if let Type::Forall(_, ref typ) = *alias.typ {
```
brendanzab: Why do we do this only if it is quantified?

check/src/typecheck.rs
```diff
@@ -952,8 +957,18 @@ impl<'a> Typecheck<'a> {
...
             for bind in bindings.iter_mut() {
                 check.set_variables(&bind.alias.typ.params());
-                let typ = &mut Alias::make_mut(&mut bind.alias).typ;
-                check.kindcheck_type(typ)?;
+                Alias::make_mut(&mut bind.alias).typ = match *bind.alias.typ {
+                    Type::Forall(ref args, ref typ) => {
+                        let mut typ = typ.clone();
+                        check.kindcheck_type(&mut typ)?;
+                        Type::forall(args.clone(), typ.clone())
+                    }
```
brendanzab: Why did you not put this in `kindcheck_type`?

check/src/typecheck.rs
Marwes: hmm, yeah didn't notice that.

check/src/typecheck.rs
Marwes: My bad, was just testing a theory and never fixed it throughly. It should handle both cases.

check/src/kindcheck.rs
```diff
@@ -171,6 +171,14 @@ impl<'a> KindCheck<'a> {
...
             Type::Variable(_) => Ok((self.subs.new_var(), typ.clone())),
             Type::Builtin(builtin_typ) => Ok((self.builtin_kind(builtin_typ), typ.clone())),
+            Type::Forall(ref params, ref typ) => {
+                let (ret_kind, ret_typ) = self.kindcheck(typ)?;
+                let kind = params.into_iter().rev().fold(ret_kind, |body, param| {
+                    Kind::function(param.kind.clone(), body)
+                });
+
+                Ok((kind, ret_typ.clone()))
+            }
```
Marwes: I don't think this is right. When encountering a `forall` the kindchecking needs to introduce variables for `params` and the resulting kind ought to be the kind of `typ`.

Does that make sense?

check/src/typecheck.rs
Marwes: No particular reason, it may make more sense inside `kindcheck_type`

Marwes:
Didn't have time to look at this again yesterday but I will look at it tonight

check/src/typecheck.rs
Marwes: No I think this should be right.

check/src/typecheck.rs
```diff
@@ -748,7 +748,10 @@ impl<'a> Typecheck<'a> {
 
                 let mut duplicated_fields = FnvSet::default();
                 for field in associated_types.iter().chain(fields.iter()) {
-                    if self.error_on_duplicated_field(&mut duplicated_fields, span, field.0.clone()) {
+                    if
+                        self.error_on_duplicated_field(&mut duplicated_fields,
```
Marwes: Still weird formatting, maybe I accidently added a newline and rustfmt is not removing it now?

check/src/kindcheck.rs
brendanzab: Hmm, try messing around with a few things and it didn't make a difference :O

check/src/typecheck.rs
brendanzab: I wonder if this  and the above logic need to be in `Kindcheck::kindcheck`?

check/src/kindcheck.rs
```diff
@@ -188,6 +188,11 @@ impl<'a> KindCheck<'a> {
             }
             Type::Variable(_) => Ok((self.subs.new_var(), typ.clone())),
             Type::Builtin(builtin_typ) => Ok((self.builtin_kind(builtin_typ), typ.clone())),
+            Type::Forall(ref params, ref typ) => {
+                self.set_variables(params);
+                let (ret_kind, ret_typ) = self.kindcheck(typ)?;
+                Ok((ret_kind.clone(), Type::forall(params.clone(), ret_typ)))
+            }
```
brendanzab: How does this look @Marwes?

brendanzab: Wait - this might clear annotations...

check/src/typecheck.rs
```diff
@@ -937,35 +939,24 @@ impl<'a> Typecheck<'a> {
...
-                // and bind the same variables to the arguments of the type binding
-                // ('a' and 'b' in the example)
-                let mut id_kind = check.type_kind();
-                let alias = Alias::make_mut(&mut bind.alias);
-                for generic in alias.args.iter_mut().rev() {
-                    check.instantiate_kinds(&mut generic.kind);
-                    id_kind = Kind::function(generic.kind.clone(), id_kind);
-                }
-                check.add_local(alias.name.clone(), id_kind);
-            }
```
brendanzab: Moved this into kindcheck.

check/src/kindcheck.rs
```diff
@@ -188,6 +183,21 @@ impl<'a> KindCheck<'a> {
...
+                let params: AppVec<_> = params.iter()
+                    .cloned()
+                    .map(|mut param| {
+                        self.instantiate_kinds(&mut param.kind);
+                        param
+                    })
+                    .collect();
+
+                self.variables.clear();
+                self.variables.extend(params.iter().cloned());
```
brendanzab: After inlining this, I wonder... is this correct?

check/src/unify_type.rs
```diff
@@ -171,9 +171,11 @@ fn do_zip_match<'a, U>(unifier: &mut UnifierState<'a, U>,
 {
     debug!("Unifying:\n{:?} <=> {:?}", expected, actual);
     match (&**expected, &**actual) {
-        (&Type::App(ref l, ref l_args), &Type::App(ref r, ref r_args)) => {
-            Ok(unify_app(unifier, l, l_args, r, r_args))
+        (&Type::App(ref l_typ, ref l_args), &Type::App(ref r_typ, ref r_args)) => {
+            Ok(unify_app(unifier, l_typ, l_args, r_typ, r_args))
         }
+        (&Type::Forall(_, ref l_typ), _) => Ok(unifier.try_match(l_typ, actual)),
+        (_, &Type::Forall(_, ref r_typ)) => Ok(unifier.try_match(expected, r_typ)),
```
brendanzab: Would I need to do something with the params here?

Marwes: It might not be unsound but it may give weird results since the types will contain uninstantiated variables (`Type::Generic(_)`).

If you do instantiate the forall then I think it needs to be only `forall ... <=> forall ...` or else this may unify which I don't believe is sound
`forall b. { id : forall a . a -> a, x: b } <=> forall a b. { id: a -> a, x: b}`

brendanzab: How did the aliases originally instantiate their generic vars?

Marwes: See type_of_alias in resolve
```f#
type Result e t = ..

// Result Int String
// Replace e => Int, t => String
// 
```

PR #217, opened 2016-12-03
Remove documentation upload from .travis.yml

brendanzab:
We now have Docs.rs! :)


Marwes:
Forgot about that old doc upload!

PR #205, opened 2016-11-26
[WIP] feat: Add pretty printing of expressions

Marwes:
Unsure how much I will do on this but I think I will at least fix the fact that the expression after `=` or `->` does not appear on the next line (I believe this is a limitation of pretty atm).

Partially fixes #133 

# map.glu
```f#
let prelude = import "std/prelude.glu"
and { Ordering, Ord, Option, List, Monoid, } = prelude
in
type Map k a = | Bin k a (Map k a) (Map k a) | Tip
in
let empty = Tip
in
let singleton k v = Bin k v empty empty
and make ord = let find k m = match m with
        | Bin k2 v l r -> match ord.compare k k2 with
            | LT -> find k l
            | EQ -> Some v
            | GT -> find k r
        | Tip -> None
    and insert k v m = match m with
        | Bin k2 v2 l r -> match ord.compare k k2 with
            | LT -> Bin k2 v2 insert k v l r
            | EQ -> Bin k v l r
            | GT -> Bin k2 v2 l insert k v r
        | Tip -> Bin k v empty empty
    and to_list m = let ++ = prelude.monoid_List.append
        in match m with
        | Bin key value l r -> to_list l ++ Cons { key, value, } to_list r
        | Tip -> Nil
    in let append l r = match l with
        | Bin lk lv ll lr -> match r with
            | Bin rk rv rl rr -> match ord.compare lk rk with
                | LT -> Bin lk lv ll append lr r
                | EQ -> Bin lk lv append ll rl append lr rr
                | GT -> Bin lk lv append ll r lr
            | Tip -> l
        | Tip -> match r with
            | Bin a b c d -> r
            | Tip -> empty
    in let monoid = { append, empty, }
    in { monoid, singleton, find, insert, to_list, }
in
{ Map, make,
}
```

Issue #107, opened 2016-08-09
Consider adding an additional intermediate representation


Marwes:
Currently the AST is the only intermediate representation (IR) and it is directly compiled into bytecode with no extra stage in between. This is mostly fine as the AST has barely any syntax sugar and thus has very little redundancy. However if we are to add additional constructs such as `Getter` as suggested in #103 (`.field` => `\x -> x.field`) or nested patterns in #1 the AST is going to become more and more complex making an additional, simpler IR more valuable as it makes it simpler to implement optimizations and generate code from it.

GHC core is the low level representation I am most familiar with but I am sure there are other alternatives to model it on.
https://ghc.haskell.org/trac/ghc/wiki/Commentary/Compiler/CoreSynType

Something to consider is how this intermediate representation interacts with optimizations such as inlining from a previously loaded module (should the IR be kept so it can be inlined into a module loaded later? If that is the case maybe the IR should be something that could be execute directly?)

brendanzab:
1ML desugars into a minor variation of System Fω. Idiris uses it's own [TT core language](https://github.com/idris-lang/Idris-dev/wiki/Idris-Internals-Guide#coretths).

Marwes:
Closed by #231

PR #220, opened 2016-12-05
Add argument sugar for declaring lambda fields

brendanzab:
Turns

```
let eq_Unit : Eq () = {
    (==) l r = True
}
```

into

```
let eq_Unit : Eq () = {
    (==) = \l r -> True
}
```

### DIFF
--- a/parser/src/grammar.lalrpop
+++ b/parser/src/grammar.lalrpop
@@ -253,8 +253,19 @@ Alternative: Alternative<Id> =
     };
 
 FieldExpr: FieldExpr<Id> = {
-    <id: Ident> "=" <body: SpExpr> => {
-        FieldExpr::Value(id, Some(body))
+    <id: Ident> <args: Ident*> "=" <body: SpExpr> => {
+        if args.is_empty() {
+            FieldExpr::Value(id, Some(body))
+        } else {
+            let body_span = body.span;
+            let fun = Expr::Lambda(Lambda {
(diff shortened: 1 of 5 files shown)


brendanzab:
Not sure how you feel about this!

Marwes:
So first of all I think we should be really careful with adding syntax sugar since it never is quite as simple as it seems at first.

The implementation as it is now I think has a couple problems which makes it not quite adequate to merge as is.

Desugaring directly in the AST is maybe not going to cut it in the long run. I hope to build on #205 to make it into a code formatter and distinguishing between the sugared and non-sugared forms may be necessary then (though it could possibly special cased to always chose the sugared form).

Just doing this desugaring may be a bit to simplistic as well, since it kind of mirrors `let` bindings I think we actually want it to desugar it to let bindings to allow the declared functions to be mutually recursive.

````f#
let eq_Unit : Eq () = {
    (==) l r = True,
    (/=) l r = False
}

let eq_Unit : Eq () =
    let (==) l r = True,
    and (/=) l r = False
    { (==), (/=) }
```

However if we do that we may run into when functions and values are mixed as values cannot be part of a mutually recursive group...

brendanzab:
Quite reasonable!

brendanzab:
Might also be another case for #107

Marwes:
Yep. If we add a core Syntax adding more sugar should be much easier

PR #222, opened 2016-12-07
Implement explicit kinds on type alias parameters

### LINKED ISSUE #221: Allow kinds to be specified on type variables declarations
```
type Monad (m : Type -> Type) = ...
```

brendanzab:
Fixes #221

### DIFF
--- a/check/src/typecheck.rs
+++ b/check/src/typecheck.rs
@@ -918,6 +918,7 @@ impl<'a> Typecheck<'a> {
                                expr: &SpannedExpr<Symbol>)
                                -> TcResult<()> {
         self.enter_scope();
+
         // Rename the types so they get a name which is distinct from types from other
         // modules
         for bind in bindings.iter_mut() {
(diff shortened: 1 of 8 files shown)


check/src/typecheck.rs
```diff
@@ -933,18 +933,24 @@ impl<'a> Typecheck<'a> {
...
                 let mut id_kind = check.type_kind();
                 let alias = Alias::make_mut(&mut bind.alias);
-                for gen in alias.args.iter_mut().rev() {
-                    gen.kind = check.subs.new_var();
-                    id_kind = Kind::function(gen.kind.clone(), id_kind);
+                for generic in alias.args.iter_mut().rev() {
+                    match *generic.kind {
+                        Kind::Hole => generic.kind = check.subs.new_var(),
+                        Kind::Variable(_) => panic!("Unexpected kind variable"),
+                        Kind::Function(_, _) => unimplemented!(),
```
brendanzab: Seems like we might have to do some sort of recursive thingy here...

Marwes: Yeah, just as type signatures are instantiated.

check/tests/pass.rs
```diff
@@ -866,3 +866,28 @@ make_applicative applicative_Function
...
+    let _ = ::env_logger::init();
+    let text = r#"
+type Test (a : Type -> Type) = a Int
+type Foo a = a
+type Bar = Test Foo
+()
+"#;
+    let result = support::typecheck(text);
+    assert!(result.is_ok(), "{}", result.unwrap_err());
+}
```
Marwes: Could you add a test with kind holes and a test using `Row` as well?

brendanzab: Not sure how to implement a `Row` test with out current syntax... :/

Marwes: ok I am blind

Marwes: Maybe just something like this

```f#
type Test (r: Row) (f : Row -> Type) = f r
() 
```

brendanzab: ahhh, good idea!

PR #224, opened 2016-12-11
Assorted cleanups to alias handling

brendanzab:
Splitting out some stylistic changes from #215 into this PR so that it is easier to review.

### DIFF
--- a/check/src/unify_type.rs
+++ b/check/src/unify_type.rs
@@ -222,13 +222,7 @@ fn do_zip_match<'a, U>(unifier: &mut UnifierState<'a, U>,
                l_args.iter().zip(r_args).all(|(l, r)| l.name.name_eq(&r.name)) &&
                l_types == r_types {
                 let new_args = walk_move_types(l_args.iter().zip(r_args), |l, r| {
-                    unifier.try_match(&l.typ, &r.typ)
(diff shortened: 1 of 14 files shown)


check/src/unify_type.rs
```diff
@@ -163,6 +162,38 @@ impl<'a> Unifiable<State<'a>> for ArcType {
...
+                match resolve::remove_alias(env, &typ)? {
+                    Some(new_typ) => typ = new_typ,
+                    None => break,
+                }
+            }
+            Ok(Some(typ))
+        }
+        None => Ok(None),
+    }
+}
```
Marwes: It can probably be simplified but the current complexity/duplication seems to be coming from the fact that `Ok (None)` is returned if no alias could be expanded and `Ok (Some (typ))` is returned if at least one alias was expanded

brendanzab: Is that an important distinction?

Marwes: If it did not expand an alias the same type would be returned which would get unified again resulting in an infinite loop :)

brendanzab: Seems like it is... down where it is used in `do_zip_match`...

brendanzab:
Does this look ok?

PR #226, opened 2016-12-15
docs: Add an example of a http server

Marwes:
Threw together a super simple http server example for fun. I'd like to extend it a bit more and add comments but I think this could make a decent example (though it is an example that requires some pretty hefty dependencies unfortunately).

### DIFF
--- a/check/src/typecheck.rs
+++ b/check/src/typecheck.rs
@@ -275,9 +275,16 @@ impl<'a> Typecheck<'a> {
     }
 
     fn find_record(&self, fields: &[Symbol]) -> TcResult<(ArcType, ArcType)> {
-        self.environment
-            .find_record(fields)
-            .ok_or(TypeError::UndefinedRecord { fields: fields.to_owned() })
+        // If fields is empty it is going to match any record which means this function probably
+        // returns the wron record as the record we expect can still contain type fields.
+        // Just return an error so that inference continues without any guessed record type.
+        if fields.is_empty() {
+            Err(TypeError::UndefinedRecord { fields: fields.to_owned() })
+        } else {
+            self.environment
+                .find_record(fields)
+                .ok_or(TypeError::UndefinedRecord { fields: fields.to_owned() })
+        }
     }
 
     fn find_type_info(&self, id: &Symbol) -> TcResult<&Alias<Symbol, ArcType>> {
--- a/check/src/typecheck.rs
+++ b/check/src/typecheck.rs
@@ -176,11 +176,18 @@ impl<'a> TypeEnv for Environment<'a> {
             .iter()
             .find(|&(_, &(_, ref alias))| match **alias.unresolved_type() {
                 Type::Record(ref row) => {
-                    fields.iter()
-                        .all(|name| row.row_iter().any(|f| f.name.name_eq(name)))
+                        // Check that all of `fields` exist in the record
+                        fields.iter()
+                            .all(|name| {
+                                row.row_iter()
+                                    .map(|f| &f.name)
+                                    .chain(row.type_field_iter().map(|f| &f.name))
+                                    .any(|other| other.name_eq(name))
+                            })
+                    }
+                    _ => false,
                 }
-                _ => false,
-            })
+            )
             // FIXME Don't use unresolved_type
             .map(|t| ((t.1).0.clone(), (t.1).1.unresolved_type().clone()))
             .or_else(|| self.environment.find_record(fields))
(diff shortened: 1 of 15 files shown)


Marwes:
An unfortunate side effect of f92534c is that gluon_vm now depends on gluon_check but there is unfortunately no way to avoid this without splitting out unification into a separate crate.

Marwes:
@brendanzab Any thoughts?

This might be better of somewhere also due to the dependencies it pulls in. Possibly it could be its own crate as I think it would be rather neat if we try_gluon used this implementation, I don't think it would take to much to implement what little that site needs. (Just thinking of doing it for dogfooding and not full fledged server).

check/src/typecheck.rs
```diff
@@ -272,9 +272,16 @@ impl<'a> Typecheck<'a> {
     }
 
     fn find_record(&self, fields: &[Symbol]) -> TcResult<(&ArcType, &ArcType)> {
-        self.environment
-            .find_record(fields)
-            .ok_or(TypeError::UndefinedRecord { fields: fields.to_owned() })
+        // If fields is empty it is going to match any record which means this function probably
+        // returns the wron record as the record we expect can still contain type fields.
```
brendanzab: spelling

check/src/typecheck.rs
```diff
@@ -739,7 +746,10 @@ impl<'a> Typecheck<'a> {
 
                 let mut duplicated_fields = FnvSet::default();
                 for field in associated_types.iter().chain(fields.iter()) {
-                    if self.error_on_duplicated_field(&mut duplicated_fields, span, field.0.clone()) {
+                    if 
+                        self.error_on_duplicated_field(&mut duplicated_fields,
+                                                       span,
+                                                       field.0.clone()) {
                         pattern_fields.push(field.0.clone());
```
brendanzab: Maybe pattern match on `field`?

check/tests/pass.rs
```diff
@@ -871,3 +871,16 @@ type Bar = Test Foo
...
+    let _ = ::env_logger::init();
+    let text = r#"
+type Test = { x : Int }
+type Test2 = Int
+
+{ Test2 }
+"#;
+    let result = support::typecheck(text);
+    assert!(result.is_ok(), "{}", result.unwrap_err());
+}
```
brendanzab: Why would this fail before?

Marwes: The `find_record` function takes `&[Symbol]` and looks for a record which contains all of those fields so if the slice is empty it will match on any record (empty set is a subset of all sets).

brendanzab: Just wondering because `Test` is not used... :/

vm/src/api.rs
```diff
@@ -1563,6 +1582,15 @@ impl<'vm, F> Getable<'vm> for Function<&'vm Thread, F> {
     }
 }
 
+impl<'vm, F> Getable<'vm> for Function<RootedThread, F> {
+    fn from_value(vm: &'vm Thread, value: Variants) -> Option<Self> {
+        Some(Function {
+            value: vm.root_value(*value.0),
+            _marker: PhantomData,
+        })//TODO not type safe
```
brendanzab: Will this be fixed?

brendanzab: It's ok if we defer it - just be sure to add an issue.

check/tests/pass.rs
Marwes: Ah, `Test` is implicitly used in `find_record`.

`find_record(&[])` is called -> matches `Test` (since `&[]` is a subset of `&["x"]`) causing it to return `{ x : Int }` as the guess for the record type.

Marwes: Maybe `find_record` should be an exact match for the record type when it guesses a type for a record constructor.

vm/src/api.rs
Marwes: #229

examples/http_server.glu
```diff
@@ -0,0 +1,23 @@
...
+    handle, empty_response, get, get_request, path, listen,
+    applicative, alternative, monad } = import "examples/http.glu"
+
+let { (*>), pure } = prelude.make_Applicative applicative
+let { (<|>) } = prelude.make_Alternative alternative
+let { (>>=) } = prelude.make_Monad monad
+
+let handler =
+    (get *> path "/" *> (pure { status = OK, body = "Hello World" })) <|>
+        (get *> path "/error" *> (pure { status = InternalServerError, body = "Error" }))
```
Marwes: Left it has handler since routing is only part of what it does.

Marwes:
r?

PR #227, opened 2016-12-18
Never assume that the kind of a type variable is ‘type’

brendanzab:
I was thinking this could show up some mistakes with kind instantiation. Seems like I was right! We get the following test failures:

```
---- function_app stdout ----
	thread 'function_app' panicked at 'assertion failed: `(left == right)` (left: `Ok(App(Builtin(Function), [Generic(Generic { id: 0x103c0c0b0:a0, kind: Hole }), Generic(Generic { id: 0x103c0c088:a1, kind: Hole })]))`, right: `Ok(App(Builtin(Function), [Generic(Generic { id: 0x103c0c0b0:a0, kind: Type }), Generic(Generic { id: 0x103c0c088:a1, kind: Type })]))`)', check/tests/completion.rs:111
```

```
---- types_should_be_fully_instantiated_even_on_errors stdout ----
	thread 'types_should_be_fully_instantiated_even_on_errors' panicked at 'assertion failed: `(left == right)` (left: `App(Builtin(Function), [Generic(Generic { id: 0x110c5d108:a0, kind: Hole }), Generic(Generic { id: 0x110c5d108:a0, kind: Hole })])`, right: `App(Builtin(Function), [Generic(Generic { id: 0x110c5d108:a0, kind: Type }), Generic(Generic { id: 0x110c5d108:a0, kind: Type })])`)', check/tests/pass.rs:758
```


brendanzab:
Seems like I was right! We get the following test failures:

```
---- function_app stdout ----
	thread 'function_app' panicked at 'assertion failed: `(left == right)` (left: `Ok(App(Builtin(Function), [Generic(Generic { id: 0x103c0c0b0:a0, kind: Hole }), Generic(Generic { id: 0x103c0c088:a1, kind: Hole })]))`, right: `Ok(App(Builtin(Function), [Generic(Generic { id: 0x103c0c0b0:a0, kind: Type }), Generic(Generic { id: 0x103c0c088:a1, kind: Type })]))`)', check/tests/completion.rs:111
```

```
---- types_should_be_fully_instantiated_even_on_errors stdout ----
	thread 'types_should_be_fully_instantiated_even_on_errors' panicked at 'assertion failed: `(left == right)` (left: `App(Builtin(Function), [Generic(Generic { id: 0x110c5d108:a0, kind: Hole }), Generic(Generic { id: 0x110c5d108:a0, kind: Hole })])`, right: `App(Builtin(Function), [Generic(Generic { id: 0x110c5d108:a0, kind: Type }), Generic(Generic { id: 0x110c5d108:a0, kind: Type })])`)', check/tests/pass.rs:758
```

Any ideas?

check/src/kindcheck.rs
```diff
@@ -249,51 +249,41 @@ impl<'a> KindCheck<'a> {
...
-        match *kind {
-            Kind::Variable(id) => subs.find_type_for_var(id).cloned().or_else(|| default.cloned()),
-            _ => None,
-        }
-    })
+                       match *kind {
+                           Kind::Variable(id) => subs.find_type_for_var(id).cloned(),
+                           _ => None,
+                       }
+                   })
```
brendanzab: I checked this independently of the rest of the changes and the tests seemed to pass ok.

check/src/typecheck.rs
```diff
@@ -1041,11 +1041,8 @@ impl<'a> Typecheck<'a> {
             s.push(c as char);
             let symbol = self.symbols.symbol(&s[..]);
             if self.type_variables.get(&symbol).is_none() {
-                self.type_variables.insert(symbol,
-                                           Type::variable(TypeVariable {
-                                               id: level,
-                                               kind: Kind::typ(),
-                                           }));
+                let var = TypeVariable::new(level, Kind::hole());
+                self.type_variables.insert(symbol, Type::variable(var));
```
brendanzab: Using `Kind::hole()` here seems to be fine.

brendanzab:
I switched back to using `Kind::typ()` in `unify_type` to see if things passed. It got past the `gluon_check` tests, but then I get:

```
   Compiling gluon v0.2.2 (file:///Users/brendan/dev/rust/gluon)
error: environment variable `OUT_DIR` not defined
  --> build.rs:42:39
   |
42 |         let out_file_name = Path::new(env!("OUT_DIR")).join(file);
   |                                       ^^^^^^^^^^^^^^^
```

Dunno if this is just for the latest nightly...

brendanzab:
The `OUT_DIR` could be related to rust-lang/cargo#3368

brendanzab:
Created a PR that fixes the environment variable problem: gluon-lang/gluon#228

brendanzab:
Hmm, I'm guessing we would not want to have kind polymorphism, so we _would_ have to default to using `Kind::Type`?

brendanzab:
I'm just wondering - if we don't allow for kind polymorphism, should we event have `Kind::Variable`?

check/src/typecheck.rs
Marwes: In practice I don't think it matters but I think it would be more correct if `next_variable` takes the kind as a parameter instead. The method is only called in a single place where it can easily take the kind of the `Type::Variable` it replaced.

check/src/unify_type.rs
```diff
@@ -133,11 +133,8 @@ impl<I> Substitutable for ArcType<I> {
     type Variable = TypeVariable;
 
     fn new(id: u32) -> ArcType<I> {
-        Type::variable(TypeVariable::new(id))
-    }
-
-    fn from_variable(var: TypeVariable) -> ArcType<I> {
-        Type::variable(var)
+        // FIXME: using `Kind::hole()` here results in test failures
+        Type::variable(TypeVariable::new(id, Kind::typ()))
```
Marwes: The `Kind::Hole` is probably never replaced with an actual type. Maybe check where this function is called? It is possible that `Kind:Type` is either correct in all cases or the method could even be removed so that only the `from_variable` function is called (which receives the kind from the caller).

Marwes:
> Created a PR that fixes the environment variable problem: gluon-lang/gluon#228

Ran into that problem as well but should have actually pushed the fix upstream :).

> I'm just wondering - if we don't allow for kind polymorphism, should we event have Kind::Variable?

As long as we have kind inference `Kind::Variable` will be needed and I don't think dropping that would be a good idea since all type variables would need to be explicitly annotated with kinds.

Marwes:
> Hmm, I'm guessing we would not want to have kind polymorphism, so we would have to default to using Kind::Type when the kind can't be inferred in the context? For example in the function_app test:

Yeah `Kind::Type` seems like the sanest default (Haskell does it as well). Wouldn't mind having kind polymorphism though as long as it does not mess up anything else in unification.

brendanzab:
Where do you think the defaulting should happen?

brendanzab:
I guess I just would prefer not to overwrite concrete kinds once they have been set.

Marwes:
After finishing checking a group of type definitions is probably the sanest way of doing it. For type signatures I don't think defaulting can happen since all type variables will appear in some type which only has concrete kinds.

Marwes:
> I guess I just would prefer not to overwrite concrete kinds once they have been set.

Does this happen somewhere today?

check/src/unify_type.rs
brendanzab: This is called whenever `Substitution<ArcType>::new_var` is called.

brendanzab:
I'm guessing we don, because as I mentioned in https://github.com/gluon-lang/gluon/pull/227#discussion_r92938697 the kind vars are created defaulting to `Kind::Type`...

check/src/unify_type.rs
Marwes: I am fairly certain the correct fix is to make `new_var` take a kind as well. Looked the uses of the function over and it looks like about ~90 % of its uses then `Kind::Type` is correct but there are a few where it should be `Kind::Row` or propagated from a generic variable.

Marwes: This should be row https://github.com/gluon-lang/gluon/blob/9adf8be8065ebeb385c94829ad2368c307c95d6e/check/src/typecheck.rs#L607
https://github.com/gluon-lang/gluon/blob/9adf8be8065ebeb385c94829ad2368c307c95d6e/check/src/typecheck.rs#L766

Kind::Hole ? https://github.com/gluon-lang/gluon/blob/9adf8be8065ebeb385c94829ad2368c307c95d6e/check/src/typecheck.rs#L1199

generic.kind https://github.com/gluon-lang/gluon/blob/9adf8be8065ebeb385c94829ad2368c307c95d6e/check/src/typecheck.rs#L1288

brendanzab: Ugh, seems we have some `new_var` calls in `unify.rs`... might be a pain to remove. I tried instead to remove `kind` from `TypeVariable` entirely. Unfortunately it is required in `finish_type_` to create `Generic`...

Marwes: Ah, the bane of being to generic :/ No idea how to handle that atm but maybe I will think of something.

brendanzab: The bane of premature abstraction 😆 

I would like to have a look at how other type checkers do it... full disclosure - been experimenting with implementing a separate System-Fω checker in a local repo to learn how it works 😅

PR #231, opened 2016-12-22
refactor: Add a simpler intermediate represenentation

Marwes:
Not yet working, just wanted to post for feedback. The current core::Expr type is really only removing/merging constructors and I have not taken how easy it is to optimize into account.

#107

### DIFF
--- a/base/src/ast.rs
+++ b/base/src/ast.rs
@@ -374,6 +374,20 @@ impl<Id: Clone> Typed for TypedIdent<Id> {
     }
 }
 
+impl Typed for Literal {
+    type Ident = Symbol;
+
+    fn env_type_of(&self, env: &TypeEnv) -> ArcType {
+        match *self {
+            Literal::Int(_) => Type::int(),
+            Literal::Float(_) => Type::float(),
+            Literal::Byte(_) => Type::byte(),
+            Literal::String(_) => Type::string(),
+            Literal::Char(_) => Type::char(),
+        }
+    }
+}
+
 impl Typed for Expr<Symbol> {
     type Ident = Symbol;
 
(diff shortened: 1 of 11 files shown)


Marwes:
Blocked on #233 (since `Expr::Type` is removed from the AST, it is not possible to resolve all types in a `core::Expr` leading code in the compiler which matches on `Type::Record`, `Type::Variant` etc to fail).

brendanzab:
Why did you decide to put the core syntax in `vm`? I would have thought that it might make more sense for it to be in `base`, then move the unelaborated AST into `parser`, perhaps? Dunno.

Marwes:
> Why did you decide to put the core syntax in vm?

Since its only used in the bytecode compiler I thought it made the most sense to put it there. We could have an even more fine grained crate structure but I don't think it is worth it quite yet.

>  I would have thought that it might make more sense for it to be in base, then move the unelaborated AST into parser, perhaps? Dunno.

Moving the unelaborated AST into parser would mean that check needs a dependency on parser which it currently manages to avoid. Maybe thats not that huge of a problem anymore however since the parser now compiles reasonably fast, but there is no real reason to force that dependency now either.

The right graph is a more fine grained crate dependency graph which would allow the vm to be independent of the byte code compiler and the typechecking. unify is a crate with just the type unification which would provide the definition for `equivalent`.

```
     +-------------------+base +----------------+
     |                     +                    |                                 +-------------------+base +---------------> core
     |                     |                    |                                 |                    +                       +
     |                     |    vm needs check  |                                 |                    |                       |
     |                     |    in PR 226 for   |                                 |                    |                       |
     +                     +    equivalency     +                                 |                    v                       v
   parser               check +- - - -  -  ->  vm                                 +                  unify+                 bytecomp
       +                    +                                                   parser                 +  |                    +
       |                    |                  ^                                    +                  |  ------------> vm     |
       |                    |                  |                                    |                  v                 ^     |
       |                    |                  |                                    |                check               |     |
       |                    |                  |                                    |                  +                 |     |
       |                    v                  |                                    |                  |                 |     |
       +---------->      gluon    <------------+                                    |                  v                 |     |
                            +                                                       +---------->      gluon <------------+     |
                            |                                                                            +  ^------------------+
                            |                                                                            |
                            |                                                                            |
                            +------------->  c-api                                                       |
                                                                                                         +------------->  c-api
```

Marwes:
Taking it to the limit

graph

PR #233, opened 2016-12-24
fix(check): Don't leak `Type::Ident` outside of aliased types

Marwes:
This is a bit of a weird one. The fundamental idea behind this commit is to make any 'alias' type always know the type it aliases. So for instance, if `Option a` is encountered we always know what the representation (`| None | Some a`) is. Before this commit this was not always the case as `Option a` could be `App(Ident("Option"), Generic("a"))` and not `App(Alias("Option"), Generic("a"))`.

However, due to Rust's ownership semantics preventing cyclic reference we need to be a bit clever. Each group of `type ... and ... in` is combined into the same type (`AliasData`) since these groups are the only places which types can be mutually recursive. Then when the aliased type (`| None | Option a`) is requested by calling `fn typ` we walk through the type and replace any occurances of `Type::Ident` with the proper `Alias` variant. Some helper methods which do not do this replacement are also provided but the `typ` field itself has been hidden to discourage its use.

### DIFF
--- a/base/src/ast.rs
+++ b/base/src/ast.rs
@@ -2,7 +2,7 @@
 
 use pos::{BytePos, Spanned};
 use symbol::Symbol;
-use types::{self, Alias, ArcType, Type, TypeEnv};
+use types::{self, Alias, AliasData, ArcType, Type, TypeEnv};
 
 pub trait DisplayEnv {
     type Ident;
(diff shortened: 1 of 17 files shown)


Marwes:
Still got some FIXMEs to tackle but I want to make sure this complexity is actually a good idea before proceeding.

An alternative to storing the representation of aliases like this would be to have some global context from which the representation of `Type::Ident` can be looked up. However a global context makes it harder/impossible to know when an alias is no longer needed.

PR #235, opened 2016-12-25
fix(vm): Deep clone OpaqueValue when it is sent between disjoint threads

Marwes:
Plugs a memory safety hole which would be very simple for a user to introduce. See the added tests from a sketch of what would have been unsafe prior to this PR. This PR ensures that the owner of a `OpaqueValue` is taken into consideration when pushing it to a thread so that it will get copied if it is not possible to share it.

Builds on #234


Marwes:
Updated the PR comment.

PR #236, opened 2016-12-26
feat(parser): Recover from more invalid syntax in the parser

Marwes:


parser/src/grammar.lalrpop
```diff
@@ -75,6 +75,11 @@ IdentStr: &'input str = {
     "(" <"Operator"> ")" => <>,
 };
 
+SkipExtraTokens: () = {
+    => (),
```
brendanzab: Heh - this is interesting syntax @nikomatsakis! 😆

brendanzab:
Would be nice to get some test coverage on the errors we are emitting from the compiler...

Marwes:
> Would be nice to get some test coverage on the errors we are emitting from the compiler...

Yep, it is hard to come up with examples with poor error messages though. It is only really after this change that cascading errors end up being a problem at least.

parser/src/grammar.lalrpop
Marwes: The epsilon production has never been more obvious, no? :)

PR #237, opened 2016-12-29
feat(vm): Add a pretty printer on vm values

Marwes:
Split out from #213 for easier reviewing.

### DIFF
--- a/vm/src/value.rs
+++ b/vm/src/value.rs
@@ -273,16 +273,22 @@ impl<'t> ValuePrinter<'t> {
             let typ = remove_aliases_cow(env, typ);
             match **typ {
                 Type::Record(ref row) => {
-                        chain![arena;
+                    chain![arena;
                             "{",
                             arena.concat(fields.into_iter().zip(row.row_iter())
                                 .map(|(field, type_field)| {
-                                    arena.space().append(p(&type_field.typ, env, Top, field).pretty(arena))
+                                chain![arena;
+                                    arena.space(),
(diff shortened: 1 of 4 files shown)


vm/src/value.rs
```diff
@@ -202,6 +207,199 @@ impl Value {
...
+                ]
+            }
+            Value::Closure(ref closure) => {
+                chain![arena;
+                    "{",
+                    arena.concat(closure.upvars.iter().map(|field| {
+                        arena.text(format!("{:?}", field))
+                    })),
+                    "}"
+                ]
```
brendanzab: What does this look like without the macro? Is it that bad?

Marwes:
```
arena.text("{")
    .append(arena.concat(closure.upvars.iter().map(|field| {
        arena.text(format!("{:?}", field))
    }))
    .append("}")
```

Just gotten into the habit of using the macro for anything that needs more than a single append. The chain macro also scales better if more stuff are added to the printing (and in this case the debugger PR will actually add some extra information to printing a closure).

brendanzab: I'm going to add pretty printing to my toy type system, so I'll see if I can figure out a way to make the pretty printing read... prettier... 😆. For now.... ok. Just feels weird exporting the macro in `gluon_base`.

Marwes: Maybe it should be moved into the pretty crate itself? Since I made some changes to the library I need to publish a new version anyway before this can be merged.

PR #238, opened 2016-12-29
fix(check): Fill in the unified types of function arguments properly

Marwes:
Split out from #213


Marwes:
Split out from #213

Marwes:
There is a test for this but it is in #213

brendanzab:
Hah, was about to ask about the test!

PR #239, opened 2016-12-30
fix(parser): Don't loop forever from inserting default blocks

Marwes:
In the added test the lalrpop parser would enter error recovery and then repeatedly ask the lexer for tokens until a token that could succeed is found. Unfortunately due to the `,` token the layout algorithm would drop all contexts resulting in a default block to be added which again would be dropped (followed by the default block being added again etc.).

This is a kind of hack but I do believe that just returning the token is the best idea here since if it cannot be recoevered from then the parser will just drop it and request a new (real) token which has a better chance of succeding.

### DIFF
--- a/parser/src/layout.rs
+++ b/parser/src/layout.rs
@@ -190,54 +190,68 @@ impl<'input, Tokens> Layout<'input, Tokens>
                 (&Token::Comma, _) => {
                     self.indent_levels.pop();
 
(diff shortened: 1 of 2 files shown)


Marwes:
The default block insertion https://github.com/Marwes/gluon/blob/572e4b1ad3e2b360eeb783e2b60a88193be0f2b0/parser/src/layout.rs#L169-L173

PR #240, opened 2016-12-31
fix(parser): Remove unwrap on tokenizer errors

### LINKED ISSUE #210: Don't panic on tokenizer errors

Marwes:
The use of `RefCell` feels a bit icky but this is the least intrusive solution I could come up with. Most of the changes are just boilerplate due to the addition of two custom iterator adapters.

Fixes #210

Closes #214

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -9,6 +9,8 @@ extern crate quick_error;
 extern crate gluon_base as base;
 extern crate lalrpop_util;
 
+use std::cell::RefCell;
+
 use base::ast::{Expr, IdentEnv, SpannedExpr};
 use base::error::Errors;
 use base::pos::{self, BytePos, Span, Spanned};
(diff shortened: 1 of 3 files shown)


brendanzab:
Wonder if it would be possible to do this the `Iterator::scan` instead? Not sure though, but it does allow you to pass along a mutable context during iteration...

Marwes:
> Wonder if it would be possible to do this the Iterator::scan instead? Not sure though, but it does allow you to pass along a mutable context during iteration...

I don't see how that would help, how do you think it could be used?

The issue is that `Err` needs to be passed around the layout algoirthm thus forcing it to be shared between the layout and the result handling parts. That is how I looked at it anyway which led to the two iterator types I added

brendanzab:
I might have a mess around with it later, but good stuff! 👍

PR #241, opened 2017-01-04
feat: Integrate futures/tokio for asynchronous computations

Marwes:
This PR makes it possible to perform asynchronous computations in extern functions called from gluon. To make this relatively convenient the `FutureResult` type is provided which wraps a future and lets the virtual machine know that it needs to resolve the future to get hold of the value. `FutureValue` was added so that an event loop won't be required for normal, synchronous computations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ gluon_vm = { path = "vm", version = "0.2.2" }
 clap = "2.2.5"
 log = "0.3.6"
 quick-error = "1.0.0"
+futures = "0.1.0"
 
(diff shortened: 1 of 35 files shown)


Marwes:
Need to fix the compilation failure but otherwise there is only one problem I am aware of. For `Compiler::run_expr` and `Compiler::load_script` it might be a good idea to retain them as synchronous functions returning a `Result` and add additional `run_expr_async` and `load_script_async` functions for when a user wants more control. Its a bit of duplication but definitely more user friendly.

PR #242, opened 2017-01-07
fix(check): Only mark the span of `let` and `type` expressions body

Marwes:
Also fixes an issue with the implicit prelude which would cause a panic in source.rs

### DIFF
--- a/check/src/typecheck.rs
+++ b/check/src/typecheck.rs
@@ -49,9 +49,7 @@ pub enum TypeError<I> {
     /// Type is not a type which has any fields
     InvalidProjection(ArcType<I>),
     /// Expected to find a record with the following fields
-    UndefinedRecord {
-        fields: Vec<I>,
-    },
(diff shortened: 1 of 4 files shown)


check/src/typecheck.rs
```diff
@@ -1441,6 +1438,26 @@ fn primitive_type(op_type: &str) -> ArcType {
     }
 }
 
+/// Returns a span appropriate to mark a type error for `e`. For example the example below returns
+/// the `^` span instead of  `~`
```
brendanzab: Maybe it should say something about 'returns the span of the innermost type or value binding body'?

Marwes: Tried to make it a bit more abstract in case this function were extended produce different spans for expressions other than `let` and `type` but I can change to to only talk about those cases.

Marwes:
Think this was good now (fixed nit) so merged.

PR #243, opened 2017-01-07
feat(vm): Add a from_ut8 method for converting `Array Byte` into `String`

Marwes:
The bulk of these changes are to replace the use of `GcPtr<Str>` with a type backed by the normal array type. This makes it possible to convert `Array Byte` directly into `String` without copying any data.

Based on #235

### DIFF
--- a/check/src/unify_type.rs
+++ b/check/src/unify_type.rs
@@ -4,6 +4,7 @@ use std::mem;
 use smallvec::VecLike;
 
 use base::error::Errors;
+use base::fnv::FnvMap;
 use base::types::{self, AppVec, ArcType, Field, Type, TypeVariable, TypeEnv};
 use base::symbol::{Symbol, SymbolRef};
 use base::resolve::{self, Error as ResolveError};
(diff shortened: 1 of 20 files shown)


check/src/unify_type.rs
```diff
@@ -600,6 +601,34 @@ fn walk_move_types2<'a, I, F, T, R>(mut types: I, replaced: bool, output: &mut R
     }
 }
 
+/// Replaces all instances `Type::Generic` in `typ` with fresh type variables (`Type::Variable`)
+pub fn instantiate_generic_variables(named_variables: &mut FnvMap<Symbol, ArcType>,
```
brendanzab: Muuuuuch better!

vm/src/gc.rs
```diff
@@ -95,6 +95,32 @@ impl<'s> WriteOnly<'s, str> {
     }
 }
 
+#[derive(Clone, Copy, Default, Debug)]
+pub struct Generation(i32);
```
brendanzab: 👍

Issue #118, opened 2016-08-18
Add bindings to useful rust libraries


Marwes:
The current standard library is extremely minimal and so lacks a lot of stuff one may expect from a standard library. Creating and maintaining libraries are a ton of work though but if we use the embedding API it should prove much simpler (and we also get native performance for free!). 

[ ] [regex](https://crates.io/crates/regex)
[ ] [rand](https://crates.io/crates/rand)
[ ] A vector/math library (Found these from a quick crates.io search [vecmath](https://crates.io/crates/vecmath) [cgmath](https://crates.io/crates/vecmath) [nalgebra](https://crates.io/crates/nalgebra))

These libraries do not necessarily need to be part of the standard distribution but bindings to these (or others!) would be really useful.

Marking as `beginner` since this should not require to deep knowledge of the internals, see the [Tutorial  on embedding](https://github.com/Marwes/gluon/blob/master/TUTORIAL.md#embedding-api) and the [primitives](https://github.com/Marwes/gluon/blob/master/vm/src/primitives.rs) module for how this is currently done. Also, it should be easier to start exporting APIs which do not require `&mut` references.

Marwes:
Adding simple bindings to `regex`, starting with just `match : String -> String -> Bool` and building from there would be a good idea.

nxnfufunezn:
I would like to take this!

Marwes:
Great! I noticed I didn't mention where to put these modules. I think for now the best place is in the main `gluon` crate as a submodule as the [io module](https://github.com/gluon-lang/gluon/blob/master/src/io.rs) is now.

Don't hesitate to ask if you run into problems.

nxnfufunezn:
What type result should be `RuntimeResult<>` ?

Marwes:
`RuntimResult` is used to indicate a function which may do a 'gluon panic' such as in [array_index](https://github.com/gluon-lang/gluon/blob/e52f033e0148c591a8bb68190b7142050e4de675/vm/src/primitives.rs#L19-L26). Just as Rust would panic when doing an out of bounds index on a slice so do gluon (but 'gluon panic' are implemented without 'rust panic').

Only use it if you need a panicking function (such as `array_index`), otherwise you can return a `Result<T, E>` instead.

```
// Pseudo Rust
fn array_index(a: Array<A>) -> RuntimeResult<A, String>

// array_index : Array a -> a
```

Marwes:
[...]

PR #244, opened 2017-01-10
Binding to the `regex` crate

vbarrielle:
Hello,

I've been amazed by the simplicity and elegance of gluon and wanted to see if I could try and contribute a tiny bit. So here's my (very primitive) attempt to advance #118.

For starters I've simply binded a `matches: String -> String -> Bool`function, but this simple attempt leads me to ask a few question:

- I would have liked to name the funcion `match` as suggested in the issue, but I can't since it's a rust keyword
- right now the function will return `False` if given an invalid regex, but this seems bad from a usability standpoint. The alternatives are returning a `Result`, making the API a bit less smooth; or triggering a panic, which has its own downsides.
- if the binding is supposed to be optional, I have no idea yet on how to implement that.

If you think this is goind in a good direction, I'll try and expand upon the binding to try and have a nice API.


src/lib.rs
```diff
@@ -15,9 +15,12 @@ pub extern crate gluon_base as base;
 pub extern crate gluon_parser as parser;
 pub extern crate gluon_check as check;
 
+extern crate regex;
```
Marwes: This could be moved in the `regex_bind` module to make it more self contained. Remember to write `use self::regex::etc` to import stuff from it afterwards.

src/lib.rs
```diff
@@ -379,5 +382,6 @@ pub fn new_vm() -> RootedThread {
     ::vm::channel::load(&vm).expect("Loaded channel library");
     ::vm::debug::load(&vm).expect("Loaded debug library");
     ::io::load(&vm).expect("Loaded IO library");
+    ::regex_bind::load(&vm).expect("Loaded regex library");
```
Marwes: I think its fine to always import it for now as the library is so small anyway. Its easy enough to make this an optional import at a later point anyway

Cargo.toml
```diff
@@ -40,6 +40,9 @@ env_logger = { version = "0.3.4", optional = true }
 lazy_static = { version = "0.2.0", optional = true }
 rustyline = { version = "1.0.0", optional = true }
 
+# Binding crates
+regex = "0.2.1"
```
Marwes: Could you make this optional and then add `regex` as a default feature. That way it is possible to avoid the dependency if necessary.

Marwes:
* I would have liked to name the funcion match as suggested in the issue, but I can't since it's a rust keyword

Thats a mistake by me, `match` is also a keyword in gluon so that won't work anyway. Actually, name the function `is_match` to mirror rust, makes it more obvious what it is binding.

* right now the function will return False if given an invalid regex, but this seems bad from a usability standpoint. The alternatives are returning a Result, making the API a bit less smooth; or triggering a panic, which has its own downsides.

If you want to expand it a bit I would suggest adding `regex.new` (`fn new(s: &str) -> Result<Regex, regex::Error>`) and make `is_match` take a `Regex` as the first parameter. I think this should be relatively easy if you take a look at repl.glu for instance. You shouldn't need a `Mutex` since `Regex` has only `&` methods but you do need to wrap the regex in a newtype as `Editor` is https://github.com/gluon-lang/gluon/blob/b2da0a8bfb74cc0130c192c6739a8e409319a3b7/src/repl.rs#L124-L140. 

* if the binding is supposed to be optional, I have no idea yet on how to implement that.

Look at Cargo.toml, there are a few other crates that are optional. Adding `"regex"` to default (as `repl` is currently) can be nice as well. Then mark the regex module with `#[cfg(features = "regex")]`.

src/lib.rs
Marwes: If you want to make regex optional you will also need to write two wrapper load functions. Then that wrapper is callable regardless if the `regex` feature is active or not.

```rust
#[cfg(features ="regex")]
fn load_regex(vm: &Thread) {
 ::regex_bind::load(&vm).expect("Loaded regex library")
}
#[cfg(not(features ="regex"))]
fn load_regex(_: &Thread) {
}
```

Marwes:
@vbarrielle  Oh, and of course let me know if you have any other questions.

vbarrielle:
Hello, I've made the binding optional and started mirroring more closely the rust api with `new: String -> Result Error Regex`.
I still need to update the test but I'm having trouble dealing with `Result` in gluon. Is there no `unwrap` as in rust?

Marwes:
> I still need to update the test but I'm having trouble dealing with Result in gluon. Is there no unwrap as in rust?

There is not, though there should be. Maybe you can add it to `std/prelude.glu`? Maybe as `unwrap : Option a -> a`, `unwrap_ok : Result e t -> t`, `unwrap_err : Result e t -> e`? You can use `error : String -> a` to do the equivalent to `panic!` in rust.

If you want to return the proper `regex::Error` type from new I would suggest you also add a function `error_to_string : Error -> String` so it is possible to view the error message in gluon. Making Rust enums easy to use from gluon is somewhere on the todo list so it may get easier in the future (though no concrete solutions atm).

vbarrielle:
> There is not, though there should be. Maybe you can add it to std/prelude.glu? Maybe as unwrap : Option a -> a, unwrap_ok : Result e t -> t, unwrap_err : Result e t -> e? You can use error : String -> a to do the equivalent to panic! in rust.

Done, though no testing for these functions right now (apart from using `unwrap_ok` and `unwrap_err` in the regex tests), I don't know if there is some place for gluon tests?

> If you want to return the proper regex::Error type from new I would suggest you also add a function error_to_string : Error -> String so it is possible to view the error message in gluon.

Done as well.

If you wish, I can push the scope of this PR a bit further by binding the `Regex::find` method. That would bring a few questions though, for instance do we want to return a `Regex::Match` object, or do we want something simpler such as a tuple (though I'm not sure gluon if gluon has tuples).

Marwes:
If you have rustfmt installed could you run it on the files you added? Otherwise everything looks good!

> Done, though no testing for these functions right now (apart from using unwrap_ok and unwrap_err in the regex tests), I don't know if there is some place for gluon tests?

If you put a .glu file in `tests/pass` or `tests/fail` it will get tested (pass expects the file to execute successfully and fail that it caused a panic through calling `error` for example). There is a small test framework which you could use though it is not documented atm.

> If you wish, I can push the scope of this PR a bit further by binding the Regex::find method. That would bring a few questions though, for instance do we want to return a Regex::Match object, or do we want something simpler such as a tuple (though I'm not sure gluon if gluon has tuples).

That would be awesome! I looked over `regex`'s API and I also think `find` should be the next simplest method. Since `Match` is basically just `struct Match { begin: usize, end: usize, text: &str }` I think the best idea is to return a record `{ begin : Int, end : Int, text : String }`. It may be a bit tricky since I haven't gotten to documenting it yet but in #226 I use a few `record_*!` macros which gets marshalled into gluon records. Specifically https://github.com/gluon-lang/gluon/pull/226/files#diff-b6d24501cd10f418450b478457e6495dR99 and below. But don't worry if it seems to difficult since the documentation is lacking (going to explain it better throughout that example later on).

vbarrielle:
> If you put a .glu file in tests/pass or tests/fail it will get tested (pass expects the file to execute successfully and fail that it caused a panic through calling error for example).

I've done that, looks like it works.

> That would be awesome! I looked over regex's API and I also think find should be the next simplest method. Since Match is basically just struct Match { begin: usize, end: usize, text: &str } I think the best idea is to return a record { begin : Int, end : Int, text : String }. It may be a bit tricky since I haven't gotten to documenting it yet but in #226 I use a few record_*! macros which gets marshalled into gluon records.

Ok so if it's a bit tricky I'll have to wait until I have some time to do it. So maybe it can be left for another PR, as you wish.

Marwes:
Nice, thanks! Give me a ping on gitter or on an issue if you want to try your hand on something else!

PR #246, opened 2017-01-16
fix(compiler): Pop the correct amount of variables in the compiler

Marwes:
See compiler.rs:919-929. By using `Split` all fields in the value are pushed but when `Compiler::pop_pattern` is called to remove these variables it only removes the number of fields in the pattern (which may be fewer). To fix this I refactored the code a bit to use `ScopedMap` which is already used to do this kind of cleanup.

Extracted from #213, it has a few changes which aren't strictly necessary (new_stack_var taking a `&Compiler` argument) but is otherwise a pretty clean extraction.


Marwes:
Extracted this from #213 since #231 needs this fix, also found that the first fix were quite correct so I rebased a fix and a test on the commit.

PR #247, opened 2017-01-22
Prevent the REPL from panicking when parsing an overflowing integer

vbarrielle:
Hello, while playing with the REPL I noticed that it would panic when trying to parse a big integer:

```
$ ./target/release/repl -i
> 12345678901234567890
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: Overflow }', ../src/libcore/result.rs:837
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Any', ../src/libcore/result.rs:837
```

So I investigated a bit and found out that `numeric_literal` in `parser/src/token.rs` would `unwrap` the integers it tried to parse, causing the panic. So I've added a new `ErrorCode` in this module to represent integer parse errors. There's also an `unwrap` for floating point parsing, but I can't find a way to make it panic so I guess it's fine.

This does not propagate a proper error message to the top level though. However, the behavior is the same as for other kinds of integer parsing issues, so this is probably comes from a larger issue.

### DIFF
--- a/parser/src/token.rs
+++ b/parser/src/token.rs
@@ -141,6 +141,9 @@ quick_error! {
         UnterminatedStringLiteral {
             description("unterminated string literal")
         }
+        NonParseableInt {
+            description("cannot parse integer, probable overflow")
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


Marwes:
ah, were just going to fix those compile errors :) (caused by travis testing being before merging and then I never noticed travis erroring). Looks good, will merge after travis finishes.

Thanks!

Marwes:
What do you mean with

> This does not propagate a proper error message to the top level though. However, the behavior is the same as for other kinds of integer parsing issues, so this is probably comes from a larger issue.

With those changes I'd expected the error to be propagated (just as any other tokenizer error).

vbarrielle:
You're welcome. Later on I'd like to try and see if I can manage to get the cause of the error to reach the top level of the repl, right now all I get is something like `unexpected EOF`. However I'll need to understand the parsing pipeline a bit better :)

Marwes:
Guess that may be true since that isn't tested properly, just assumed it should work.

vbarrielle:
To be clearer about not reaching the top level, what happens is this:

```
> 12345678901234567890
<top>:Line: 1, Column: 1: Undexpected end of file
12345678901234567890
^
```

This is similar to what happens in toher integer parsing issues, eg:

```
> 12d2
<top>:Line: 1, Column: 1: Undexpected end of file
12d2
^
```

Marwes:
Just realized the cause, it should work fine if there is a token after the integer but if its the last token there is a quirk of how tokenizer errors are passed around the layout algorithm which means the error is forgotten. I think I will do the fix myself since that particular part of the parser is quite odd.

PR #248, opened 2017-01-22
fix(parser): Return tokenizer errors from the parser

Marwes:
Fixes the problem mentioned in #247 where tokenizer errors would not be propagated out from the parser


Marwes:
@vbarrielle The relevant fix is in the third commit if you are interested. It's a hack but works for now, probably in the future I may integrate error recovery into the tokenizer as well so that lexing can continue even after encountering a malformed token.

vbarrielle:
Interesting, thanks for the quick fix!

PR #258, opened 2017-02-18
style: using clap_app macro for readability

salpalvv:
I also fixed a spelling error in an error message.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -22,7 +22,7 @@ use gluon::vm::Error as VMError;
 
 mod repl;
 
-const GLUON_VERSION: &'static str = "0.3.0";
+const GLUON_VERSION: &'static str = env!("CARGO_PKG_VERSION");
 
 
 #[cfg(not(test))]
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -18,11 +19,11 @@ use gluon::{new_vm, Compiler, Thread, Error, Result};
 use gluon::vm::thread::ThreadInternal;
 #[cfg(not(test))]
 use gluon::vm::Error as VMError;
-#[cfg(not(test))]
-use clap::{Arg, App};
 
 mod repl;
 
+const GLUON_VERSION: &'static str = "0.3.0";
```
Marwes: You could use [CARGO_PKG_VERSION](http://doc.crates.io/environment-variables.html#environment-variables-cargo-sets-for-crates).

salpalvv: I don't have a strong opinion on the name nor whether the value should even exist. In my opinion, hard-coding the version in the macro just moves it into the code and obfuscates it. 

It wouldn't be troublesome to change it later, either way.

Issue #42, opened 2016-07-17
Suggest improvements to the documentation


Marwes:
There is still a lot of parts of gluon which lack documentation and while I will do my best to recognize and improve it where I can, it would be very helpful to get suggestions for which part I should prioritize. Opening a new issue for any needed documentation improvements is often still a good idea but consider this issue a place to add suggestions which do not seem to warrant a separate issue.

salpalvv:
I'd like to recommend generally that the functions and types you use in the tutorial should take priority in documentation. 

I can make specific issues but examples are `Compiler` struct with `new` and `run_expr` methods and `Thread` struct with `get_global` and `define_global` 

Also, adding the bit about importing your own `.glu` files with $GLUON_PATH would be helpful.

Marwes:
Definitely. I actually thought I had documented all functions as core as those already!

> Also, adding the bit about importing your own .glu files with $GLUON_PATH env variable would be helpful.

I am going to refactor the core of the macro expansion code pretty soon so I will add a more thorough explanation of the import macro then as well!

salpalvv:
Yeah, there is a short description for mostly everything. 

One documentation improvement would be to add short examples like [this](https://aturon.github.io/crossbeam-doc/crossbeam/struct.Scope.html)

I'd be happy to take some functions and add [examples + tests](https://doc.rust-lang.org/book/documentation.html) to make the docs more comprehensive. 

I could start with the functions used in the tutorial. I assume most people would gravitate to those first.

PR #259, opened 2017-02-18
[WIP] docs: Add examples to functions used in Tutorial

salpalvv:
Examples added to run_expr(), define_global(), and get_global() docs

`doc = false` so one may run `cargo doc --open` at the root 

For [#42](https://github.com/gluon-lang/gluon/issues/42)

Maybe adding examples to function documentation should be its own issue?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ path = "src/lib.rs"
 
 name = "gluon"
 path = "src/main.rs"
+doc = false
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


Marwes:
For examples in `vm/src/*.rs` you will have problems due to the `gluon_vm` crate not having access to `Compiler` (defined in the main `gluon` crate) so I am not really sure how to write doc tests for that crate yet :/.

PR #296, opened 2017-06-14
feat: Add a serializer and deserializer from/to gluon values

Marwes:
This makes it possible to easily marshal values to and from gluon by utilizing serde's serialization and deserialization.


Marwes:
@eldyer Feel free to test this out as a git dependency now, otherwise I will try to make a 0.5 release by the end of this week.

PR #312, opened 2017-07-08
[WIP] feat(parser): Let gluon files specify shebang

jgreene:
Not sure if this was the proper direction to go but I figured it was better to do it within the parser rather than before the parsing happened to help maintain proper line numbers for any tooling.

If this direction is OK I will probably tackle the pretty printer next.


parser/src/grammar.lalrpop
```diff
@@ -21,6 +21,7 @@ extern {
     type Error = Spanned<Error, BytePos>;
 
     enum Token<'input> {
+        "shebang line" => Token::ShebangLine(<String>),
```
Marwes: `&'input str` or empty (see later comment)

parser/src/layout.rs
```diff
@@ -173,6 +173,7 @@ where
         loop {
             // Retrieve the current indentation level if one exists
             let offside = match (&token.value, self.indent_levels.last().cloned()) {
+                (&Token::ShebangLine(_), _) => return Ok(token),
```
Marwes: One of the trickiest parts of the the compiler. But this should be good as shebang should not affect the layouting in any way. 👍

parser/src/token.rs
```diff
@@ -7,6 +7,7 @@ use self::Error::*;
 
 #[derive(Clone, PartialEq, Debug)]
 pub enum Token<'input> {
+    ShebangLine(String),
```
Marwes: You can use `&'input str` here for a (negligble) performance increase (more just to e consistent with the other variants).

Or thinking about it, since only the shell executing the file needs to know about it the variant could be empty instead. *Maybe* it could be useful in the future but do whatever seems simplest.

parser/src/token.rs
```diff
@@ -16,7 +17,7 @@ pub enum Token<'input> {
     ByteLiteral(u8),
     FloatLiteral(f64),
     DocComment(Comment),
-
+
```
Marwes: nit: trailing whitespace

parser/src/token.rs
```diff
@@ -364,6 +366,20 @@ impl<'input> Tokenizer<'input> {
         self.error(start, UnterminatedStringLiteral)
     }
 
+    fn shebang_line(&mut self, start: Location) -> Option<SpannedToken<'input>> {
+
+        let (end, line) = self.take_until(start, |ch| ch == '\n');
+        
+        if line.starts_with("#!") {
+            let skip = 2;
+            let result = line[skip..].trim_right().to_string();
```
Marwes: If you remove the `to_string` the you will get an `&'input str instead` (or omit it, see above)

parser/src/token.rs
```diff
@@ -473,6 +489,12 @@ impl<'input> Iterator for Tokenizer<'input> {
                         Err(err) => Some(Err(err)),
                     }
                 }
+                '#' if start.absolute == BytePos::from(0) && self.test_lookahead(|ch| ch == '!') => {
```
Marwes: `start.absolute == BytePos::from(0)` 👍

parser/src/token.rs
```diff
@@ -523,14 +545,19 @@ mod test {
 
     fn test(input: &str, expected: Vec<(&str, Token)>) {
         let mut tokenizer = tokenizer(input);
-
+        let mut count = 0;
+        let length = expected.len();
         for (token, (expected_span, expected_tok)) in tokenizer.by_ref().zip(expected.into_iter()) {
+            count = count + 1;
```
Marwes: nit: `count += 1`

parser/src/token.rs
```diff
@@ -523,14 +545,19 @@ mod test {
...
             println!("{:?}", token);
             let start = loc(expected_span.find("~").unwrap());
             let end = loc(expected_span.rfind("~").unwrap() + 1);
             assert_eq!(Ok(pos::spanned2(start, end, expected_tok)), token);
         }
 
+        assert_eq!(count, length);
+        assert_eq!(true, count > 0);
```
Marwes: 👍  Never noticed it failed to check that expected did not contain additional tokens

parser/tests/basic.rs
```diff
@@ -726,3 +726,15 @@ fn doc_comment_on_record_field() {
...
+fn shebang_at_top_is_ignored() {
+    let _ = ::env_logger::init();
+    let text = r"#!/bin/gluon
+{ Test, x }";
+    let e = parse_new!(text);
+    assert_eq!(
+        e,
+        record_a(vec![("Test".into(), None)], vec![("x".into(), None)])
+    )
+}
```
Marwes: nit: Missing newline

(A newline at the end of a file is nice because git will otherwise see the line with `}` as changed if something is added at the end in a later commit)

parser/src/token.rs
```diff
@@ -62,7 +64,7 @@ impl<'input> fmt::Display for Token<'input> {
             ByteLiteral(_) => "ByteLiteral",
             FloatLiteral(_) => "FloatLiteral",
             DocComment { .. } => "DocComment",
-
+
```
Marwes: nit: trailing whitespace

Marwes:
For the formatter just note that `self` has the `source` field available in which the source as it was before is stored which you can use to extract the line.

parser/src/layout.rs
jgreene: Yeah, the layout caused me a bit of grief since it auto inserted a Semi token which caused an Unexpected Token error.  I thought I was misunderstanding how lalrpop was supposed to work for a little bit.

I'm kind of wondering if it wouldn't make sense to have the tokenizer tests use the Layout so this type of issue is more easily visible.

Marwes: Sorry about that. I never thought about the layered approach to tokenizer + layout may be hard to discover. Had I figured you would do this change in the parser rather than as a preprocess I would have warned you (probably).

Stuffing it front and centre in the token tests may help but it feels wrong to inject something that should be irrelevant for the tests themselves there... Wonder if there is a cleaner way to explain it

jgreene: No worries, now that I know it shouldn't be a big deal in the future.  The layered approach is definitely the way to go overall it just had a bit of a learning curve.  This was my first rust code and my first time using lalrpop so I figured I was missing something.

Marwes:
Oh. And tests for pretty printing are in parser/tests/pretty_print.rs (since the parser is needed to parser what is then formatted)

jgreene:
I've added some pretty printer logic to handle the shebang line.

base/src/pretty_print.rs
```diff
@@ -264,14 +264,52 @@ impl<'a> ExprPrinter<'a> {
...
                 expr.span.end,
                 BytePos::from(self.source.src().len()),
             )),
         )
     }
 
+    fn find_shebang_line(&'a self) -> Option<&str> {
+        let src = self.source.src();
+        let len = src.len();
+        if len > 1 && &src[0..2] == "#!" {
```
Marwes: You can use `src.starts_with("#!")` here instead.

base/src/pretty_print.rs
```diff
@@ -264,14 +264,52 @@ impl<'a> ExprPrinter<'a> {
...
+            let newline_index = src.find('\n');
```
Marwes: Instead of find + indexing it is possible to use `src.lines().next()` to get the first line.

https://doc.rust-lang.org/std/primitive.str.html#method.lines

base/src/pretty_print.rs
```diff
@@ -264,14 +264,52 @@ impl<'a> ExprPrinter<'a> {
...
+            }
+        }
+        else {
+            None
+        }
+    }
+
+    pub fn pretty_expr_with_shebang_line<Id>(
+        &'a self,
+        previous_end: BytePos,
```
Marwes: This parameter is only ever 0 so it can  be removed.

base/src/pretty_print.rs
jgreene: No idea why I didn't go with this in the first place.  This is a whole lot cleaner.  I think my brain was assuming rust didn't have these methods built in.  VSCode autocomplete seems to break a lot so I probably should look over the APIs in the docs more.  Thanks for the tips.

Marwes:
Thanks!

PR #314, opened 2017-07-10
[WIP] feat(rust_format): integrate rust format with travis ci

jgreene:
Most files in this pull request have been modified by rust fmt.  

There are 5 files that haven't:
CONTRIBUTING.md
.travis.yml
rustfmt.toml
rust_format.sh -- formats all cargo packages
rust_format_diff.sh -- errors when rust_format.sh would make changes

I'm not sure how to test the travis ci setup without creating a pull request.  It is entirely possible that it will need some tweaking.


Marwes:
I think the check fails because https://github.com/rust-lang-nursery/rustfmt/pull/1705 isn't in crates.io yet so can't merge this before that.

Otherwise I think it is a good idea, I have been a bit sloppy with rustfmt code since the VSCode extension got a bug a month or so back which made formatting stop working.

rust-format-diff.sh
```diff
@@ -0,0 +1,8 @@
+for d in $(find . -name "Cargo.toml" -printf '%h\n')
+do
+    (cd $d; cargo fmt -- --write-mode=diff)
+    status=$?
+    if [ $status -ne 0 ] && [ $status -ne 101  ]; then # 101 is due to a bug in rustfmt that the ./vm package is triggering
+        exit $status;
+    fi
+done
```
jgreene: I looked for an option like that and wasn't able to find it.  I'll give it a try when I get a chance.

.travis.yml
```diff
@@ -14,10 +14,13 @@ rust:
 os:
 - linux
 before_script:
+- export PATH="$PATH:$HOME/.cargo/bin"
+- which rustfmt || cargo install rustfmt
```
Marwes: If you change this to `cargo +nightly install rustfmt-nightly` this PR should pass as that crate is the one that is updated from now on! Should have remembered that earlier.

May need to do `rustup install nightly` as well if it does not work on stable/beta.

https://crates.io/crates/rustfmt-nightly

Marwes:
Thought this may take a while to get working since rustfmt needed updating so I merged #313 before this which caused some conflicts as I moved the travis/test logic into its own directory (the root directory had become a bit crowded).

Marwes:
Closing since this has gone rather stale.

Marwes:
Probably still a good idea, just not a priority for me to update and merge this PR.

PR #319, opened 2017-08-24
Rename pure to wrap

brendanzab:
I saw this name used in [this blog post](http://joyfulmantis.github.io/posts/2017-08-23-functors-applicatives-monads.html) and thought it made more sense.


examples/http.glu
```diff
@@ -123,8 +123,8 @@ let catch_error action catch : Handler a -> (String -> Handler a) -> Handler a =
 /// Takes a `Handler` and a `Request` tries to process the request
 let handle handler state : Handler Response -> HttpState -> IO Response =
     let not_found _ _ =
-        pure { status = NotFound, body = "Page not found" }
-    handler (\response _ -> pure response) not_found state
+        wrap { status = NotFound, body = "Page not found" }
+    handler (\response _ -> wrap response) not_found state
```
brendanzab: This looks a bit clearer now in my opinion! Thoughts?

Marwes: Tentatively I'd agree. `pure` only makes any sense when looking at it as "embedding a pure" value in an impure calculation. But that is not how my thought process goes really as the "pure value" is often derived from a previous impure calculation (as is the case here with `response`) which makes it not feel very "pure". `pure` does have precendence from Haskell though (as do `return` but that is confusing for other reasons).

Let me digest it a bit before merging!

Also, hello again @brendanzab !

brendanzab: Yep, no worries, take your time!

I think the other nice thin is that it mirrors the `unwrap` from Rust, but I dunno... :/

I've been also thinking about how we could make extensible rows more useful using namespaced fields, but I'll save that for somewhere else!

vmchale: Would this really make things easier? As it stands, I can imagine it causing problems for people looking at docs from other languages (Haskell and Idris, in my case). I can't really see it making things *easier* for anyone.

brendanzab: I dunno - the name 'pure' is pretty impenetrable for most people outside the Haskell world. I added some more documentation here, with references to the haskell names https://github.com/gluon-lang/gluon/pull/319/commits/f8357afb325cf5b07011e5d6e7e8d87d3be53e74

brendanzab: In the original paper, [Applicative Programming with Effects](http://www.staff.city.ac.uk/~ross/papers/Applicative.pdf), the `pure` function is described as follows:

> The idea is that `pure` embeds pure computations into the pure fragment of an effectful world—the resulting computations may thus be shunted around freely, as long as the order of the genuinely effectful computations is preserved.

But it is still very weird to call the function `pure` because it's kind of doing the opposite - wrapping a pure value to convert it into some impure thing...

Marwes:
Seems to be a parse error in the type declaration. Opened an issue #321

examples/http.glu
Marwes: Ok, feeling better about `wrap` than `pure` at least and I haven't been able to find any other alternatives either.

While `pure` will be easier for people coming from Haskell I'd argue it is more intuitive for everyone else and anyone coming from Haskell will know to look in the `Applicative` type for the real name. Possibly we could provide `pure` as an alternative name for the same function in what is currently `make_Applicative` (or with implicits later on) though I'd be vary about providing to many such renamings.

Marwes:
@brendanzab Just so you know, it is failing on a formatting test as it is not possible to keep comments within types at the moment. I have a half-finished commit which will add position tracking to the explicit types in the AST which should fix it though.

brendanzab:
Yeah, was actually just looking at this. :)

brendanzab:
Should I just remove the f27607c commit @Marwes?

Marwes:
Yeah, do that, doc comments will be accepted on typea in the parser with #324 since that is no more work than just preserving normal comments

Marwes:
Enabled sudo on travis to get some extra memory which seems to have done the trick.

Marwes:
All good now! 👍

PR #320, opened 2017-08-24
feat: Implement universal quantification

### LINKED ISSUE #184: Add universally quantified types

Marwes:
Fixes #184
Closes #215


std/prelude.glu
```diff
@@ -207,13 +207,13 @@ let category_Function : Category (->) = {
     compose = \f g x -> f (g x)
 }
 
-let make_Category cat =
+let make_Category cat : Category cat -> _ =
```
brendanzab: Would it be possible to throw an error, saying that all generic types must be explicitly quantified? So you would need to write:

```
let make_Category cat : forall cat . Category cat -> _ =
```

Marwes: I went with inserting forall's implicitly since that is done anyway if one omits the type declaration.

```f#
let id x = x // works
let id x : a -> a = x // Error: `a` is not defined
let id x : forall a . a -> a = x // Have to add forall
```

It just seems a bit counterproductive that omitting the type entirely lets the type be inferred but typing in the variables causes it to fail... :/ Feel free to convince me otherwise though.

Marwes:
Currently working on adding skolemization (based on purescript) which seems to be the last missing part to get everything working.

From what I can gather about it, it is a way to "instantiate" a universally quantified type with fresh type variables without losing any information about the fact that the type were universally quantified in the first place. Or in other words, skolemization lazily inserts fresh type variables so that the `forall` only disappears if required by unification.

std/prelude.glu
brendanzab: https://github.com/purescript/purescript/issues/766#issuecomment-165842979

> I believe explicit scoping right off the bat helps prevent confusion when moving to higher-rank types and such. Hiding System F for a special (if common) case hides a key fact about how types really work.

https://github.com/purescript/purescript/issues/766#issuecomment-167426170

> I'd also suggest that many beginning-to-intermediate Haskell programmers have an inaccurate understanding of what something like `id :: a -> a` means. I'm not sure exactly how to describe it, but it's like they picture the preceding id function as being a function from some unknown type a to that same type.

brendanzab: I just really hate ML and Haskell for mysteriously magicking up unbound identifiers 😆

Marwes: I like to view it a bit like lifetime elision in rust. However, since type variables are scoped in gluon I'd say that is a really good argument for forcing `forall`...

On another note, `forall` is not a very good keyword to declare variables unless one already knows about universal quantification... Reminds me of `public static void main` in Java. 

> What does `forall` mean?

Uhh, it just lets you write declare generic paramaters.

> But what does it **mean** ?

**Points to stack of university literature and runs away**

brendanzab: It mean's 'for all types `a`, there is a function called `id` defined that goes from `a` to `a`'.

brendanzab: Personally I prefer a syntax that makes it look more like a function though. Like:

```
id: ?(a : Type) -> a -> a
```

That way it's clear that it's just another parameter, albeit at the type level.

Marwes: > It mean's 'for all types a, there is a function called id that goes from a to a'.

That is a good explanation but I still feel like that explanation can be difficult to grok from someone coming from Java, Rust etc.

> id: ?(a : Type) -> a -> a

I do like that on a theoretical level, but on a practical level I fear it makes generic functions more cumbersome than non-generic ones

```
// The first function is the better/more general function but it is much shorter to write the second...
// Each additional type parameter makes the difference larger as well
let id x: ?(a : Type) -> a -> a = x
let id x: Int -> Int = x
```

brendanzab: Perhaps we could still allow folks to drop the kind annotations. 🤔

Other langs allow folks to bunch the implicits together:

```rust
let compose : ?[a b c] -> (b -> c) -> (a -> b) -> a -> c
```

Then annotate them by name:

```rust
let gf = compose ?[c=Int] ?[a=Float] g f
```

Came across an interesting post on the problems with Agda's implicit arguments btw: https://github.com/bitonic/tog/wiki/Implicit-Arguments - all seems rather complicated :/

Marwes: > `let compose : ?[a b c] -> (b -> c) -> (a -> b) -> a -> c`

That looks really clean actually. It kind of splits the middle between `<A, B, C>/[A, B, C]` declared parameters and having a common syntax for types and values. (`[]` has some precedence as use for generic variables as well)

> Came across an interesting post on the problems with Agda's implicit arguments btw

Will take a look!

check/src/typecheck.rs
```diff
@@ -1670,6 +1674,7 @@ impl<'a> Typecheck<'a> {
                         })
                     }
                     Type::Forall(ref params, ref typ, _) => {
+                        println!("{:?}", self.type_variables);
```
brendanzab: Debugging stuff left in? 😆

check/src/unify_type.rs
```diff
@@ -873,6 +862,7 @@ impl<'a, 'e> Unifier<State<'a>, ArcType> for Merge<'e> {
             (&Type::Hole, _) => Ok(None),
             (&Type::Variable(ref l), &Type::Variable(ref r)) if l.id == r.id => Ok(None),
             (&Type::Generic(ref l_gen), &Type::Variable(ref r_var)) => {
+                println!("123 {:?}", unifier.unifier.variables);
```
brendanzab: Ditto

check/src/typecheck.rs
Marwes: Yah, it is in the last WIP commit :). Making sure it is backed up since I have had some computer issues recently.

base/src/types.rs
```diff
@@ -567,6 +598,9 @@ pub enum Type<Id, T = ArcType<Id>> {
     Alias(
         #[cfg_attr(feature = "serde_derive", serde(state))] AliasRef<Id, T>,
     ),
+    Skolem(
+        #[cfg_attr(feature = "serde_derive", serde(state))] Skolem<Id>,
+    ),
```
brendanzab: Maybe we should add some docs on these?

Marwes: Will do after everything is working.

check/tests/pass.rs
```diff
@@ -436,9 +432,9 @@ let function_test: Test ((->) a) = {
 function_test.test
 ";
     let result = support::typecheck(text);
-    let expected = Ok(Type::function(vec![typ("a")], typ("Int")));
+    let expected = Ok("forall a . a -> Int".to_string());
 
-    assert_req!(result, expected);
+    assert_req!(result.map(|typ| typ.to_string()), expected);
```
brendanzab: These stringified tests are much easier to read! 👏

Marwes: They are however not as accurate :/ Not a problem in the tests in question though.

Probably better to do it the other way around and parse a string into the type representation. Where that is possible at least.

Marwes:
Got ~3 tests still failing in the check crate and then likely a few more from compiling the prelude, etc. Been slow work on these as I haven't had much time the last few weeks. I am hopeful I can get all the tests passing this weekend though. Will need to do a fair bit of cleanup after that still however.

Marwes:
Gonna rubberduck here since I am rather stuck at the moment at the following example.

```f#
let f x : Int -> Int = x
let f x : String -> String = x
f 1
```

In this example the actual inference works exactly as I expect it to, however when we try to figure out the overload we find that the type of `f` in the last line is `forall a . a -> a`. This is actually correct and is what purescript would infer. However due to overloading being a thing `f` must be assigned the more specific type `Int -> Int` so we know which binding to load (this is not a problem in purescript as it only infers and checks and does not care what the specific type of a binding after inference is done).

Now, in an earlier version of this PR this actually worked! But it would mean that another example failed instead.

```f#
let make id =
    let id2 : forall a . a -> a = id
    id 2
    // Error `a` does not unify with `Int`
    id2

```

This earlier version would create type variables for each variable in `forall` right when the skolem scope was created instead of always creating fresh ones on each use of `id` as the current version does (and purescript does).

So since the variables were created and bound to `Type::Forall` they would be unified to `Int` and then the `forall` could be removed once we saw that they had be unified with a concrete type. But this version instead fails this second example as these variables would first be unified with the skolem variable `a` and then the same type variable would be unified with `Int`.

brendanzab:
Wondering if we are missing a notion of type abstraction. This is one of the things I'm currently wrangling my head around at work with my DDL. See section 5 in this: http://babel.ls.fi.upm.es/~pablo/Papers/Notes/f-fw.pdf

eg. the following:

```f#
let id : forall a . a -> a = \(x : a) -> x
id 1
```

...should expand to:

```f#
type Id : forall a . a -> a = \\(a : Type) -> a
let id : forall a . a -> a = \\(a : Type) -> \(x : a) -> x
id [a] 1
```

where `\\x : kind -> expr` is for type abstraction, and `expr [x]` is for type application.

Perhaps we need to do the same for types too?

Eg.

```f#
type Id (a : Type) = a
Id a
```
...would become:
```f#
type Id : Type -> Type = \\(a : Type) -> a
Id a
```
This would get you to System Fω.

Marwes:
> Wondering if we are missing a notion of type abstraction on expressions.

That might actually work. It may be a bit finicky to insert the type arguments on each application but not unduly so. Also, when implicit (value) arguments are added I guess that would need more or less the same thing so even if I feel a bit iffy about needing to modify the expressions by inserting additional arguments, that is something that will need to be done later anyway.

Marwes:
Its happening

Marwes:
So there is still a bit of work to be done to clean up duplication and debug printing I missed. That said, right now   am just going to be happy for a bit that I got through all the tests!

Marwes:
As there are no more know bugs with this and the fact that this ends up both fixing some bugs in the typechecker (some of which made me find bugs in the standard library) I am going to go ahead and merge this. I still need to do some clean up and add better/more comments but I want go and fix some other more important things and wait with that until later.

TUTORIAL.md
```diff
@@ -405,6 +406,64 @@ Higher-kinded types are a fairly abstract concept in gluon and you may create en
...
+}
+
+let module = Module {
+    id: Box::new(|x| x),
+};
+(module.id)(0);
+(module.id)("");
+```
+
+Alas, this does not work in Rust since `module` will be inferred to the type `Module<i32>` which makes the second call to `id` a type error. In gluon it works as the type of `module` is actually `{ id : forall a . a -> a }` and not `forall a . { id : a -> a }` which is the closest analogue to the Rust example.
```
brendanzab: Really nice description! 👏

brendanzab: Perhaps we could mention higher ranked lifetimes? ie. `for<'a> &'a T`?

base/src/types/mod.rs
```diff
@@ -240,6 +242,21 @@ pub struct TypeVariable {
...
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "serde_derive", derive(DeserializeState, SerializeState))]
+#[cfg_attr(feature = "serde_derive", serde(deserialize_state = "Seed<Id, T>"))]
+#[cfg_attr(feature = "serde_derive",
+           serde(bound(deserialize = "
+           Id: DeserializeState<'de, Seed<Id, T>> + Clone + ::std::any::Any")))]
+#[cfg_attr(feature = "serde_derive", serde(de_parameters = "T"))]
+#[cfg_attr(feature = "serde_derive", serde(serialize_state = "SeSeed"))]
+#[cfg_attr(feature = "serde_derive", serde(bound(serialize = "Id: SerializeState<SeSeed>")))]
```
brendanzab: Lol, these annotations are getting kinda hilarious.

base/src/types/mod.rs
```diff
@@ -240,6 +242,21 @@ pub struct TypeVariable {
 #[cfg_attr(feature = "serde_derive", serde(de_parameters = "T"))]
 #[cfg_attr(feature = "serde_derive", serde(serialize_state = "SeSeed"))]
 #[cfg_attr(feature = "serde_derive", serde(bound(serialize = "Id: SerializeState<SeSeed>")))]
+pub struct Skolem<Id> {
+    #[cfg_attr(feature = "serde_derive", serde(state))] pub name: Id,
+    pub id: u32,
+    #[cfg_attr(feature = "serde_derive", serde(state))] pub kind: ArcKind,
+}
```
brendanzab: Might be worth adding some doc comments to each of these types.

base/src/types/mod.rs
```diff
@@ -520,6 +545,12 @@ pub enum Type<Id, T = ArcType<Id>> {
     Opaque,
     /// A builtin type
     Builtin(BuiltinType),
+    /// Universally quantified types
+    Forall(
+        #[cfg_attr(feature = "serde_derive", serde(state))] Vec<Generic<Id>>,
+        #[cfg_attr(feature = "serde_derive", serde(state))] T,
+        #[cfg_attr(feature = "serde_derive", serde(state))] Option<Vec<T>>,
```
brendanzab: What is the `Option<Vec<T>>` here?

base/src/types/mod.rs
```diff
@@ -425,20 +442,19 @@ where
...
     #[cfg_attr(feature = "serde_derive", serde(state))]
     typ: T,
 }
 
-impl<Id, T> AliasData<Id, T> {
+impl<Id, T> AliasData<Id, T>
+where
+    T: From<Type<Id, T>>,
+{
     pub fn new(name: Id, args: Vec<Generic<Id>>, typ: T) -> AliasData<Id, T> {
```
brendanzab: Wondering wif we could simplify this now by removing `args` from the constructor, and explicitly using `Type::forall` where necessary.

base/src/types/mod.rs
```diff
@@ -558,6 +589,16 @@ pub enum Type<Id, T = ArcType<Id>> {
     /// when the binding is refered to.
     Generic(#[cfg_attr(feature = "serde_derive", serde(state))] Generic<Id>),
     Alias(#[cfg_attr(feature = "serde_derive", serde(state))] AliasRef<Id, T>),
+    Skolem(#[cfg_attr(feature = "serde_derive", serde(state))] Skolem<Id>),
```
brendanzab: So we didn't end up gaining a `Type::Abs` variant? How do we represent `type Id a = a`? Afaik, this should become `Id : Type -> Type = \a : Type -> a`...

std/prelude.glu
```diff
@@ -156,25 +160,27 @@ type Applicative (f : Type -> Type) = {
...
     ///
     /// * Known as `pure` in Haskell
-    wrap : a -> f a
+    wrap : forall a . a -> f a
 }
 
-// FIXME: Causes panic in VM: thread 'main' panicked at 'Op Split called on non data type', tests/main.rs:32:8
-// let make_Applicative applicative : Applicative f -> _ =
-let make_Applicative applicative =
+let make_Applicative applicative : Applicative f -> _ =
```
brendanzab: Yayy! 🎉

base/src/ast.rs
```diff
@@ -1,4 +1,5 @@
 //! Module containing the types which make up `gluon`'s AST (Abstract Syntax Tree)
+use std::fmt;
 use std::marker::PhantomData;
 use std::ops::Deref;
```
brendanzab: Wondering why we don't gain `Expr::TAbs` and `Expr::TApp` variants? How does this work for example:

```
id : forall a. a -> a = \x : a -> x
test = id 1
``` 
work?

Would have assumed we needed to expand to something like:

```
id : forall a. a -> a = \x : Type -> \x : a -> x
test = id Int 1
```

Wondering how Purescript represents this 🤔

Marwes: So the type of `id` is with some pseudo rust `Type::Forall(["a"], "a -> a", None)`. When we refer to `id` later in the code we add a new "skolem scope", this fills in the `None` part with fresh variables `Type::Forall(["a"], "a -> a", Some([$1]))`, one for each parameter to the `forall`. This basically amounts saying from which scope a variable originated so that we can check for escaping variables later.

This is purescript gets away with a slightly simpler implementation as it only marks this scope with an integer which bears no relation to the type variables. And that works fine for checking the types but falls a bit short when it comes to inferring the specialized type for each expression (ie `id : Int -> Int` rather than `id : forall a , a -> a` in the example above). This is where `TAbs` etc could have fixed the problem but I found it complicated to represent :/ . Instead, I inspect the variables bound in the `Some` part of the forall after type inference has finished and if a variable there has been unified, it gets removed from the `Type::Forall`. So since `$1` would be unified with `Int` we end up removing the forall completely and we get the specialized type.

It seems to work fine at least :P

base/src/types/mod.rs
Marwes: `Type::Alias` work fine to represent that case so there is no specific to have a specific `Type::Abs` case.

In core expressions lambdas actually gets desugared this `\x -> x` => `let fun x = x in func`.

base/src/ast.rs
Marwes: Maybe I am over exaggerating the complexity of `TAbs` but I didn't want to get into how to compute the types of `TAbs` and as I figured the solution above should work I thought that would be simpler.

Marwes:
@brendanzab  I will go through and update the docs per your comments later, now I just want this merged before I start getting conflicts with other PRs again.

brendanzab:
Good plan, sorry for all the last minute comments!

base/src/types/mod.rs
Marwes: Just now "getting hilarious" :) ? They have always been hilarious 🤡 

Its mostly the `cfg_attr` part that makes it messy though. Yesterday I shortened it a bit in https://github.com/gluon-lang/gluon/pull/373 and thought that maybe rustc could automatically treat custom attributes as wrapped in `cfg_attr` if the dependency is optional...

```rust
#[serde(de_parameters = "T")]
// => desugared to
#[cfg_attr(feature = "serde", serde(de_parameters = "T"))]
```

Might be worth floating the idea in internals though I think attributes are in a slightly hacky state right now which may make it hard to get implemented.

TUTORIAL.md
Marwes: > Perhaps we could mention higher ranked lifetimes? ie. for<'a> &'a T?

Might be a good anecdote, but I think most people don't even know what higher ranked lifetimes are so probably should not use it to explain anything.

base/src/types/mod.rs
Marwes: Yep, going to newtype/alias `u32` as well since it has started leaking out in a few to many places.

base/src/types/mod.rs
Marwes: I try to explain it in https://github.com/gluon-lang/gluon/pull/320#discussion_r148456145

PR #324, opened 2017-09-03
feat: Track positions of explicitly declared types

Marwes:
Bit of a hack to be able to reuse `enum Type` but it is fairly self-contained so it shouldn't be to bad in the end

Needed by #319 
cc @brendanzab


Marwes:
Should be done now, just need to run the travis gauntlet.

The doc comment storage is in a slightly weird place but it will do for now as any other place would require much more invasive changes.

PR #330, opened 2017-09-05
Add some error-biased instances for Result

brendanzab:
Cruelty to type systems. Sorry!

Purescript [seems to handle it](http://try.purescript.org/?backend=core&gist=4da33bd35f6673169e2fe4f1b983631c&session=c95cd67f-5c74-1cf1-b634-4dbccf405f09)!


Marwes:
I guess that should work, in theory at least 😛 .

Should probably have a better way of debugging formatting errors in the .glu files. The tests do generate a file if they find a mismatch and put it under `target/debug` so what I do is I use an external diff tool on that file and the current file under `std/`.

Of course, if you are brave you could always run `gluon fmt` on the file :) (it does store a backup file at least).

Marwes:
Needs rebase (to fix conflict and to get the travis sudo changes)

PR #335, opened 2017-09-10
Updates to prelude

brendanzab:


brendanzab:
For some reason I get the following test failure:

```
     Running target/debug/deps/main-9d09b68e5207c057

running 1 test
test main ... FAILED

failures:

---- main stdout ----
	test tests/pass/alternative.glu
test tests/pass/arithmetic.glu
test tests/pass/channel.glu
test tests/pass/deep_clone_userdata.glu
test tests/pass/lazy.glu
test tests/pass/lisp.glu
test tests/pass/list.glu
test tests/pass/map.glu
test tests/pass/parser.glu
test tests/pass/reference.glu
test tests/pass/state.glu
thread 'main' panicked at 'Op Split called on non data type', tests/main.rs:32:8


failures:
    main

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--test main'
```

brendanzab:
Seems to be the `make_Applicative` function that is causing it

Marwes:
> Seems to be annotating the make_Applicative function that is causing it

It is something to do with overloading, will take a look.

brendanzab:
Seem to be getting errors on nightly, but not related to this PR

Marwes:
@brendanzab  Compiletest 0.3 got released which should fix the error! Probably simplest you bump it in this PR and push to kick of another build, otherwise it will need a rebase anyway.

brendanzab:
Still seems to fail 😖

Marwes:
Yeah :/ Maybe it is something in travis so it only works locally... Will try and take a look tonight.

Marwes:
As for the overloading issue, it is possible that #320 would fix it so I will try to get that done first.

std/list.glu
```diff
@@ -0,0 +1,149 @@
...
+    let many_v _ = some_v () <|> wrap Nil
+    and some_v _ = functor.map (\h l -> Cons h l) x <*> many_v ()
+    many_v ()
+
+let some alternative x : Alternative f -> f a -> f (List a) =
+    let { applicative, (<|>) } = prelude.make_Alternative alternative
+    let { functor, (<*>), wrap } = prelude.make_Applicative applicative
+
+    let many_v _ = some_v () <|> wrap Nil
+    and some_v _ = functor.map (\h l -> Cons h l) x <*> many_v ()
```
Marwes: Duplicated from the `many` impl

std/prelude.glu
```diff
@@ -18,54 +15,180 @@ let unwrap_err res : Result e a -> e =
...
-/// * `append x empty == x`
-/// * `append empty x == x`
-/// * `append x (append y z) == append (append x y) z`
-type Monoid m = {
-    /// # Note
-    ///
-    /// * Known as `mappend` in Haskell
-    append : m -> m -> m,
+/// * `forall x . append x (append y z) == append (append x y) z`
+type Semigroup a = {
```
Marwes: Something which has been bugging me is where these types and implementations should live? The prelude has become a bit of a kitchen sink of implementions whereas it should probably just reexport most if its contents from elsewhere.

Perhaps `SemiGroup`, `Monad` etc should each live in their own file? Or if that is overkill then maybe. But then where should the implementations live (`Bool`, `Option` etc)? Should they be in `std/semigroup.glu` or in `std/bool.glu`. If they live in `std/bool.glu` (ie they live the the type _implementing_) then perhaps `std/semigroup.glu` is to small of a file to warrants its own existence and we should instead have `std/functional.glu` as a module for all these types.

Should probably open an issue to work these things out since I don't think it needs to be fixed in this PR!

Marwes:
Sooooo it works now ¯\\_(ツ)_/¯ . Merge or fix the nit I found at your discretion!

Issue #337, opened 2017-09-12
As-patterns


brendanzab:
I've noticed this pattern come up quite a bit:

```ml
let list = import! "std/list.glu"
let { List } = list
```

This could be alleviated if you could (like [in Haskell](https://en.wikibooks.org/wiki/Haskell/Pattern_matching#As-patterns) and [in Rust](https://doc.rust-lang.org/book/second-edition/ch18-03-pattern-syntax.html#-bindings)) give another name binding in patterns:

```ml
let list @ { List } = import! "std/list.glu"
```

Marwes:
Yep, definitely need this. Easy bit of sugar which removes lots of boilerplate given how frequent `let` bindings that unpack are used (since they double as import directives in Haskell, `use` in Rust).

Marwes:
Should also be added to match expressions (can be done separately if it complicates things)

```f#
match x with
| y @ Some _ -> y
| None -> z
```

brendanzab:
Isn't the lhs of a match arm just a pattern? I was thinking we'd add another variant to `ast::Pattern` called `As` or something...

Marwes:
> Isn't the lhs of a match arm just a pattern?

Yep, I guess I should have said it should work with any pattern which an `As` variant would do. Just wanted to document that it should work for patterns in `match` as well as `let` but due to `match` having a more complicated desugar it may not be as easy to get everything working there.

PR #338, opened 2017-09-12
Prelude breakups

brendanzab:
As mentioned in #335 at https://github.com/gluon-lang/gluon/pull/335#discussion_r138100396, this pulls more functionality out of the prelude and puts them into separate modules. Some ergonomics suffers as a result, but at the same time using some of the imported instances is starting to look prettier. Also the prelude looks much less intimidating. #337 should help on the ergonomics front too.

### DIFF
--- a/vm/src/primitives.rs
+++ b/vm/src/primitives.rs
@@ -244,7 +244,6 @@ extern "C" fn error(_: &Thread) -> Status {
 #[allow(non_camel_case_types)]
 pub fn load(vm: &Thread) -> Result<()> {
     use std::f64;
-    use std::char;
     type float = f64;
     vm.define_global(
         "float",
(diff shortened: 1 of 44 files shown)


Marwes:
https://github.com/gluon-lang/gluon/blob/572af1c53185182f1bdac34670b64fc98e11d1c0/src/import.rs#L50 needs an update as well. Need to figure out a way to distribute .glu files with `cargo install` so it can be removed.

brendanzab:
Now that's an interesting error...

Marwes:
Its a typevariable from a previous typecheck pass which wasn't "generalized" to a `Type::Generic` after typechecking concluded. As such its index was used in a later typecheck.

vm/src/primitives.rs
```diff
@@ -246,164 +248,167 @@ pub fn load(vm: &Thread) -> Result<()> {
...
-        len_utf16 => primitive!(1 char::len_utf16),
-        is_alphabetic => primitive!(1 char::is_alphabetic),
-        is_lowercase => primitive!(1 char::is_lowercase),
-        is_uppercase => primitive!(1 char::is_uppercase),
-        is_whitespace => primitive!(1 char::is_whitespace),
-        is_alphanumeric => primitive!(1 char::is_alphanumeric),
-        is_control => primitive!(1 char::is_control),
-        is_numeric => primitive!(1 char::is_numeric)
-    ),
+        "char_prim",
```
Marwes: Serialization requires that the name of the binding (`char::is_digit` is mapped to `char.is_digit`) matches what can be found in the global environment.

So here it fails because it tries to find `char.is_digit` but it fails because `char_prim` exists but not `char`. Easiest way to fix is to add `use std::char as char_prim` and rename the bindings below here to `char_prim::is_digit` etc. Long term we need a better solution to this `mod_prim + mod` naming.

Marwes: Or rename it back to `char` since that shouldnt cause any collision

PR #339, opened 2017-09-13
Move float and int instances into separate modules

brendanzab:


brendanzab:
Alas, this fails to run because there is some cyclic dependency. Not sure where though. 🤔

Marwes:
I'd recommend trying to improve the cyclic dependency error https://github.com/gluon-lang/gluon/blob/dac02eca4477cd786b6d467ca685fa0098168095/src/import.rs#L236. All the previously imported modules are already in the vec so it should be straightforward to generate a nicer error from it.

Marwes:
Ah, I misremembered thought all `std` modules were excluded from the implicit prelude but it was only `std.types` https://github.com/gluon-lang/gluon/blob/dac02eca4477cd786b6d467ca685fa0098168095/src/import.rs#L245

PR #342, opened 2017-09-23
Implement as-patterns

### LINKED ISSUE #337: As-patterns

brendanzab:
Fixes #337


brendanzab:
I'm unsure how to proceed in the VM. Here are the exhaustiveness failures I get:

```
λ cargo test --features test --all
   Compiling gluon_check v0.5.0 (file:///Users/brendan/dev/rust/gluon/check)
   Compiling gluon_vm v0.5.0 (file:///Users/brendan/dev/rust/gluon/vm)
error[E0004]: non-exhaustive patterns: `As(_, _)` not covered
    --> vm/src/core/mod.rs:1133:19
     |
1133 |             match equation.patterns.first().expect("Pattern").value {
     |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ pattern `As(_, _)` not covered

error[E0004]: non-exhaustive patterns: `As(_, _)` not covered
   --> vm/src/core/mod.rs:941:27
    |
941 |                     match equation.patterns.first().unwrap().value {
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ pattern `As(_, _)` not covered

error[E0004]: non-exhaustive patterns: `As(_, _)` not covered
    --> vm/src/core/mod.rs:1186:19
     |
1186 |             match pattern.value {
     |                   ^^^^^^^^^^^^^ pattern `As(_, _)` not covered

error: aborting due to 3 previous errors

error: Could not compile `gluon_vm`.

To learn more, run the command again with --verbose.
```

It would also be good to double check whether my changes to the typechecker make sense...

Marwes:
😍  First, these are links I used to implement the pattern match translator. Forgot to add a link to them in the translator http://marijnhaverbeke.nl/hob/saga/pattern_matching.html http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.257.6166&rep=rep1&type=pdf


```
error[E0004]: non-exhaustive patterns: `As(_, _)` not covered
    --> vm/src/core/mod.rs:1133:19
     |
1133 |             match equation.patterns.first().expect("Pattern").value {
     |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ pattern `As(_, _)` not covered
```

`varcon` main job is to differentiate between constructor patterns which imply branching and other patterns. I guess `at` patterns may fail into `CType::Ident`? Or its own variant maybe...

```
error[E0004]: non-exhaustive patterns: `As(_, _)` not covered
   --> vm/src/core/mod.rs:941:27
    |
941 |                     match equation.patterns.first().unwrap().value {
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ pattern `As(_, _)` not covered
```

Just group it under `unreachable!`

```
error[E0004]: non-exhaustive patterns: `As(_, _)` not covered
    --> vm/src/core/mod.rs:1186:19
     |
1186 |             match pattern.value {
     |                   ^^^^^^^^^^^^^ pattern `As(_, _)` not covered
```

The `pattern_identifiers` function extract one level of identifiers from a pattern to translate it into a core pattern (which is not nested). Since `x@_` is equivalent to `x` I would try and mimic `Pattern::Identifier`.

Another alternative would be to add a `case binder` to core patterns as Haskell's  core has (https://ghc.haskell.org/trac/ghc/wiki/Commentary/Compiler/CoreSynType). That should make the translation more straightforward.

Hope that helps somewhat. The pattern match translation is not very pretty. I need to clean it up some day...

Marwes:
Seems like we may be using different versions of rustfmt? I am using 0.2.6 but I see that 0.2.7 was just released. What are you using?

brendanzab:
```
λ rustfmt --version
0.9.0-nightly ( )
```
😆

check/src/completion.rs
```diff
@@ -266,6 +270,17 @@ where
             self.enclosing_match = Match::Pattern(current);
         }
         match current.value {
+            Pattern::As(ref id, ref pat) => {
+                let id_span = Span::new(
+                    current.span.start,
+                    current.span.start + BytePos::from(id.as_ref().len()),
+                );
+                if id_span.containment(&self.pos) == Ordering::Equal {
+                    self.found = Some(Some(Match::Pattern(current)));
```
Marwes: Make this `Match::Ident` instead. That way only the identifier part will be marked in editors.

Marwes:
Well that explains it :) Better update to `rustfmt-nightly` then!

parser/src/grammar.lalrpop
```diff
@@ -331,6 +332,12 @@ NoErrorPattern = {
 
         Pattern::Constructor(id, args)
     },
+
+    <id: Ident> "@" <pat: Sp<AtomicPattern>> => {
```
Marwes: Can it be `Pattern` instead of `AtomicPattern``? Otherwise `x@Some y` won't work. Or should we keep it as is and require `x@(Some y)`?

brendanzab: It will be nontrivial to make this work, I think, because of grammar ambiguities. I think it could be done if we mucked around with the rules, but I'd like to do that as a separate issue.

Marwes:
```
thread 'main' panicked at 'Undefined variable execState:178', vm/src/compiler.rs:524:31

note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Looks like desugaring misses a variable binding. Core expressions can actually be pretty printed which may help debug it (though it may have bugs so don't be to surprised if the printing looks a bit strange).

Marwes:
After looking at this again I believe the best idea is to ignore `@` patterns as part of pattern translator algorithm since this is how `@` work semantically. So anything matching on a `ast::Pattern` as part of the algorithm should be

```rust
fn unwrap_at(pattern: &ast::Pattern<Symbol>) -> &ast::Pattern<Symbol> {
    match *pattern {
        ast::Pattern::As(_, ref pattern) => unwrap_at(pattern),
        _ => pattern
    }
}

match unwrap_at(pattern) {
    ...
    ast::Pattern::As(..) => unreachable!()
}
```

Then in each call to [translate](https://github.com/Marwes/gluon/blob/8c8ffa6b135bf6cf83d08fbbd131e561a7815819/vm/src/core/mod.rs#L1287) (which translates a single pattern into a non-nested pattern) the `@` patterns of `first` (the pattern translated) need to be extracted and wrap the [returned expr](https://github.com/Marwes/gluon/blob/8c8ffa6b135bf6cf83d08fbbd131e561a7815819/vm/src/core/mod.rs#L1330) with `let` bindings.

```
match x with
| s@Some _ -> EXPR
| s2@None -> EXPR2
// ==>
let s = x
let s2 = x
match x with
| Some _ -> EXPR
| None -> EXPR2
```

I do hope this helps you make some progress but if not I can look into it (not sure when, but I'd like to get this in before the next release at least).

brendanzab:
Thanks for the extra help!

Marwes:
@brendanzab Are you working on this? If not I think I will try and finish it over the weekend before cutting a new release.

brendanzab:
Yeah, not got time at the moment - sorry. I've tried to keep it up to date though!

PR #347, opened 2017-10-03
[WIP] Move completion module to its own crate

### LINKED ISSUE #332: Split out the completion module to its own crate
The completion module only contains non-critical functionality for the core gluon compiler and interpreter and is only used by the repl and the language server. Thus we could just move it to its own crate.

- [ ] Create a new crate (`gluon_completion`?) next to the others
- [ ] Update the bash scripts with the new crate in https://github.com/gluon-lang/gluon/tree/master/scripts
- [ ] (Maybe) Split apart the completion module itself (860 loc)

dwnusbaum:
This is a WIP PR to move the completion module to a separate crate. I'm relatively new to rust, so please let me know if I'm approaching this in the wrong way.

- [x] Move check/tests/completion.rs to new crate. (Currently uses check/tests/support module so can't be moved trivially)
- [x] ~~Refactor completion crate's public API (Make internal types/functions private)~~ Looking at the module again, it is already hiding non-public functions and interfaces. I think it makes sense to save any further refactoring for another PR.
- [x] The issue mentions that the completion is used in the language server. Find out where this is, I only saw the completion code used in repl/, check/tests/completion.rs, and tests/vm.rs

Resolves #332


check/Cargo.toml
```diff
@@ -25,6 +25,8 @@ gluon_parser = { path = "../parser", version = "0.5.0", optional = true } # GLUO
 collect-mac = "0.1.0"
 pretty_assertions = "0.3.4"
 
+gluon_completion = { path = "../completion", version = "0.5.0" } # GLUON
```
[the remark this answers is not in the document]
dwnusbaum: I completely agree, I just wasn't sure on your preference on what to do with check/tests/support/mod.rs. If more tests are added and more duplication is needed, it could be split into a separate test-harness crate, but for now I think the duplication is a reasonable solution. I'll make the change.

completion/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[dependencies]
+gluon_base = { path = "../base", version = "0.5.0" } # GLUON
+
+[dev-dependencies]
+collect-mac = "0.1.0"
+env_logger = { version = "0.3.4" }
+
+gluon_check = { path = "../check", version = "0.5.0" } # GLUON
+gluon_parser = { path = "../parser", version = "0.5.0" } # GLUON
```
dwnusbaum: In other crates, you use optional dependencies sometimes for things that are only used in tests, and activate them using features. I couldn't find good info on doing that vs just making them dev-dependencies. Should I switch these to be optional and activated for a feature named `test`?

dwnusbaum:
Once the gluon_completion crate is published I'll update the [language server](https://github.com/gluon-lang/gluon_language-server) with the new dependencies.

Marwes:
> Once the gluon_completion crate is published I'll update the language server with the new dependencies.

Great! It should be a simple change. Only annoyance is that you will need to use a git dependency for testing it. After I publish 0.6 of gluon (which is overdue by now) I will change the language server's version to point to 0.6 again.

completion/Cargo.toml
Marwes: They should be dev-dependencies where it is possible. However, at least in the case of the `gluon_parser` crate it was necesssary to provide [this](https://github.com/gluon-lang/gluon/blob/e5671cabae13a44ee491994845bb536c317f9a34/parser/src/lib.rs#L416-L422) function specialization (where `Ident=String`). That function was only required for testing but to use it in tests defined in the `tests` directory I couldn't use plain `cfg(test)` since items `cfg` by that do not appear in the exported (`gluon_parser`) library. (Why I needed the specialization was because `gluon_parser` used to take 10+ minutes to compile in LLVM so I really needed it to be specialized immediately to avoid that cost when compiling gluon_check, etc).

Sorry if that is a bit confusing. It is fine to leave them as optional if like since --features=test will be necessary regardless for the time being at least.

Marwes: If you have already moved them to `dev-dependencies` and all the tests pass (which they do on travis!) then that is better than optional dependencies so leave them as is. There is simply no real need to go over and move other optional dependencies to dev since there is at least a couple of them which can't be a `dev-dependency`. So thus we are still forced to pass `--features=test` to run all the tests. :)

PR #351, opened 2017-10-05
Fix warnings raised when 'serialization' feature is not specified.

ordovicia:
When compiled without 'serialization' feature, some 'unused import' warnings occur
(I used rustc 1.20.0).
This PR fixes them by adding `cfg` attributes.


Marwes:
Thanks! I think I always compile with this feature set so I never noticed. I think I will try and enable warnings as errors on travis to make sure they get fixed.

PR #352, opened 2017-10-05
Move expr formatter to its own crate

### LINKED ISSUE #346: Split out code formatting to its own module
The formatting module only contains non-critical functionality for the core gluon compiler and interpreter and is only used by the repl and the language server. Thus we could move it to its own crate.

The type printing/formatting will need to stay in `gluon_base` since it is used to print error messages but the expression printing can be moved out.

- [ ] Create a new crate (`gluon_format`/`gluon_fmt`?) next to the others
- [ ] Update the bash scripts with the new crate in https://github.com/gluon-lang/gluon/tree/master/scripts
- [ ] (Maybe) Split apart the format module itself (800 loc) https://github.com/gluon-lang/gluon/blob/master/base/src/pretty_print.rs

cc #345

ordovicia:
This PR solves #346.

* Splits out expression formatter to newly created crate `gluon_format`.
* Update scripts for release.

It seemed to be difficult (or impossible) to me to split apart the whole format module
because it is used to print error message, you know.

When this PR is merged and new gluon is released, I will fix `gluon_language_server` and send PR.

Closes #346


ordovicia:
`env!("OUT_DIR")` in `format/tests/pretty_print.rs` does not work in Travis.
I am fixing this and retry CI.

Marwes:
> It seemed to be difficult (or impossible) to me to split apart the whole format module
because it is used to print error message, you know.

My bad! I forgot that I merged `struct Printer` so that type and expression printing both uses this type. Before that type printing had `TypePrinter` (a separate type) which should have made it relatively easy to move the `base/src/pretty_print.rs` wholesale.

Probably the simplest solution for now would be to add a newtype `struct Printer(::base::types::Printer);` to `gluon_format` with `std::ops::Deref<Target = ::base::types::Printer>` so it coerces to the type printer (unless I am missing some other dependency between `pretty_print.rs` and `types.rs`.

I can open that change as a separate issue though.

ordovicia:
I overlooked `travis.sh`. Thank you very much!
Now that `travis.sh` passed locally, I hope tests on Travis server will also pass.

ordovicia:
> Probably the simplest solution for now would be to add a newtype struct Printer(::base::types::Printer); to gluon_format with std::ops::Deref<Target = ::base::types::Printer> so it coerces to the type printer (unless I am missing some other dependency between pretty_print.rs and types.rs.

`::base::types::Printer` is a mistake for `::base::pretty_print::Printer`?

I could not understand this is really a solution.
Even if we add `Printer(::base::pretty_print::Printer)` to `gluon_format`, `gluon_base` still uses `base::pretty_print::Printer`, so gluon core will not be simplified at all.

Marwes:
> ::base::types::Printer is a mistake for ::base::pretty_print::Printer?

I meant that the struct `::base::pretty_print::Printer` would be moved to `::base::types::Printer` and then the expression printer could wrap that type with a new type (`struct Printer(::base::types::Printer);`).

Sorry for being unclear! Let me know if you want to finish up that last part, otherwise this looks good!

ordovicia:
Sorry for misunderstanding, likewise!

I have done the last part in the latest commit.

* Moved `base::pretty_print` module to `base::types::pretty_print`.
  * Created `types/mod.rs` and `types/pretty_print.rs` to avoid concatenating large files.
* `base::types` re-export `base::types::pretty_print::Printer`. 
* In `gluon_format`, added newtype struct `Printer(base::types::Printer)`.

Marwes:
One last thing I failed to explain (sorry)! This part https://github.com/ordovicia/gluon/blob/75aaf32959fbf42629171303d1a1650cbb4b0310/base/src/types/pretty_print.rs#L174-L797 is almost entirely pretty printing on expressions (rather than types) so it could be moved and implemented on `gluon_format::Printer` instead. There are a few functions dealing with whitespace and comments which are used in type printing as well so those need to still be in `gluon_base` but the rest can be moved out.

Marwes:
Great work! Thanks!

PR #357, opened 2017-10-11
Version 0.6.1

Marwes:
Found out that `not` weren't exported in the implicit prelude so I am releasing a point release to fix this. Due to that first fix requiring a way to restrict what modules import the implicit prelude I also cherry-picked #339.

Closes #339


std/float.glu
```diff
@@ -1,3 +1,5 @@
+//@NO-IMPLICIT-PRELUDE
+
```
brendanzab: lololololol

Marwes: Just making it ugly to make it clearer that it is not a permanent solution :P .

Long term I think we want a way to write attributes which can attach to `let`, `type` as well as top level (like #![no_std]) etc.

brendanzab: Yeah, I assumed that was the case.

PR #360, opened 2017-10-12
Add functional record update

### LINKED ISSUE #25: Add functional record update
Currently there is no good way of creating a new record which has most of the fields of an existing record with a few of them modified. Adding functional record update ala Haskell or Rust would make this easier in some cases.

Functiona record update may be a bit limiting in some/many cases so it may be worth investigating a more general way of updating data structures (something like lenses).

Marwes:
The improvements in `int.glu` etc speak for themselves but I think this is something we want regardless!

Closes #25


std/char.glu
```diff
@@ -13,15 +13,6 @@ let show : Show Char = { show = prim.show_char }
...
-    len_utf16 = char_prim.len_utf16,
-    is_alphabetic = char_prim.is_alphabetic,
-    is_lowercase = char_prim.is_lowercase,
-    is_uppercase = char_prim.is_uppercase,
-    is_whitespace = char_prim.is_whitespace,
-    is_alphanumeric = char_prim.is_alphanumeric,
-    is_control = char_prim.is_control,
-    is_numeric = char_prim.is_numeric
+    ..
+    char_prim
```
brendanzab: 👏

brendanzab:
Can the fields have different types to the original field?

Marwes:
> Can the fields have different types to the original field?

Oops, I were a bit to restrictive and currently disallow the base record to even have any fields in common with the constructed record. (I ran into a problem with `string_prim` defining `eq: String -> String -> Bool` whereas the `string` module needs `eq : Eq String` and so locked it down completely)

PR #368, opened 2017-10-18
Move function instances out of prelude

brendanzab:


std/prelude.glu
```diff
@@ -202,21 +165,15 @@ let make_Applicative applicative =
     let { functor, apply, wrap } = applicative
 
     let (<*>) : _ (a -> b) -> _ a -> _ b = apply
-    let (<*) l r : _ a -> _ b -> _ a = functor.map const l <*> r
-    let (*>) l r : _ a -> _ b -> _ b = functor.map (const id) l <*> r
+    let (<*) l r : _ a -> _ b -> _ a = functor.map (\x _ -> x) l <*> r
+    let (*>) l r : _ a -> _ b -> _ b = functor.map (\_ x -> x) l <*> r
```
Marwes: Did you try doing `import! "std/function.glu"` instead of writing lambdas? It should just work since you added the `NO-IMPLICIT-PRELUDE`.

brendanzab: Hmm, no, I didn't try. I thought that seeing as `std/function.glu` depends on `std/prelude.glu`, it wouldn't work 🤔

Marwes: To tired today, I just read the `NO-IMPLICIT-PRELUDE` part and missed that it still explicitly imported the prelude...

PR #371, opened 2017-10-29
Substitute each call to panic! with an ice! macro

### LINKED ISSUE #362: Replace all panic! calls with a macro which reports a gluon bug
Similarly to how rustc reports panic's as ICE (Internal compiler errors) we could add some extra information to each panic call to make it clear that it is a bug that should be reported.

Ie, substitute each call to `panic!` with an`ice!` macro (defined in `gluon_base`) which formats an error explaining that it is a bug and should be reported at https://github.com/gluon-lang/gluon/issues or gitter, etc

bostelk:
Fixes #362.

I am not too familiar with rust or the project, but here goes. I was a little selective in the substitution as I discarded changes to files outside of `**/src/*`. `cargo test` fails with a few seemingly unrelated errors, but `cargo build` succeeds.

The macro wraps `panic!`. I wasn't too sure if it should have been inlined, or not.

### DIFF
--- a/base/src/macros.rs
+++ b/base/src/macros.rs
@@ -1,0 +1,13 @@
+
+#[macro_export]
+macro_rules! ice {
+    () => ({
+        ice!("ICE: Please report an issue at https://github.com/gluon-lang/gluon/issues")
+    });
+    ($msg:expr) => ({
+        panic!(concat!($msg, ". Please report an issue at https://github.com/gluon-lang/gluon/issues"))
+    });
+    ($fmt:expr, $($arg:tt)+) => ({
+        panic!(concat!($fmt, ". Please report an issue at https://github.com/gluon-lang/gluon/issues"), $($arg)+)
+    });
+}
--- a/vm/src/api/de.rs
+++ b/vm/src/api/de.rs
@@ -65,7 +65,7 @@ let thread = new_vm();
 let (De(vec), _) = Compiler::new()
 #   .implicit_prelude(false)
     .run_expr::<De<Vec2>>(&thread, "test", "{ x = 1.0, y = 2.0 }")
-    .unwrap_or_else(|err| ice!("{}", err));
+    .unwrap_or_else(|err| panic!("{}", err));
 assert_eq!(vec, Vec2 {
         x: 1.0,
         y: 2.0
@@ -113,7 +113,7 @@ Compiler::new()
         "test",
         r#" type Enum = | A Int | B String Float in { Enum } "#,
     )
-    .unwrap_or_else(|err| ice!("{}", err));
+    .unwrap_or_else(|err| panic!("{}", err));
 
 let (De(enum_), _) = Compiler::new()
 #   .implicit_prelude(false)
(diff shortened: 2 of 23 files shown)


vm/src/core/mod.rs
```diff
@@ -18,7 +18,7 @@ macro_rules! assert_deq {
         match (&($left), &($right)) {
             (left_val, right_val) => {
                 if !(*left_val == *right_val) {
-                    panic!("assertion failed: `(left == right)` \
+                    ice!("assertion failed: `(left == right)` \
```
Marwes: This assert is only used in tests so they aren't really an ice and could remain as panic.

vm/src/core/mod.rs
```diff
@@ -8,7 +8,7 @@ macro_rules! assert_deq {
         match (&$left, &$right) {
             (left_val, right_val) => {
                 if !(*left_val == *right_val) {
-                    panic!("assertion failed: `(left == right)` \
+                    ice!("assertion failed: `(left == right)` \
```
Marwes: ditto

vm/src/api/ser.rs
```diff
@@ -110,13 +110,13 @@ let f e =
...
-    .unwrap_or_else(|err| panic!("{}", err));
+    .unwrap_or_else(|err| ice!("{}", err));
 
 let mut f: FunctionRef<fn (Ser<Enum>) -> i32> = thread
     .get_global("test.f")
-    .unwrap_or_else(|err| panic!("{}", err));
+    .unwrap_or_else(|err| ice!("{}", err));
 
-let result = f.call(Ser(Enum::B("".to_string(), 4))).unwrap_or_else(|err| panic!("{}", err));
+let result = f.call(Ser(Enum::B("".to_string(), 4))).unwrap_or_else(|err| ice!("{}", err));
```
Marwes: These are doc tests so they should still be panic! (since they display an example on what to write).

vm/src/api/mac.rs
```diff
@@ -18,7 +18,7 @@ macro_rules! primitive_cast {
 /// #[macro_use]
 /// extern crate gluon_vm;
 /// fn test(_x: i32, _y: String) -> f64 {
-///     panic!()
+///     ice!()
```
Marwes: doc test

vm/src/api/de.rs
```diff
@@ -133,7 +133,7 @@ let (De(enum_), _) = Compiler::new()
         "test",
         r#" let { Enum } = import! "test" in B "abc" 3.14 "#,
     )
-    .unwrap_or_else(|err| panic!("{}", err));
+    .unwrap_or_else(|err| ice!("{}", err));
```
Marwes: Doc test (as are the other panic/ice in this file)

Marwes:
Thanks!

base/src/macros.rs
```diff
@@ -0,0 +1,13 @@
+
+#[macro_export]
+macro_rules! ice {
+    () => ({
+        ice!("ICE: Please report an issue at https://github.com/gluon-lang/gluon/issues")
```
FortressBuilder: @bostelk @Marwes This should probably be "panic!", as otherwise the "Please report ..." will be printed twice.

bostelk: Oops, you're right. Fixed in #386.

PR #372, opened 2017-11-01
test: Try enabling skeptic again

Marwes:
Skeptic has been updated a bit (as have rustc) so lets see if we can run skeptic tests again


Marwes:
Still broken :(

PR #374, opened 2017-11-02
fix(repl): Run io expression when running files from the command line

### LINKED ISSUE #365: no output on command line
there is no output 👍 

$ cat t.glu 

```
let prelude = import! "std/prelude.glu"
let io = import! "std/io.glu"

let run x : () -> IO () =
    io.println "gluon (:h for help, :q to quit)"
run
```

$ ./target/debug/gluon t.glu

Marwes:
Fixes #365


repl/tests/print.glu
```diff
@@ -0,0 +1,2 @@
+let io = import! "std/io.glu"
+io.println "123"
```
brendanzab: Missing trailing new line on this file 😆

Marwes: 😞 I thought I ran with editorconfig to ensure this would never be a problem... Maybe the something broke in a recent update.

PR #375, opened 2017-11-02
feat(check): Let type aliases be referred via projection

### LINKED ISSUE #192: Allow types to be referred to through by projection
Allow something like this to be written

```
let prelude = import "std/prelude.glu"
let x : prelude.Option Int = ...
x
```

Instead of

```
let { Option} = import "std/prelude.glu"
let x : Option Int = ...
x
```

### LINKED ISSUE #370: ICE when trying to use record field as a type
This program
```
type Foo a = bar.Baz a
{}
```
crashes gluon:
```
thread '<unnamed>' panicked at 'ICE: Alias group were not able to resolve an identifier', /checkout/src/libcore/option.rs:839:4
```

Looks like gluon does not allow specifying types via field access at all. In other situations field as a type produces 'type does not exist' or unification errors, which makes me think that perhaps this syntax should be rejected by the parser?

Marwes:
Since we want to provide the correct spans in the errors I had to hook into the ast type translation which was slightly awkward (need to clone `type_cache` to make lifetimes work).

## Example

```f#
type Test = Int
let module = { Test }
let x : module.Test = 1
x
```

Closes #192

Fixes #370

### DIFF
--- a/base/src/types/mod.rs
+++ b/base/src/types/mod.rs
@@ -2010,18 +2010,16 @@ where
     }
 }
 
-pub fn translate_alias<Id, T, U>(
-    cache: &TypeCache<Id, U>,
-    alias: &AliasData<Id, T>,
-) -> AliasData<Id, U>
+pub fn translate_alias<Id, T, U, F>(alias: &AliasData<Id, T>, mut translate: F) -> AliasData<Id, U>
 where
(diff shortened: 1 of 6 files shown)


Marwes:
I realized that it was actually quite simple to implement this so I figured I might as well do it instead of removing rule from the grammar!

Issue #340, opened 2017-09-15
Rewrite how global values are handled


Marwes:
Currently the gluon vm stores a single `HashMap` of all global values defined via `define_global`, `load_script`, `import!`. These values can then be retrieved in the bytecode via the `GetGlobal` instruction.

All of this works works as intended right now but it suffers from some limitations such as making it impossible to replace or remove (#293) previously defined values from the environment. This is because removing or replacing any value may break bytecode which has been typechecked and compiled under the assumption that the value it retrieves via `GetGlobal` exists and has a specific type.

There is also a disconnect in that primitives are currently defined via `define_global` and must then be retrieved via implicit global variables which can be accessed from any gluon file, thus exposing implementation details https://github.com/gluon-lang/gluon/blob/dac02eca4477cd786b6d467ca685fa0098168095/vm/src/primitives.rs#L349-L370  https://github.com/gluon-lang/gluon/blob/dac02eca4477cd786b6d467ca685fa0098168095/std/string.glu#L4.

## How to fix

Completely remove the global table and the `GetGlobal` instruction. Instead move all of this logic into the `import!` macro. As the top level expression is currently compiled into a zero argument closure all "globals" will instead be inserted into the environment of this closure and accessed through. By doing this instead, each and every compiled module will hold on the "global" values it requires and no longer have any dependency on global state.

Since bytecode no longer refers to global state module any loaded modules can then bee freely replaced or removed at runtime

Finally the `import!` macro will be generalized to support registering `fn load(thread: &Thread) -> Result<()> { }` functions, letting them be loaded through the same framework as normal gluon files.

Marwes:
Closed by #383

PR #383, opened 2017-11-05
feat(vm): Rewrite global handling to be abstracted from users

Marwes:
Will tackle #340 as well as some other related things with global variables (see  e26c12803c4f691ae895d5176c356ec8cd4c08bd).

🚲 🏠 

* What should `string_prim`, etc be called?
* Should we hide them away with more obviously internal names or lock them away with something like `@NO-IMPLICIT-PRELUDE`?
* `array`, `reference`, `lazy` are full modules meant to be used. Should we perhaps put make them loadable as `import! "std.array"` etc? (Probably)
* Since we have non-file modules now. Should we call `import!` as `import! "std.bool"` instead of `import! "std/bool.glu"`?
  * `import! "std.bool"` vs `import! std.bool`?

### DIFF
--- a/repl/src/repl.rs
+++ b/repl/src/repl.rs
@@ -18,13 +18,13 @@ use base::pos;
 use base::symbol::{Symbol, SymbolModule};
 use base::types::ArcType;
 use parser::parse_partial_let_or_expr;
-use vm::Error as VMError;
+use vm::{self, Error as VMError};
 use vm::api::{Function, Userdata, VmType, WithVM, IO};
 use vm::gc::{Gc, Traverseable};
 use vm::internal::ValuePrinter;
 use vm::thread::{RootStr, RootedValue, Thread, ThreadInternal};
 
-use gluon::{new_vm, Compiler, Result as GluonResult, RootedThread};
+use gluon::{add_extern_module, new_vm, Compiler, Result as GluonResult, RootedThread};
 use gluon::compiler_pipeline::{Executable, ExecuteValue};
 
 fn type_of_expr(args: WithVM<RootStr>) -> IO<Result<String, String>> {
@@ -268,32 +268,35 @@ fn set_globals(
     }
 }
 
-fn compile_repl(vm: &Thread) -> Result<(), Box<StdError + Send + Sync>> {
+fn load_rustyline(vm: &Thread) -> vm::Result<vm::ExternModule> {
     vm.register_type::<Editor>("Editor", &[])?;
 
-    vm.define_global(
-        "rustyline",
-        record!(
-        new_editor => primitive!(1 new_editor),
-        readline => primitive!(2 readline)
-    ),
-    )?;
-    vm.define_global(
-        "repl_prim",
-        record!(
-        type_of_expr => primitive!(1 type_of_expr),
-        find_info => primitive!(1 find_info),
-        find_kind => primitive!(1 find_kind),
-        eval_line => primitive!(1 eval_line)
-    ),
-    )?;
+    vm::ExternModule::new(record!(
+            new_editor => primitive!(1 new_editor),
+            readline => primitive!(2 readline)
+        ))
+}
+
+fn load_repl(vm: &Thread) -> vm::Result<vm::ExternModule> {
+    vm::ExternModule::new(record!(
+            type_of_expr => primitive!(1 type_of_expr),
+            find_info => primitive!(1 find_info),
+            find_kind => primitive!(1 find_kind),
+            eval_line => primitive!(1 eval_line)
+        ))
+}
+
+fn compile_repl(vm: &Thread) -> Result<(), Box<StdError + Send + Sync>> {
+
+    add_extern_module(vm, "repl.prim" load_repl);
+    add_extern_module(vm, "rustyline", load_rustyline);
 
     const REPL_SOURCE: &'static str =
         include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/src/repl.glu"));
 
     let mut compiler = Compiler::new();
     compiler
-        .load_script_async(vm, "std.repl", REPL_SOURCE)
+        .load_script_async(vm, "repl", REPL_SOURCE)
         .sync_or_error()?;
     Ok(())
 }
@@ -302,7 +305,7 @@ fn compile_repl(vm: &Thread) -> Result<(), Box<StdError + Send + Sync>> {
 pub fn run() -> Result<(), Box<StdError + Send + Sync>> {
     let vm = new_vm();
     compile_repl(&vm)?;
-    let mut repl: Function<&Thread, fn(()) -> IO<()>> = vm.get_global("std.repl")?;
+    let mut repl: Function<&Thread, fn(()) -> IO<()>> = vm.get_global("repl")?;
     debug!("Starting repl");
     repl.call(())?;
     Ok(())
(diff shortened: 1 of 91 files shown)


endeav0r:
I have a bunch of primitive types in my bindings, which I wrap with gluon. I call them, `falcon_il_prim` and `falcon_loader_prim`, etc. This works, well, so maybe `gluon_string_prim` would be a good name for `string_prim`.

I would recommend not locking things away. Deconflict names, but don't lock things away. Provide _convenience_ around these concepts, but leave them exposed in convenient namespaces.

I would recommend going _the other direction_ as to making things load with `import!`. As an embeddable scripting language, it would be helpful if more things are loaded by default. There was a recommendation for a `let {*} = something` a while ago. If this happens, I will create something along the lines of `default.glu`, and do `let {*} = default.glu`, to clean up all of this stuff at the top of my scripts.

std/char.glu
```diff
@@ -14,5 +14,5 @@ let show : Show Char = { show = prim.show_char }
     ord,
     show,
     ..
-    char_prim
+    import! "char_prim"
```
brendanzab: Oh, this is nice!

brendanzab:
I do like the idea of `import! std.bool`. Perhaps we could also have `import! std.bool.prim`?

brendanzab:
One other thought: perhaps we could make `import` more 'sugary'. Eg.

```
import! std.function { (|>) }
import! std.io as io
import! std.list as list { List }
import! std.map as map { Map }
import! std.prelude as prelude { Ord }
import! std.string as string
import! std.types { Option, Bool }
```

This would expand to:

```
let { (|>) } = load! std.function
let io = load! std.io
let list @ { List } = load! std.list
let map @ { Map } = load! std.map
let prelude @ { Ord } = load! std.prelude
let string = load! std.string
let { Option, Bool } = load! std.types
```

Marwes:
> I would recommend not locking things away. Deconflict names, but don't lock things away. Provide convenience around these concepts, but leave them exposed in convenient namespaces.

Just so we are on the same page. I'd like to hide away `int_prim`, `float_prim`, `prim.show_int` as all of the functionality is re-exported in a more consistent interface (`std.int`, `std.float`, `std.int.show.show`). So no functionality is removed, I just want to remove/limit access to the implementation details.

> If this happens, I will create something along the lines of default.glu, and do let {*} = default.glu, to clean up all of this stuff at the top of my scripts.

I like that idea, it is not implicit but it is as close to it as it can be while still avoiding the pitfall of having two different preludes conflict with each other which would be possible if the current implicit prelude where exposed to be used outside being an implementation detail.

Marwes:
> One other thought: perhaps we could make import more 'sugary'. Eg.

Yeah, I have been thinking that as well. It won't be trivial though as the way macros are parsed need to be changed (they are currently parsed as function applications so it is not possible to put in patterns).

Arguably, we could make `import!` a part of the normal syntax (adding a `import` keyword). The main reason it is a macro is just to dog-food the macro system and make it powerful for users (though it is basically undocumented right now).

If we keep it as a macro I think we need to hack the tokenizer to handle macros (identifiers ending with `!`) specially. In a sense the macro would need state to what it expects (expression, pattern, keyword) and then the tokenizer calls back into the parser (starting a new parse) and then finally when it is done it would produce a single token (`Token::Macro(Macroparse)`) to the calling parser. Perhaps this could be done more elegantly but then it would likely require some non-trivial changes to LALRPOP.

brendanzab:
Kind of think just making it syntactic sugar would be easiest.

Marwes:
> Kind of think just making it syntax would be easiest/nicest.

Definitely. The only slight annoyance is that we may very well end up extending macro to work like I mentioned, in which case the syntax may be seen as redundant (though I suppose that could be said for just about any syntax sugar).

Marwes:
What if we add `<EXPR> as <PATTERN> in <EXPR>` then it would be possible to write

```f#
import! std.string as { find } in find "2" "123"

import! std.int as { signum }
signum 1 
```

src/lib.rs
```diff
@@ -562,7 +562,7 @@ pub fn new_vm() -> RootedThread {
     add_extern_module(&vm, "float_prim", ::vm::primitives::load_float);
     add_extern_module(&vm, "string_prim", ::vm::primitives::load_string);
     add_extern_module(&vm, "char_prim", ::vm::primitives::load_char);
-    add_extern_module(&vm, "array", ::vm::primitives::load_array);
+    add_extern_module(&vm, "std.array", ::vm::primitives::load_array);
```
brendanzab: I'm wondering if it would be a good idea to have an `std/array.glu` file, simply for documentation purposes. Would reexport the stuff from `array_prim`.

Also, I'm wondering if it would be better to call these prim modules `std.float.prim`?

Marwes: > I'm wondering if it would be a good idea to have an std/array.glu file, simply for documentation purposes. Would reexport the stuff from array_prim.

True, the `Array` type needs a bunch of functionality such as `Functor`, `Eq` etc anyway so that file will be needed regardless.

> Also, I'm wondering if it would be better to call these prim modules std.float.prim, etc?

That sounds nicer, but I am worried it may be more easier to discover. Perhaps I should just make them impossible to load except by the standard library. Or I can punt that worry until later.

brendanzab: We will need some pattern for libraries like those that @endeav0r is working on, so it would probably be a good idea if it the same thing could apply there. Elm uses `*.Native` as a pattern to denote JS code, and requires library authors to specify the modules they expose via their `elm-package.json`.

repl/src/repl.rs
```diff
@@ -268,32 +268,35 @@ fn set_globals(
...
+            find_info => primitive!(1 find_info),
+            find_kind => primitive!(1 find_kind),
+            eval_line => primitive!(1 eval_line)
+        ))
+}
+
+fn compile_repl(vm: &Thread) -> Result<(), Box<StdError + Send + Sync>> {
+
+    add_extern_module(vm, "repl.prim" load_repl);
+    add_extern_module(vm, "rustyline", load_rustyline);
```
brendanzab: I'm thinking that this should be `rustyline.prim` too, for consistency. 🤔

Marwes: I think its fine to omit the `prim` part if the "primitive" module exports the entire interface as it is meant to be used as there is no difference when writing `import!`.

Marwes: 🤔 It is debatable wheter `prim` is a good name for it in any case. It is more or less an artifact of "primitive operations" being what I called things I implemented with rust functions rather than adding op codes for.

brendanzab: Other options:

- `.ffi`
- `.extern`
- `.native`

Marwes: Maybe `extern` is better? It might be a bad idea though if `extern` we end up wanting to declare "`extern function signatures" (C or Rust function) in gluon code. (Might be useful as a way to declare signatures of functions provided with a dynamic library).

I am leaning towards opening an issue and punting on it. Since it does not need to be a public API and any user can name the module whatever they want. If we want to "standarize" `prim/ffi/extern/native` then we are better of doing that in conjunction with treating that module specially in some way.

Marwes:
@brendanzab Any objections? I feel the user facing changes are all there now so I think I am going to go ahead and merge (a few other things I want to move on to).

brendanzab:
🎉

Issue #385, opened 2017-11-07
Add hex literals


Marwes:
These should be supported for both integer literals (`0xdeadbeef`) and byteliterals. Unfortunately byte literals are defined with a trailing `b` (`123b`) which makes it ambiguous. So we need a new way to write byte literals or make normal integer literals convertible to bytes in inference.

Doing it for just integers should be a straightforward affair though. The number tokenizing is done at https://github.com/gluon-lang/gluon/blob/3f78fe2ddc2ed06f5c09995930bdd1fb7e0449f7/parser/src/token.rs#L406-L431 so there just needs to be an extra case which detects and parses hex literals into an `i64`.

brendanzab:
Could be `0b...` for byte literals.

Marwes:
That's bit literals though `0b10101011` (which may also be useful) but they don\t solve the problem of figuring out wheter `123` is `Int` or `Byte`.

brendanzab:
Ohhhhh! I see the problem now. Sorry, I was mixed up with binary-integer and byte literals!

PR #393, opened 2017-11-18
parser: Adjust parser to accept integer hex literals

zummenix:
#385, This supports only integer hex literals.

### DIFF
--- a/parser/src/token.rs
+++ b/parser/src/token.rs
@@ -461,10 +461,17 @@ impl<'input> Tokenizer<'input> {
             }
             Some((end, 'b')) => {
                 self.bump(); // Skip 'b'
-                if let Ok(val) = int.parse() {
-                    (start, end.shift('b'), Token::ByteLiteral(val))
-                } else {
-                    return self.error(start, NonParseableInt);
+                match self.lookahead {
+                    Some((pos, ch)) if is_ident_start(ch) => {
+                        return self.error(pos, UnexpectedChar(ch))
+                    }
+                    _ => {
+                        if let Ok(val) = int.parse() {
+                            (start, end.shift('b'), Token::ByteLiteral(val))
+                        } else {
+                            return self.error(start, NonParseableInt);
+                        }
+                    }
                 }
             }
             Some((start, ch)) if is_ident_start(ch) => return self.error(start, UnexpectedChar(ch)),
@@ -918,6 +925,14 @@ mod test {
                 (r#"   ~~~~    "#, ByteLiteral(255)),
                 (r#"        ~~~"#, ByteLiteral(45)),
             ],
+        );
+    }
+
+    #[test]
+    fn byte_literals_unexpected_char() {
+        assert_eq!(
+            tokenizer(r#"3bs"#).last(),
+            Some(error(loc(2), UnexpectedChar('s')))
         );
     }
 
--- a/parser/src/token.rs
+++ b/parser/src/token.rs
@@ -139,11 +139,11 @@ quick_error! {
         HexLiteralUnderflow {
             description("cannot parse hex literal, underflow")
         }
-        HexLiteralInvalid {
-            description("cannot parse hex literal, invalid digit")
-        }
         HexLiteralWrongPrefix {
             description("wrong hex literal prefix, should start as '0x' or '-0x'")
+        }
+        HexLiteralIncomplete {
+            description("cannot parse hex literal, incomplete")
         }
     }
 }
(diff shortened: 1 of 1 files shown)


parser/src/token.rs
```diff
@@ -752,6 +825,56 @@ mod test {
...
+    fn hex_literals() {
+        test(
+            r#"0x1f 0xf 0x123 0x001 -0xA 0x"#,
+            vec![
+                (r#"~~~~                   "#, IntLiteral(31)),
+                (r#"    ~~~                "#, IntLiteral(15)),
+                (r#"       ~~~~~           "#, IntLiteral(291)),
+                (r#"            ~~~~~      "#, IntLiteral(1)),
+                (r#"                 ~~~~  "#, IntLiteral(-10)),
+                (r#"                     ~~"#, IntLiteral(0)),
```
zummenix: Should we forbid this?

Marwes: I'd say yes, it seems weird since it is sort of a prefix followed by nothing. Tried it in rust and it complained abot "a number without any digits aswell" so that is some precedence for rejecting it.

parser/src/token.rs
```diff
@@ -516,6 +548,47 @@ impl<'input> Iterator for Tokenizer<'input> {
     }
 }
 
+/// Converts partial hex literal (i.e. part after `0x` or `-0x`) to 64 bit signed integer.
+///
+/// This is basically a copy and adaptation of `std::num::from_str_radix`.
+fn i64_from_hex(hex: &str, is_positive: bool) -> Result<i64, Error> {
+    let radix = 16;
```
Marwes: Nit: Could be a `const`

Marwes: Nit: Could be a const

parser/src/token.rs
```diff
@@ -516,6 +548,47 @@ impl<'input> Iterator for Tokenizer<'input> {
...
+    let digits = hex.as_bytes();
+    let mut result = 0i64;
+    if is_positive {
+        for &c in digits {
+            let x = match (c as char).to_digit(radix) {
```
Marwes: The string is already validated to be all hex digits at this point so just do `expect` here which also makes it possible to remove the `HexLiteralInvalid` case.

parser/src/token.rs
```diff
@@ -516,6 +548,47 @@ impl<'input> Iterator for Tokenizer<'input> {
...
+            };
+            result = match result.checked_mul(radix as i64) {
+                Some(result) => result,
+                None => return Err(HexLiteralUnderflow),
+            };
+            result = match result.checked_sub(x as i64) {
+                Some(result) => result,
+                None => return Err(HexLiteralUnderflow),
+            };
+        }
```
Marwes: I assume you split the positive and negative case to deal with abs(i64::MIN) != abs(i64::MAX) but I think it should also work if `x` is multiplied by the sign before each `checked_add`. What do you think? It should remove the duplication I think.

parser/src/token.rs
```diff
@@ -516,6 +548,47 @@ impl<'input> Iterator for Tokenizer<'input> {
...
+                None => return Err(HexLiteralOverflow),
+            };
+        }
+    } else {
+        for &c in digits {
+            let x = match (c as char).to_digit(radix) {
+                Some(x) => x,
+                None => return Err(HexLiteralInvalid),
+            };
+            result = match result.checked_mul(radix as i64) {
```
Marwes: nit: Could write the option matches as `result.checked_mul(radix as i64).ok_or(HexLiteralInvalid)?`

parser/src/token.rs
```diff
@@ -412,6 +428,22 @@ impl<'input> Tokenizer<'input> {
                 let (end, float) = self.take_while(start, is_digit);
                 (start, end, Token::FloatLiteral(float.parse().unwrap()))
             }
+            Some((end, 'x')) => {
+                self.bump(); // Skip 'x'
+                let (end, hex) = self.take_while(end.shift('x'), is_hex);
```
Marwes: This needs a check that the next character is not an identifier otherwise `0x1q` will be tokenized as `0x1 q` . This is done later in the function for normal (decimal) integers but is actually also bugged for floats! That is, `1.0q` is accepted today. If you want it would be great if you can add the same check for floats as well  😅

parser/src/token.rs
```diff
@@ -412,6 +428,22 @@ impl<'input> Tokenizer<'input> {
...
+                match int {
+                    "0" | "-0" => {
+                        let is_positive = int == "0";
+                        match i64_from_hex(hex, is_positive) {
+                            Ok(val) => (start, end, Token::IntLiteral(val)),
+                            Err(err) => return self.error(start, err),
+                        }
+                    }
+                    _ => {
+                        return  self.error(start, HexLiteralWrongPrefix);
```
Marwes: super nit: Extra whitespace after return (just run it through rustfmt if you have it).

zummenix:
I think I addressed all issues.
Also fixed float and byte literals with unexpected char.

P.S: I noticed that checking of the location in tests doesn't work. For example in:
```rust
assert_eq!(
    tokenizer(r#"0x1q"#).last(),
    Some(error(loc(3), UnexpectedChar('q')))
);
```
it doesn't matter if `loc(3)` or `loc(0)` test succeeds anyway. I'm not sure this is intentional.

Marwes:
👍 

> it doesn't matter if loc(3) or loc(0) test succeeds anyway. I'm not sure this is intentional.

That is actually intentional. There are a bunch of [parser tests](https://github.com/gluon-lang/gluon/blob/22863d39034e8ffd56e929e7a0258f544f3fe1bd/parser/tests/basic.rs#L711-L745) which compares the AST produced by the parser to one that is manually created and I didn't want to spec out the spans for each individual expression in those (since it is relevant for those tests and its quite difficult to mess up the spans so it is not worthwhile to test).

In hindsight though it would be better to traverse the AST generated by the parser and zero out all spans before doing those comparisons though. Then spans could matter in `PartialEq for Span`.

Marwes:
Thanks again for the PR!

PR #398, opened 2017-11-22
fix(vm): Don't let rustc think the Array type can't exist

### LINKED ISSUE #396: Build failing on nightly
Its seems to me that building on nightly (for Mac) is failing -

```
$ cargo run -p gluon_repl -- -i
   Compiling gluon_repl v0.6.2 (file:///Users/pradipcaulagi/src/gluon/repl)
    Finished dev [unoptimized + debuginfo] target(s) in 6.88 secs
     Running `target/debug/gluon -i`
[1]    72669 segmentation fault  cargo run -p gluon_repl -- -i

$ cargo test --features test
   Compiling gluon_check v0.6.2 (file:///Users/pradipcaulagi/src/gluon/check)
   Compiling gluon_vm v0.6.2 (file:///Users/pradipcaulagi/src/gluon/vm)
   Compiling gluon v0.6.2 (file:///Users/pradipcaulagi/src/gluon)
    Finished dev [unoptimized + debuginfo] target(s) in 74.44 secs
     Running target/debug/deps/gluon-8ca2f57759c1f7df

running 1 test
error: An unknown error occurred

To learn more, run the command again with --verbose.

$ rustup show
Default host: x86_64-apple-darwin

installed toolchains
--------------------

stable-x86_64-apple-darwin
beta-x86_64-apple-darwin
nightly-x86_64-apple-darwin (default)

active toolchain
----------------

nightly-x86_64-apple-darwin (default)
rustc 1.23.0-nightly (33374fa9d 2017-11-20)
```

Marwes:
The use of `Void` in the `Array` type was meant to absolutely prevent the `Array` type from existing except behind a pointer which is invalid (except the empty array I guess) as the elements are stored inline behind the length field.

Unfortunately https://github.com/rust-lang/rust/pull/45225 made rustc think that fields of the `Array` type could not exist at all and placed fields of `Array` at the start of any struct containing them, causing any write to the `Array` to overwrite other fields.

Fixes #396

### DIFF
--- a/vm/src/array.rs
+++ b/vm/src/array.rs
@@ -7,15 +7,12 @@ use std::slice;
 
 use gc::{Gc, Traverseable};
 
-enum Void {}
-
 /// Abstract array type which have their length store inline with the elements.
 /// Fills the same role as Box<[T]> but takes only 8 bytes on the stack instead of 16
+#[repr(C)]
 pub struct Array<T: Copy> {
     len: usize,
     array_start: [T; 0],
-    /// Prevents the array from being instantiated directly
-    _void: Void,
 }
 
 impl<T: Copy> AsRef<[T]> for Array<T> {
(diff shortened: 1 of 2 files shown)


Marwes:
Needs https://github.com/laumann/compiletest-rs/pull/87

eddyb:
Since having a valid value of type `Array` depends on a valid value of type `Void`, which cannot possibly exist, `Array` also cannot possibly exist, and it's UB to craft one out of thin air.

vm/src/array.rs
```diff
@@ -7,12 +7,17 @@ use std::slice;
...
+    pub struct CantConstruct(());
+}
+
 /// Abstract array type which have their length store inline with the elements.
 /// Fills the same role as Box<[T]> but takes only 8 bytes on the stack instead of 16
 #[repr(C)]
 pub struct Array<T: Copy> {
     len: usize,
     array_start: [T; 0],
+    _marker: self::internal::CantConstruct,
```
brendanzab: I thought just having public fields meant that folks couldn't construct it 🤔

eddyb: I think the intention is that it shouldn't be constructed even accidentally from *this module*.

PR #399, opened 2017-11-25
Add PartialEq for spans

### LINKED ISSUE #394: Compare the span part in PartialEq for Spanned
From https://github.com/gluon-lang/gluon/pull/393#issuecomment-345504474 

>> it doesn't matter if loc(3) or loc(0) test succeeds anyway. I'm not sure this is intentional.

>That is actually intentional. There are a bunch of parser tests which compares the AST produced by the parser to one that is manually created and I didn't want to spec out the spans for each individual expression in those (since it is relevant for those tests and its quite difficult to mess up the spans so it is not worthwhile to test).

>In hindsight though it would be better to traverse the AST generated by the parser and zero out all spans before doing those comparisons though. Then spans could matter in PartialEq for Span.

Probably a bit of tedious thing to do but should be straightforward.

1. Update [PartialEq](https://github.com/gluon-lang/gluon/blob/22863d39034e8ffd56e929e7a0258f544f3fe1bd/base/src/pos.rs#L261) to compare span

2. Add a `MutVisitor` which overwrites the spans in an entire AST with a dummy span

3. Use the visitor to make sure that the spans are the same as the ones in the manually created ASTs here https://github.com/gluon-lang/gluon/blob/22863d39034e8ffd56e929e7a0258f544f3fe1bd/parser/tests/basic.rs#L711-L745 (may be a few other places as well).

4. (optional) There may be some places where tests fail now where the span was wrong and it should be updated to be correct, rather than using the Visitor to insert a dummy span.

zummenix:
- add additional visitor methods in MutVisitor
- clean span values for testing purpuses
- fix all tests

Closes #394 

P.S: Sorry for the giant commit, I was working on this during the week and completely messed up git history.

### DIFF
--- a/parser/src/token.rs
+++ b/parser/src/token.rs
@@ -430,32 +430,28 @@ impl<'input> Tokenizer<'input> {
                     Some((_, ch)) if is_ident_start(ch) => {
                         return self.error(end, UnexpectedChar(ch))
                     }
-                    _ => {
-                        (start, end, Token::FloatLiteral(float.parse().unwrap()))
-                    }
+                    _ => (start, end, Token::FloatLiteral(float.parse().unwrap())),
                 }
             }
             Some((end, 'x')) => {
                 self.bump(); // Skip 'x'
                 let (end, hex) = self.take_while(end.shift('x'), is_hex);
                 match int {
-                    "0" | "-0" => {
-                        match self.lookahead {
-                            Some((_, ch)) if is_ident_start(ch) => {
-                                return self.error(end, UnexpectedChar(ch))
+                    "0" | "-0" => match self.lookahead {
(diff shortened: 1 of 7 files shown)


parser/src/token.rs
```diff
@@ -999,6 +999,7 @@ mod test {
     }
 
     #[test]
+    #[ignore]
```
zummenix: What is the best way to fix this test?

Marwes: Not sure atm, I will look into it

Marwes: If you cherry-pick https://github.com/Marwes/gluon/commit/fafe7b342486a23f81fcbab9568dd2e489e0be63 it should work.

Marwes:
Awesome! Thanks for the PR!

Issue #303, opened 2017-06-21
Try to nanbox gluon's Value type


Marwes:
The [Value](https://github.com/gluon-lang/gluon/blob/59aeab59a259c78beb4ffeac2e6c6bb9048f11fa/vm/src/value.rs#L237-L250) enum is gluon's most pervasive type and it currently takes up 16 bytes on 64-bit platforms. This is rather large since half of it are used for the tag and alignment. It would be interesting to try and use nanboxing to pack the type into 8 bytes instead which could be done with a [nanboxing library](https://github.com/Marwes/nanbox) I have experimented with.

This may be a pretty interesting issue for anyone looking to contribute as it could have a large impact on gluon (its performance especially, halving the memory used for `Value` ought to reduce cache misses substantially!) but some caveats apply.

* The Nanbox implementation currently only support 8 variants.
I may be able to increase this in the nanbox library. Otherwise the `Function`, `Closure` and `PartialApplication` variants can be merged together, the `Thread` variant can use `UserData` and `String` can be merged into `Array` (with the utf-8 validity being encoded in `ValueArray`'s `Repr` instead), which brings it down to 8 variants.

* Nanboxing reduces the size of ints to 32-bit instead of 64-bit.

Marwes:
> Values that don't need all 48 bits (Int, Byte, VmTag) can be crammed into one tag. That is, they might have a three bit tag of 000 and then use the next byte to distinguish between each other. If this is done, the second tag should probably start at 1 to ensure that it can't be used as a float on accident.

Not to keen on doing a two-level check as that implies overhead but I think with a schema like that it would still be possible to have a single branch when checking `is_int` (INT_MASK & value == INT_MASK). It is only if one wants to get the tag to dispatch onto multiple types two branches are needed

```
if extended_tag(value) {
    // 32-bit integers, bytes etc
} else {
    // Pointer, float etc
}
```

> I think that the sign bit can also be used for the tag, which allows for 16 variants. I did some tests and got a fpe whether the sign bit was 1 or 0.

It was a while since I delved into this but I seem to remeber something about NaNs generated by modern CPUs are either always positive or negative but the IEEE 754 spec specifies that NaNs could be either of them and the sign bit should be ignored. It might be possible though, I may have just been to much on the safe side and it could actually be used in the real world (as a part of the theoretical)

> I took a quick look at your nanbox library. Maybe I missed it but are you handling pointers correctly?

[...]

PR #400, opened 2017-11-26
fix(repl): Print out Char as the "character" in thre repl

### LINKED ISSUE #395: Print out Char as the "character" instead of code point integer in the repl
```f#
> 'a'
97

// Should be

> 'a'
'a'
```

oooooba:
Fixes #395

### DIFF
--- a/vm/src/value.rs
+++ b/vm/src/value.rs
@@ -439,7 +439,23 @@ impl<'a, 't> InternalPrinter<'a, 't> {
             Value::Userdata(ref data) => arena.text(format!("{:?}", data)),
             Value::Thread(thread) => arena.text(format!("{:?}", thread)),
             Value::Byte(b) => arena.text(format!("{}", b)),
-            Value::Int(i) => arena.text(format!("{}", i)),
+            Value::Int(i) => {
+                use base::types::BuiltinType;
+                match **self.typ {
+                    Type::Builtin(BuiltinType::Int) => arena.text(format!("{}", i)),
+                    Type::Builtin(BuiltinType::Char) =>
+                        if i < ::std::u32::MAX as isize {
+                            match ::std::char::from_u32(i as u32) {
+                                Some('"') => arena.text(format!("'{}'", '"')),
+                                Some(c) => arena.text(format!("'{}'", c.escape_default())),
+                                None => unreachable!(),
+                            }
+                        } else {
+                            unreachable!()
+                        },
+                    _ => unreachable!(),
+                }
+            },
             Value::Float(f) => arena.text(format!("{}", f)),
         }
     }
--- a/vm/src/value.rs
+++ b/vm/src/value.rs
@@ -444,7 +444,7 @@ impl<'a, 't> InternalPrinter<'a, 't> {
                 match **self.typ {
                     Type::Builtin(BuiltinType::Int) => arena.text(format!("{}", i)),
                     Type::Builtin(BuiltinType::Char) =>
-                        if i < ::std::u32::MAX as isize {
+                        if 0 <= i && i <= ::std::u32::MAX as isize {
                             match ::std::char::from_u32(i as u32) {
                                 Some('"') => arena.text(format!("'{}'", '"')),
                                 Some(c) => arena.text(format!("'{}'", c.escape_default())),
--- a/vm/src/value.rs
+++ b/vm/src/value.rs
@@ -444,16 +444,15 @@ impl<'a, 't> InternalPrinter<'a, 't> {
                 match **self.typ {
                     Type::Builtin(BuiltinType::Int) => arena.text(format!("{}", i)),
                     Type::Builtin(BuiltinType::Char) =>
-                        if 0 <= i && i <= ::std::u32::MAX as isize {
-                            match ::std::char::from_u32(i as u32) {
-                                Some('"') => arena.text(format!("'{}'", '"')),
-                                Some(c) => arena.text(format!("'{}'", c.escape_default())),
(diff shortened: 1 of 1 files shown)


vm/src/value.rs
```diff
@@ -439,7 +439,23 @@ impl<'a, 't> InternalPrinter<'a, 't> {
...
+            Value::Int(i) => {
+                use base::types::BuiltinType;
+                match **self.typ {
+                    Type::Builtin(BuiltinType::Int) => arena.text(format!("{}", i)),
+                    Type::Builtin(BuiltinType::Char) =>
+                        if i < ::std::u32::MAX as isize {
+                            match ::std::char::from_u32(i as u32) {
+                                Some('"') => arena.text(format!("'{}'", '"')),
+                                Some(c) => arena.text(format!("'{}'", c.escape_default())),
+                                None => unreachable!(),
```
brendanzab: Why the special case for `"`?

Also kind of concerning that we are storing unicode chars as `isize`s. What if they are larger than 32 bits? Not really an issue for this PR though 🤔

vm/src/value.rs
```diff
@@ -439,7 +439,23 @@ impl<'a, 't> InternalPrinter<'a, 't> {
...
+                        if 0 <= i && i <= ::std::u32::MAX as isize {
+                            match ::std::char::from_u32(i as u32) {
+                                Some('"') => arena.text(format!("'{}'", '"')),
+                                Some(c) => arena.text(format!("'{}'", c.escape_default())),
+                                None => unreachable!(),
+                            }
+                        } else {
+                            unreachable!()
+                        },
+                    _ => unreachable!(),
```
Marwes: Rather than using `unreachable!` it would be better to just print out the integer as-is. That should be a sensible default regardless of what the type actually is and reduces the number of potential panics.

vm/src/value.rs
```diff
@@ -439,7 +439,23 @@ impl<'a, 't> InternalPrinter<'a, 't> {
             Value::Userdata(ref data) => arena.text(format!("{:?}", data)),
             Value::Thread(thread) => arena.text(format!("{:?}", thread)),
             Value::Byte(b) => arena.text(format!("{}", b)),
-            Value::Int(i) => arena.text(format!("{}", i)),
+            Value::Int(i) => {
+                use base::types::BuiltinType;
+                match **self.typ {
+                    Type::Builtin(BuiltinType::Int) => arena.text(format!("{}", i)),
+                    Type::Builtin(BuiltinType::Char) =>
+                        if 0 <= i && i <= ::std::u32::MAX as isize {
```
Marwes: I think it should be fine to just omit this comparison. When `isize` is 32-bit I think `i as u32` should should preserve all the bits (that is, `i as i32 as u32 == i`).

vm/src/value.rs
```diff
@@ -439,7 +439,23 @@ impl<'a, 't> InternalPrinter<'a, 't> {
...
+                            match ::std::char::from_u32(i as u32) {
+                                Some('"') => arena.text(format!("'{}'", '"')),
+                                Some(c) => arena.text(format!("'{}'", c.escape_default())),
+                                None => unreachable!(),
```
Marwes: Could you change this to  something akin to `ìce!("Invalid character passed to pretty printing")` (`ice!` reports in internal compiler error with a link to gluon's issue list).

(Using `unreachable!` is fine but I think it is best to limit it to places where it is possible to locally verify that the branch is `unreachable`. Here the character can be created far away from this branch)

Marwes:
@brendanzab
> Also kind of concerning that we are storing unicode chars as isizes. What if they are larger than 32 bits? Not really an issue for this PR though 🤔

I'd view `Value`'s variants more like different "representations" of data instead of them having knowledge about the underlying type. Though it would be easy to add `Value::Char` I have avoided that since 1, I am lazy, 2, if we implement nan-boxing (#303) of values then we will be limited to 8 (I think) variants in which case we can't "waste" variants on primitive non-boxed types such as `Char`.

brendanzab:
More the issue is that it makes characters system-dependant, and reduces the number of valid unicode characters we can store.

vm/src/value.rs
brendanzab: Is it really an internal compiler error though? This is more of a runtime problem, no?

Marwes: Well the "compiler" (speaking about it in a broader sense) has made a mistake causing an invalid character to be stored. Is there any reason to separate "Internal compiler errors" from "Internal virtual machine errors"?

Marwes:
> More the issue is that it makes characters system-dependant, and reduces the number of valid unicode characters we can store.

Only if isize is 16-bit though, since the integer will just be treated as a set of 32 bits rather than a signed integer. I opened https://github.com/gluon-lang/gluon/issues/403 though since `VmInt` should be fixed size.

vm/src/value.rs
oooooba: When `isize` is 64-bit, how can I deal with it?
Should I forcibly cast i to u32, then call `char::from_u32` (that is, `char::from_u32(i as u32)`)?

Marwes: Yep. Just truncating with `as` is fine.

vm/src/value.rs
oooooba: `char::escape_default` returns `'\"'` for input `'"'`.
https://doc.rust-lang.org/std/primitive.char.html#method.escape_default
I thought that input `'"'` to repl should corresponds to output `'"',` and treated it specially.

Marwes:
👍 Thanks!

PR #401, opened 2017-11-26
chore: Write lalrpop files to OUT_DIR instead of the source tree

Marwes:
Should let gluon be ran under cargobomb https://github.com/rust-lang-nursery/crater/issues/157 but I think that regardless of that it is better to generate files outside of the source tree

### DIFF
--- a/parser/build.rs
+++ b/parser/build.rs
@@ -3,6 +3,9 @@ extern crate lalrpop;
 fn main() {
     ::std::env::set_var("LALRPOP_LANE_TABLE", "disabled");
 
-    lalrpop::Configuration::new().process_current_dir().unwrap();
-    println!("cargo:rerun-if-changed=src/grammar.lalrpop");
+    lalrpop::Configuration::new()
+        .use_cargo_dir_conventions()
+        .process_file("src/grammar.lalrpop")
+        .unwrap();
+    println!("cargo:rerun-if-changed=grammar.lalrpop");
 }
--- a/vm/build.rs
+++ b/vm/build.rs
@@ -1,10 +1,14 @@
-
 #[cfg(feature = "test")]
 mod build {
     extern crate lalrpop;
+
     pub fn main() {
-        lalrpop::Configuration::new().process_current_dir().unwrap();
-        println!("cargo:rerun-if-changed=src/core/grammar.lalrpop");
+        lalrpop::Configuration::new()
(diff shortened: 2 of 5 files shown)


vm/build.rs
```diff
@@ -1,10 +1,18 @@
...
-        lalrpop::Configuration::new().process_current_dir().unwrap();
-        println!("cargo:rerun-if-changed=src/core/grammar.lalrpop");
+        let out_dir = ::std::env::var("OUT_DIR").unwrap();
+        let manifest_dir = ::std::env::var("CARGO_MANIFEST_DIR").unwrap();
+
+        lalrpop::Configuration::new()
+            .set_out_dir(&out_dir[..])
+            .set_in_dir(&manifest_dir[..])
+            .process_file("src/core/grammar.lalrpop")
+            .unwrap();
```
brendanzab: There is a [`use_cargo_dir_conventions`](https://github.com/nikomatsakis/lalrpop/blob/master/lalrpop/src/api/mod.rs#L64-L70) option... would this simplify things?

Marwes: Ah, that is a bit better. Was struggling to get the directories right so I just coded them all explicitly to make sure they were right.

parser/build.rs
```diff
@@ -3,6 +3,10 @@ extern crate lalrpop;
...
     ::std::env::set_var("LALRPOP_LANE_TABLE", "disabled");
 
-    lalrpop::Configuration::new().process_current_dir().unwrap();
-    println!("cargo:rerun-if-changed=src/grammar.lalrpop");
+    let out_dir = ::std::env::var("OUT_DIR").unwrap();
+    lalrpop::Configuration::new()
+        .use_cargo_dir_conventions()
+        .process_file("src/grammar.lalrpop")
+        .unwrap();
+    println!("cargo:rerun-if-changed={}/grammar.rs", out_dir);
```
brendanzab: Why not:

```rust
println!("cargo:rerun-if-changed=src/grammar.lalrpop");
```
?

Marwes: That is what it was before... Must have copy pasted and accidentally changed it. (The way it is now won't even work, as the .rs ONLY changes if the build script is run 🤦‍♂️

brendanzab: That's what I was thinking... 😛

PR #402, opened 2017-11-26
Run script tests in parallel

Marwes:
This uses a small test library I hacked together https://github.com/Marwes/tensile to allow tests defined with .glu files to run in parallel and fail independently of eachother (currently when one script fails it fails the entire test and skips running any other tests). (It is inspired by https://github.com/haskell/test-framework)

It also fixes a number of bugs with parallel gluon programs I found when starting to run the tests in parallel.

Ultimately I want to further integrate this so that we can use gluon itself to define tests that `tensile` understands (ie, printing out individual pass/fail messages and possibly running them in parallel).


scripts/travis.sh
```diff
@@ -10,6 +10,7 @@
   cargo test --features test --package gluon_vm &&
   cargo test --features test --package gluon_format &&
   cargo test --features test --package gluon &&
+  cargo test --features test --package gluon_test &&
```
Marwes: I want to highlight that I added a separate crate for the gluon script tests here https://github.com/gluon-lang/gluon/pull/402/commits/3b7073f2fb5e4f9130f49b702200de08743c2b6c . Still undecided which way is the right approach here...

Marwes:
Moved all the tests back into the main crate as that seems easier. To combat the drawback of having to declare each of the tests in Cargo.toml I added a step to the build script to make sure that all tests are declared there or else we panic and stop the build

PR #412, opened 2017-12-12
Release version 0.7

Marwes:
Cleanup of warts I have encountered while trying to write a better example for the readme. #411 was a rather large diversion but I should be back on track now.


std/parser.glu
```diff
@@ -88,6 +88,13 @@ let uncons stream : OffsetString -> Option { char : Char, rest : OffsetString }
 let update_position c position : Char -> Position -> Position =
     position + char.len_utf8 c
 
+let (<?>) p msg : Parser a -> String -> Parser a =
+    parser (\stream ->
+        match p stream with
+        | Ok x -> Ok x
+        | Err _ -> Err msg)
```
brendanzab:
```
    parser <| \stream ->
        match p stream with
        | Ok x -> Ok x
        | Err _ -> Err msg
```
perhaps? (nitpicking)

brendanzab:
Btw, it might be worth looking at porting https://github.com/elm-tools/parser/ at some stage - seems to be a nicer approach than traditional parser combinators.

Marwes:
> Btw, it might be worth looking at porting https://github.com/elm-tools/parser/ at some stage - seems to be a nicer approach than traditional parser combinators.

It has some good ideas to steal but also some issues so probably should port it straight up (remember looking at it a while back but can't remember where I wrote this down, `delayedCommit` is encourages inneficient code is one thing I believe).

It may also be worth rewriting the parser into CPS now that universal quantification is implemented as that is how both parsec and attoparsec are implemented.

base/src/ast.rs
```diff
@@ -708,94 +708,101 @@ pub fn walk_pattern<'a, V: ?Sized + Visitor<'a>>(v: &mut V, p: &'a Pattern<V::Id
 pub trait Typed {
     type Ident;
 
-    fn env_type_of(&self, env: &TypeEnv) -> ArcType<Self::Ident>;
+    fn env_type_of(&self, env: &TypeEnv) -> ArcType<Self::Ident> {
+        self.try_type_of(env).unwrap()
+    }
```
brendanzab: Would be interesting to remove this method, then bubble up the unwraps explicitly - see what is relying on it!

Marwes: I think it may be worth keeping actually. The original assumption with this method is that it is called on something that knows what type it has and in that case it should be infallible (this is the case of parser AST -> core AST translation, only expressions that has been typechecked without error should be passed to translation). However, as the language server now needs to query the type of expression, patterns etc even if the typechecker finds an error it needs a fallible variant of this. I do need to document this assumption though.

std/parser.glu
Marwes: Filed under https://github.com/gluon-lang/gluon/issues/414

Marwes:
This should be 0.7 🎉 . I had hoped to get this out with bit more of a bang but since I am not done with the post I was writing for this and we now enter the holidays I figure it will be better to wait with that until after new years instead (with a probable 0.7.1) and just publish this as is. Just one last nit on the language server then I am done.

PR #415, opened 2017-12-17
Change Getable::from_value to return a value instead of Option

### LINKED ISSUE #229: Panic on an invalid call to Getable::from_value instead of returning Option
`Getable::from_value` is used to marshal values from gluon into Rust and it currently returns an `Option` since obviously a `Value` may not contain the correct type. However in some cases such as https://github.com/gluon-lang/gluon/blob/9adf8be8065ebeb385c94829ad2368c307c95d6e/vm/src/api.rs#L333-L336 it is not possible to know if the value is of the correct type (or at least not efficiently).

Since it is already the case that a return value of `Some` is no guarantee that the value is well typed it may be better to leave it up to the caller to call it on a valid valid and simply panic if it is an obvious error (where one would return `None` now).

zummenix:
All tests pass. Let me know if I need to add additional checks.

Closes #229

### DIFF
--- a/c-api/src/lib.rs
+++ b/c-api/src/lib.rs
@@ -188,10 +188,9 @@ pub unsafe extern "C" fn glu_get_string(
 ) -> Error {
     let mut context = vm.context();
     let stack = context.stack.current_frame();
-    match stack
-        .get_variants(index)
-        .and_then(|value| <&str>::from_value(vm, value))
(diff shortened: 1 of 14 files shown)


Marwes:
Looks good, nice to get rid of some unwraps and get some better panic messages!

PR #417, opened 2017-12-20
Allow pattern matching on literals

### LINKED ISSUE #358: Allow matching on strings and integers
https://gitter.im/gluon-lang/gluon?at=59dec1e2210ac26920d5e686

```f#
match "test" with
| "a" -> 1
| "test" -> 2
| _ -> 3


match 1 with
| 1 -> 11
| 2 -> 22
| _ -> 33
```

zummenix:
This will close #358

### DIFF
--- a/tests/pass/match_literal.glu
+++ b/tests/pass/match_literal.glu
@@ -34,6 +34,23 @@ let r1 input =
     | { x = 3 } -> 12
     | _ -> 13
 
+type R2 = { x: Int, y: Int }
+
+let r2 input =
+    match input with
+    | { x = 2, y = 3 } -> 14
+    | { x = 3, y = 4 } -> 15
+    | _ -> 16
+
+type TestM = | TestM Int String
+
+let test_m input =
+    match input with
+    | TestM 1 "hello" -> 20
+    | TestM _ "world" -> 21
+    | TestM 2 _ -> 22
+    | _ -> 23
+
 let match_ns =
     assert_ieq (ns (A 1)) 7
         *> assert_ieq (ns (A 2)) 8
@@ -56,7 +73,18 @@ let match_r1 =
         *> assert_ieq (r1 { x = 3 }) 12
         *> assert_ieq (r1 { x = 4 }) 13
 
+let march_r2 =
+    assert_ieq (r2 { x = 2, y = 3 }) 14
+        *> assert_ieq (r2 { x = 3, y = 4 }) 15
+        *> assert_ieq (r2 { x = 4, y = 4 }) 16
+
+let match_test_m =
+    assert_ieq (test_m (TestM 1 "hello")) 20
+        *> assert_ieq (test_m (TestM 2 "world")) 21
+        *> assert_ieq (test_m (TestM 2 "hello")) 22
+        *> assert_ieq (test_m (TestM 3 "")) 23
+
 let tests =
-    match_ints *> match_strings *> match_ns *> match_r1
+    match_ints *> match_strings *> match_ns *> match_r1 *> march_r2 *> match_test_m
 
 run tests
--- a/tests/pass/match_literal.glu
+++ b/tests/pass/match_literal.glu
@@ -73,7 +73,7 @@ let match_r1 =
         *> assert_ieq (r1 { x = 3 }) 12
         *> assert_ieq (r1 { x = 4 }) 13
 
-let march_r2 =
+let match_r2 =
     assert_ieq (r2 { x = 2, y = 3 }) 14
         *> assert_ieq (r2 { x = 3, y = 4 }) 15
         *> assert_ieq (r2 { x = 4, y = 4 }) 16
@@ -85,6 +85,6 @@ let match_test_m =
         *> assert_ieq (test_m (TestM 3 "")) 23
 
 let tests =
-    match_ints *> match_strings *> match_ns *> match_r1 *> march_r2 *> match_test_m
+    match_ints *> match_strings *> match_ns *> match_r1 *> match_r2 *> match_test_m
 
 run tests
--- a/tests/pass/match_literal.glu
+++ b/tests/pass/match_literal.glu
@@ -26,6 +26,14 @@ let ns input =
     | B -> 9
     | _ -> 10
 
+type R1 = { x: Int }
+
+let r1 input =
+    match input with
+    | { x = 2 } -> 11
+    | { x = 3 } -> 12
+    | _ -> 13
+
 let match_ns =
     assert_ieq (ns (A 1)) 7
         *> assert_ieq (ns (A 2)) 8
@@ -43,7 +51,12 @@ let match_strings =
         *> assert_ieq (strings "B") 5
         *> assert_ieq (strings "") 6
 
+let match_r1 =
+    assert_ieq (r1 { x = 2 }) 11
+        *> assert_ieq (r1 { x = 3 }) 12
+        *> assert_ieq (r1 { x = 4 }) 13
+
 let tests =
-    match_ints *> match_strings *> match_ns
+    match_ints *> match_strings *> match_ns *> match_r1
 
 run tests
--- a/tests/pass/match_literal.glu
+++ b/tests/pass/match_literal.glu
@@ -1,0 +1,49 @@
+let prelude = import! std.prelude
+let { run, writer, assert_eq } = import! std.test
+let int = import! std.int
+let { (*>) } = prelude.make_Applicative writer.applicative
+
+let assert_ieq = assert_eq int.show int.eq
+
+let ints input =
+    match input with
+    | 0 -> 1
+    | 1 -> 2
+    | _ -> 3
+
+let strings input =
+    match input with
+    | "A" -> 4
+    | "B" -> 5
+    | _ -> 6
+
+type N = | A Int | B
+
+let ns input =
+    match input with
+    | A 1 -> 7
+    | A 2 -> 8
+    | B -> 9
+    | _ -> 10
+
+let match_ns =
+    assert_ieq (ns (A 1)) 7
+        *> assert_ieq (ns (A 2)) 8
+        *> assert_ieq (ns B) 9
+        *> assert_ieq (ns (A 3)) 10
+
+let match_ints =
+    assert_ieq (ints 0) 1
+        *> assert_ieq (ints 1) 2
+        *> assert_ieq (ints 2) 3
+        *> assert_ieq (ints 3) 3
+
+let match_strings =
+    assert_ieq (strings "A") 4
+        *> assert_ieq (strings "B") 5
+        *> assert_ieq (strings "") 6
+
+let tests =
+    match_ints *> match_strings *> match_ns
+
+run tests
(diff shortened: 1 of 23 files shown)


Marwes:
@zummenix  I think you are on the right track by returning `CType::Variable` for literals since they share the common property that they do not have any nested patterns.

What you want to do is tweak the construction of the match expression in `compile_variable`
https://github.com/gluon-lang/gluon/blob/483dfef2cc1e2ff58383713d26430bd39245c979/vm/src/core/mod.rs#L1260-L1267

so that instead of just creating a `match` on the first alternative/pattern you also create an `Alternative` for each of the literal patterns (also add a `Literal` variant to `core::Pattern`.

```
let patterns: Vec<core::Pattern> = equations
                .iter()
.map(|equation| match *equation.patterns.first().unwrap() {
    Pattern::Literal => core::Literal,
    Pattern::Ident => core::Ident,
    _ => unreachable!()
})
```

Lastly you need to add to compiler.rs and generate similar code to how constructors are handled https://github.com/gluon-lang/gluon/blob/483dfef2cc1e2ff58383713d26430bd39245c979/vm/src/compiler.rs#L734-L752 . But instead of using `TestTag` you will need to generate equality comparisons https://github.com/gluon-lang/gluon/blob/483dfef2cc1e2ff58383713d26430bd39245c979/vm/src/compiler.rs#L842 or generate a function call to the string comparison primitive. (I recommend you add it as a global just like `@error` https://github.com/gluon-lang/gluon/blob/483dfef2cc1e2ff58383713d26430bd39245c979/vm/src/primitives.rs#L431-L434 then you can call `Compiler::find(self, &Symbol::new("@string_eq"))`

That is about it but let me know if you get stuck.

zummenix:
I'm definitely doing something wrong here.

vm/src/core/mod.rs
```diff
@@ -1267,6 +1276,34 @@ impl<'a, 'e> PatternTranslator<'a, 'e> {
...
+        let patterns: Vec<_> = equations
+            .iter()
+            .map(|equation| match equation.patterns.first().unwrap().value {
+                ast::Pattern::Literal(ref l) => Pattern::Literal(l.clone()),
+                ast::Pattern::Ident(ref id) => Pattern::Ident(id.clone()),
+                _ => unreachable!()
+            }).collect();
+
+        let alternatives = patterns.into_iter()
+            .zip(equations.iter().map(|e| e.result))
```
Marwes: Instead of `e.result`, this should be the same `Expr` calculated by `compile_variable` (take the first binding in that function `let expr` and make that be the right hand side of each `Alternative`).

If you are wondering why, it is because it recurses down into all the patterns we need to match on *after* the current pattern. The algorithm keeps on recursing like that (taking the first column of equations, creating a core pattern and recursing on the rest) and when `equations` is empty, then `e.result` is returned which is the expression that is on the right hand side of the AST alterantive.

Marwes: This is maybe not the problem you have right now though since you will only notice it if you have a test with nested patterns.

vm/src/compiler.rs
```diff
@@ -719,10 +719,21 @@ impl<'a> Compiler<'a> {
...
+                        Pattern::Literal(ref l) => {
+                            let instruction = match *l {
+                                ast::Literal::Byte(_) => ByteEQ,
+                                ast::Literal::Int(_) | ast::Literal::Char(_) => IntEQ,
+                                ast::Literal::Float(_) => FloatEQ,
+                                ast::Literal::String(_) => unimplemented!(),
+                            };
+                            function.emit(instruction);
+                            start_jumps.push(function.function.instructions.len());
+                            function.emit(CJump(0));
```
Marwes: What you are missing here is to push the arguments to the instruction to the top of the stack.

The value that we are inspecting (the value between `match ... with`) should be at the top of the stack here so emit a `Push(function.stack_size() - 1)` instruction for that.

For the literal itself it is easy for bytes,chars and floats, just emit `PushInt/PushFloat`. For strings you can instead call `emit_string` and it will make sure to add the string constant to the environment and emit the `PushString` instruction.

After those values are pushed you can emit the comparison.

```
// scrutinee is what is between match scrutinee with
[..., scrutinee]
Push(stack_size - 1)
[..., scrutinee, scrutinee]
PushInt(123)
[..., scrutinee, scrutinee, 123]
IntEQ
[..., scrutinee, bool]
CJump
[..., scrutinee]
// Note that we couldn't skip the `Push` instruction since wee need scrutinee on the top of the stack after CJump executes since we may have another alternative after that compares against a different value
```

Marwes:
There is not an explicit literal parser but integer constants are defined here https://github.com/gluon-lang/gluon/blob/0909139a2a80389a56d6546e921aef9e52abf9cc/vm/src/core/grammar.lalrpop#L70 . Don't worry about edge cases in the core parser since it is only used in tests. Only define exactly what you need instead.

Marwes:
> I'm definitely doing something wrong here.

:) Let me know if you are still stuck. Gitter can be good for quick discussion back and forth as well, assuming I am online.

zummenix:
Thanks for explanation! Comparisons are working now. What I still don't understand is how to handle nested patterns and blank `_` (default) pattern.

Also please let me know what tests and where you want to see for this feature. For example, this test:
https://github.com/zummenix/gluon/blob/c03a27e2c8736ff5a3c99cd9092425eed0d9cde0/tests/pattern_match.rs#L45-L56
is failing with an error: `Undefined variable @string_eq in <top>` 

> Gitter can be good for quick discussion back and forth as well, assuming I am online.

Yeah, I think we have incompatible timezones :)

vm/src/core/mod.rs
Marwes: So the nested pattern handling is a bit tricky still, I will try and explain as best I can but If you still have trouble I can jump in and finish that part.

Essentially you want to take this first part of `compile_constructor` https://github.com/gluon-lang/gluon/blob/0909139a2a80389a56d6546e921aef9e52abf9cc/vm/src/core/mod.rs#L1120-L1140 to group all identical `Literal` patterns together to make a `Map<Literal, Vec<Equation>>`.

The for each entry in this map (just like for the constructor case) you generate one `Alternative` with the literal as the key and use the `Vec<Equation>` value to generate the `expr` field of `Alternative` https://github.com/gluon-lang/gluon/blob/0909139a2a80389a56d6546e921aef9e52abf9cc/vm/src/core/mod.rs#L1178 . You can do more or less the same thing as in `compile_constructor` but you don't have to add the nested fields (since there are none). https://github.com/gluon-lang/gluon/blob/0909139a2a80389a56d6546e921aef9e52abf9cc/vm/src/core/mod.rs#L1162-L1165 .

Also you only need to handle `complete == false` since there is no way to have a complete match on literals.

zummenix:
> I am a bit slow to respond due to christmas and everything but I hope this can help out on these last few parts.

No worries and thank you for the explanation. I'm also busy and will be able to continue only in the new year.

Marwes:
```
diff --git a/vm/src/compiler.rs b/vm/src/compiler.rs
index 91cd35df..cf1e6fc5 100644
--- a/vm/src/compiler.rs
+++ b/vm/src/compiler.rs
@@ -783,6 +783,8 @@ impl<'a> Compiler<'a> {
                         Pattern::Literal(_) => {
                             function.function.instructions[start_index] =
                                 CJump(function.function.instructions.len() as VmIndex);
+                            // Add a dummy variable to mark where the literal itself is stored
+                            function.new_stack_var(self, self.empty_symbol.clone(), Type::hole());
                         }
                     }
                     self.compile(&alt.expr, function, tail_position)?;
diff --git a/vm/src/core/grammar.lalrpop b/vm/src/core/grammar.lalrpop
index f4084dfb..3fe9f90b 100644
--- a/vm/src/core/grammar.lalrpop
+++ b/vm/src/core/grammar.lalrpop
@@ -52,17 +52,21 @@ Alternative: Alternative<'a> = {
     },
 };
 
+FieldExpr : (Symbol, Option<Expr<'a>>) = {
+    <Identifier> <("=" <Expr>)?>,
+};
+
 AtomicExpr: Expr<'a> = {
     "(" <Expr> ")",
-    "{" <args: Comma<Identifier>> "}" => {
+    "{" <args: Comma<FieldExpr>> "}" => {
         let id = TypedIdent {
             name: symbols.symbol("<record>"),
             typ: Type::record(vec![], args.iter()
-                .map(|arg| Field { name: arg.clone(), typ: Type::hole(), })
+                .map(|&(ref arg, _)| Field { name: arg.clone(), typ: Type::hole(), })
                 .collect()),
         };
         let args = args.into_iter()
-            .map(|id| Expr::Ident(TypedIdent::new(id), Span::default()));
+            .map(|(id, expr)| expr.unwrap_or_else(|| Expr::Ident(TypedIdent::new(id), Span::default())));
         let args = allocator.arena.alloc_extend(args);
         Expr::Data(id, args, BytePos::default(), ExpansionId::default())
     },
diff --git a/vm/src/core/mod.rs b/vm/src/core/mod.rs
index da349622..ab52c598 100644
--- a/vm/src/core/mod.rs
+++ b/vm/src/core/mod.rs
@@ -2132,4 +2132,31 @@ mod tests {
 
         check_translation(expr_str, expected_str);
     }
+
+    #[test]
+    fn match_record_with_literal() {
+        let expr_str = r#"
+            match { x = 2, y = 3 } with
+            | { x = 2, y = 3 } -> "4"
+            | _ -> "6"
+        "#;
+
+        let expected_str = r#"
+            let p = { x = 2, y = 3 } in
+            match p with
+            | { x = x, y = y } ->
+                match x with
+                | 2 ->
+                    match y with
+                    | 3 -> "4"
+                    | _ -> "6"
+                    end
+                | _ -> "6"
+                end
+            end
+        "#;
+
+        check_translation(expr_str, expected_str);
+    }
+
 }

```

That should fix the problem with index out of bounds (the dummy variable is the actual fix but I added a test to check that it did translate correctly)

zummenix:
Thanks!

Marwes:
@zummenix I forgot about this over the last week as I got tangled up with some other projects. Is there anything else you want to add to this? (I can't think of anything missing at least)

zummenix:
@Marwes It's still a WIP. Match on records with two and more fields doesn't work yet, also I'm planning to add more test cases in `tests/pass/match_literal.glu`

Marwes:
Ahh, kk didn't think about those being absent. Those cases shouldn't need any changes to compiler.rs at least so as long as you get the translation in `core/mod.rs` correct I'd expect it to work.

Adding a test matching on a variant value with multiple literals could also be good to add.
```
type Test = | Test Int String
match Test 1 "" ...
```

tests/pass/match_literal.glu
```diff
@@ -56,7 +73,18 @@ let match_r1 =
         *> assert_ieq (r1 { x = 3 }) 12
         *> assert_ieq (r1 { x = 4 }) 13
 
+let march_r2 =
```
brendanzab: Spelling error here :)

Marwes:
Turns out the `stack_size` function were rather misleading... oops https://github.com/gluon-lang/gluon/pull/417/commits/87e4c95e3c66c6f4ca905b62111ccc0aa7fcd9f0 .

zummenix:
I think it's ready for final review. Please let me know if we need anything else here. And thanks for the help!

Marwes:
Merged! Thanks for implementing this @zummenix even though you had to brave what is perhaps the most complex/hackiest (depending on how you want to look at it) part of the compiler.

Issue #355, opened 2017-10-10
Add glob imports


Marwes:
```f#
// Imports `List, monad, ...`
let { monad = { .. } } = import! "std/io.glu"
// or perhaps
let { { * } } = import! "std/list.glu"
```

This would import all fields (both type and value fields) into scope of the given record.

One potential problem is polymorphic records

```f#
let test record =
    // At this point all we can say is that `record` must be a record but we don't know which fields it has
    // so we can't bring any bindings into scope
    let { .. } = record
    // Apparently `x` was a field of `record` but this will still error
    x + record.x
```

May need to make it a hard error to use a glob on a polymorphic record

Marwes:
Turns out I did have an issue for this #2 😆 . This issue is more comprehensive though so closing it in favor for this.

Boscop:
Yes please :)

Maybe `let * = import! "std/list.glu"` or `let _ = import! "std/list.glu"`.
I would prefer a syntax that doesn't require doubly nested braces :)

It would be very useful for artistic live scripting to have glob imports.
(In that scenario, speed of typing (expressing ideas) matters most, and the code can be cleaned up later.)

PR #425, opened 2018-01-01
feat: Add implicit arguments

Marwes:
Closes https://github.com/gluon-lang/gluon/issues/98


parser/src/token.rs
```diff
@@ -538,6 +540,9 @@ impl<'input> Iterator for Tokenizer<'input> {
                 '}' => Some(Ok(pos::spanned2(start, start.shift(ch), Token::RBrace))),
                 ']' => Some(Ok(pos::spanned2(start, start.shift(ch), Token::RBracket))),
                 ')' => Some(Ok(pos::spanned2(start, start.shift(ch), Token::RParen))),
+                '?' => Some(Ok(
+                    pos::spanned2(start, start.shift(ch), Token::QuestionMark),
```
brendanzab: I think `Question` would be ok here, so we keep a consistent identifier length.

check/tests/pass.rs
```diff
@@ -866,3 +869,63 @@ match Test { x = 1 } with
...
+#[test]
+fn function_implicit_arg() {
+    let _ = ::env_logger::init();
+    let text = r#"
+
+let f eq l r: [a -> a -> Bool] -> a -> a -> Bool = eq l r
+let eq_int l r : Int -> Int -> Bool = True
+let eq_string l r : String -> String -> Bool = True
+f 1 2
+f "" ""
```
brendanzab: oooh, so cool!

Marwes: Well it does not work yet :P

Marwes:
Bit further along now, still experimenting with how this will work. The use of `@<key> <value>` as a way to associate key-value metadata to bindings were added which I think should work well in general (though for implicits in particular, maybe we want explicit syntax).

I was able to piggy back on the metadata functionality with this solution though and so managed to get a working `/=` function.

README.md
```diff
@@ -339,3 +339,4 @@ This language takes its primary inspiration from [Lua][Lua], [Haskell][Haskell]
 [Haskell]: http://www.haskell.org
 [OCaml]: http://www.ocaml.org
 [Rust]: http://www.rust-lang.org
+st-lang.org
```
brendanzab: 🤔

Marwes: Ooops, vim gets weird sometimes...

Marwes:
Trying to work through the standard library, replacing any uses of the old overloading system and the `make_Alternative` (or similar `make_` functions) with implicits. Been studying https://github.com/ocamllabs/ocaml-modular-implicits in the hope that the remaining problems would have been fixed in that implementation but unfortunately the implementation appears to have the same problems with inference.

```ocaml
module type Monad = sig
  type 'a t
  val return : 'a -> 'a t
  val bind : 'a t -> ('a -> 'b t) -> 'b t
end;;

implicit module OptionMonad = struct
  type 'a t = 'a option
  let return x = Some x
  let bind x f = match x with
    | None -> None
    | Some x -> f x
end;;

let return {M : Monad} x = M.return x;;
let bind {M : Monad} m k = M.bind m k;;

let abc (f : 'a -> 'a OptionMonad.t) = 1;;
abc return ;;
```

gives 

```
Characters 4-10:
  abc return ;;Error: This expression has type {M : Monad} -> 'a -> 'a M.t
       but an expression was expected of type 'b -> 'b OptionMonad.t
```

But i'd really like this to work or that is going to be a pretty annoying limitation...

README.md
```diff
@@ -86,14 +86,14 @@ let string = import! std.string
 // Since imports in gluon returns regular values we can load specific parts of a module using pattern matches.
 // This match in particular brings in the equality operator for the `Char` type (this is required in
 // the current version of gluon but will be fixed in the future).
-let { eq = { (==) } } = import! std.char
+let char = import! std.char
 
-let { (<>) } = prelude.make_Semigroup string.semigroup
+let { (<>) } = import! std.prelude
```
brendanzab: Why do we also not need to import the implicit instances as well?

Marwes: So currently I just search all reachable bindings marked as implicit and try to match them against the implicit type. This obviously extremely slow and doesn't allow implicit bindings to not be imported. That is, unless you are really careful to not make them reachable you will get a lot of (maybe) unintended implicit binding included. (That is really annoying since being careful when bringing in things in scope with `let int = import! std.int` actually has the opposite effect on implicits as all implicit bindings in the `int` record are made available for implicit resolution, whereas `let { num } = import! std.int` only brings in the `Num` instance).

Open for ideas on how to best represent importing `implicit` instances right now.

brendanzab: Was thinking you could do have an implicit pattern thingy, eg:

```
let { [num] } = import! std.int
```
or:
```
let { num } = import! std.int
let [num] = num
```

brendanzab: Just an idea! Not sure how well it would work with the function parameter syntax we were thinking of:

```
let { [num] } = import! std.int

let double : forall a. [Num a] -> a -> a = 
    \[num] x -> x + x

let double2 = double 2
```

We might also want to reserve `[pattern]` for pattern mathing on lists.... dilemmas!

brendanzab: The disadvantage of the above would be you couldn't do:
```
let { .. } = import! std.int
```
... and have all the implicits come into scope. 🤔

brendanzab:
Did you see these blog posts?

- http://jobjo.github.io//2018/01/28/implicit-functors.html
- http://jobjo.github.io//2018/02/05/lenses-via-modular-implicits.html

README.md
Marwes: > let { [num] } = import! std.int

That is perhaps the most straightforward and explicit solution, but I wonder if it still ends up being to much work. All languages with type classes (that I can think of) tend to not need each instance to be imported. Maybe it isn't that bad though and I guess we could still have a syntax for "bring all implicit bindings into scope"

Marwes: > We might also want to reserve [pattern] for pattern matching on lists.... dilemmas!

Or arrays which currently get the expression syntax for `[expr]`. Still a bit torn on whether lists or arrays deserve this syntax. Lists are much more flexible than arrays with immutability, but on the other hand they should also be used sparingly...

brendanzab: Scala requires you to import implicits. As does Agda, Coq, and Lean. If you have a globally coherent (and hence, non-modular) typeclass system like Haskell, Purescript, or Rust, then you don't need to.

Marwes:
> Did you see these blog posts?

Yep, they (as all other parts of OCamls implicits) manage to avoid the case in https://github.com/gluon-lang/gluon/pull/425#issuecomment-364802670 which proves a bit annoying. I think I know how to fix it but it requires some careful subsumption checking.

README.md
Marwes: Yeah, we do need a way to limit what is considered in implicit resolution. I just meant that individually importing every single binding would be tedious so we need a way to "import every implicit binding in x"

OvermindDL1:
OCaml's Implicit Modules was heavily based on Scala's Implicit Arguments, maybe look at how it sees it as Scala uses function-level HM type inference as well?

Marwes:
@OvermindDL1 Scala has some fairly notorious problems with implicits so I am a bit vary of taking ideas from it. It may still be worth looking into but I haven't been able to find where the type inference is actually done in Scala yet 😅

brendanzab:
You'd probably be better off looking at Dotty for a better look at how Scala does it. But yes, the saying goes, 'nobody understands Scala's implicit resolution algorithm'. Not really something we'd want to emulate too closely.

brendanzab:
This seems to be the file: https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/typer/Implicits.scala

brendanzab:
Basically Scala tries far too hard to resolve ambiguities without programmer intervention, leading to a great deal of confusion and paranoia. A better approach would be to cry for help when they overlap, and require the programmer to explicitly say what implicit they intended to use.

OvermindDL1:
I'm entirely fine with occasionally having to explicitly pass in the witness module, I'd want to do that to override defaults anyway.

Marwes:
This mostly works now 🎉 . To make it possible to limit the implicits being considered I added the `?` syntax in record patterns https://github.com/gluon-lang/gluon/pull/425/files#diff-ff1563c4c968e716093f2061b79541a3R11 . If a record pattern contains `?` then all fields marked as `@implicit` and all fields that hold a type marked as `@implicit` (such as `Functor`) are considered as implicit candidates. I am open to alternatives to the specific syntax here if there are any better alternatives, though I may end up merging this with the current syntax and change it later (gluon_format will make the change trivial unless the change is complicated!). (I considered `..` as well but I think that is better used as a normal glob import https://github.com/gluon-lang/gluon/issues/355).

I also added the `?` prefix in function argument declarations as a way to explicitly refer to an implicit argument. If `?` is not specified then it is an explicit argument and the compiler will implicitly bind any implicit arguments preceding the explicit argument in the type.

Some stuff that still needs to be done is to let `[Eq a] -> Eq (List a)` be considered for the implicit resolution (should be fairly easy actually), take another pass on the standard library to add implicit arguments where appropriate and explain everything in the tutorial.

Handling ambiguities also needs to be done but I will either punt on that until later (just don't write any ambiguous instances!) or do this in a very simple but restrictive way.

Marwes:
:/ All tests pass on my machine (Windows), going to get a vm running to hopefully reproduce.

Got `[Eq a] -> Eq (List a)` working just need to stamp out edge cases.

PR #438, opened 2018-02-03
Use check_translation in all tests

zummenix:
- move the `check_translation` function up and use it in all tests.
- consistent formatting for hanging `";` and `"#;`


Marwes:
Looks good! Thanks! I tend to factor out the duplication after I have written a few tests but sometimes I forget to go back and update the old ones.

Issue #405, opened 2017-11-28
Run clippy


Marwes:
It has been a while since I last ran clippy on gluon so it would be a good idea to run it again and fix whatever warnings pop up (ignoring whatever warnings seems unhelpful).

Bonus points for setting up automatic clippy runs on travis like [serde has](https://github.com/serde-rs/serde/blob/c650a92bf7773af66e41add39d79c226f982ff08/travis.sh#L24). Travis runs with clippy should probably be allowed to fail though to prevent the build from breaking when clippy updates https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

Dylan-DPC:
Hi

I decided to work on this but when i run `cargo test` i receive the following errors on master:

```
error[E0433]: failed to resolve. Could not find `ser` in `api`
  --> examples/marshalling.rs:35:14
   |
35 |         api::ser::Ser(self).push(thread, context)
   |              ^^^ Could not find `ser` in `api`

error[E0433]: failed to resolve. Could not find `de` in `api`
  --> examples/marshalling.rs:41:14
   |
41 |         api::de::De::from_value(thread, value).0
   |              ^^ Could not find `de` in `api`

error[E0433]: failed to resolve. Could not find `typ` in `api`
  --> examples/marshalling.rs:51:28
   |
51 |     let enum_source = api::typ::make_source::<Enum>(&thread)?;
   |                            ^^^ Could not find `typ` in `api`

error[E0277]: the trait bound `str: std::marker::Sized` is not satisfied
  --> examples/marshalling.rs:51:9
   |
51 |     let enum_source = api::typ::make_source::<Enum>(&thread)?;
   |         ^^^^^^^^^^^ `str` does not have a constant size known at compile-time
   |
   = help: the trait `std::marker::Sized` is not implemented for `str`
   = note: all local variables must have a statically known size

error[E0277]: the trait bound `str: std::marker::Sized` is not satisfied
  --> examples/marshalling.rs:51:23
   |
51 |     let enum_source = api::typ::make_source::<Enum>(&thread)?;
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `str` does not have a constant size known at compile-time
   |
   = help: the trait `std::marker::Sized` is not implemented for `str`
   = note: all local variables must have a statically known size

error: aborting due to 5 previous errors

error: Could not compile `gluon`.
```

[...]

PR #442, opened 2018-02-05
Fix all clippy warnings in completion module

zummenix:
There are some left that I'm not ready to do now:
```
warning: consider using `Option<T>` instead of `Option<Option<T>>` or a custom enum if you need to distinguish all 3 cases
   --> completion/src/lib.rs:208:12
    |
208 |     found: Option<Option<Match<'a>>>,
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: #[warn(option_option)] on by default
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.184/index.html#option_option

warning: this `match` has identical arm bodies
   --> completion/src/lib.rs:735:19
    |
735 |             }) => id.name.clone(),
    |                   ^^^^^^^^^^^^^^^
    |
    = note: #[warn(match_same_arms)] on by default
note: same as this
   --> completion/src/lib.rs:730:19
    |
730 |             }) => id.name.clone(),
    |                   ^^^^^^^^^^^^^^^
note: consider refactoring into `Match::Expr(&Spanned {
                value: Expr::Ident(ref id),
                ..
            }) | Match::Pattern(&Spanned {
                value: Pattern::Ident(ref id),
                ..
            })`
   --> completion/src/lib.rs:730:19
    |
730 |             }) => id.name.clone(),
    |                   ^^^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.184/index.html#match_same_arms

warning: this argument is passed by value, but not consumed in the function body
  --> completion/tests/suggest.rs:30:25
   |
30 | fn suggest_query(query: SuggestionQuery, s: &str, pos: BytePos) -> Result<Vec<Suggestion>, ()> {
   |                         ^^^^^^^^^^^^^^^ help: consider taking a reference instead: `&SuggestionQuery`
   |
   = note: #[warn(needless_pass_by_value)] on by default
   = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.184/index.html#needless_pass_by_value
```

### DIFF
--- a/completion/src/lib.rs
+++ b/completion/src/lib.rs
@@ -202,10 +202,16 @@ impl<E: TypeEnv> OnFound for Suggest<E> {
     }
 }
 
+enum Location<'a> {
+    NotFound,
+    Empty,
+    Found(Match<'a>),
+}
+
 struct FindVisitor<'a, F> {
     pos: BytePos,
     on_found: F,
-    found: Option<Option<Match<'a>>>,
+    found: Location<'a>,
     enclosing_matches: Vec<Match<'a>>,
     near_matches: Vec<Match<'a>>,
 }
@@ -249,7 +255,7 @@ where
     type Ident = Symbol;
 
     fn visit_expr(&mut self, e: &'a SpannedExpr<Self::Ident>) {
-        if self.0.found.is_none() {
+        if let Location::NotFound = self.0.found {
             if e.span.expansion_id == NO_EXPANSION {
                 self.0.visit_expr(e);
             } else {
@@ -315,13 +321,13 @@ where
                     current.span.start + BytePos::from(id.as_ref().len()),
                 );
                 if id_span.containment(&self.pos) == Ordering::Equal {
-                    self.found = Some(Some(Match::Pattern(current)));
+                    self.found = Location::Found(Match::Pattern(current));
                     return;
                 }
                 let (_, pattern) = self.select_spanned(args, |e| e.span);
                 match pattern {
                     Some(pattern) => self.visit_pattern(pattern),
-                    None => self.found = Some(None),
+                    None => self.found = Location::Empty,
                 }
             }
             Pattern::Record {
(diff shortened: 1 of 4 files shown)


zummenix:
Part of #405

Marwes:
```
warning: consider using `Option<T>` instead of `Option<Option<T>>` or a custom enum if you need to distinguish all 3 cases
   --> completion/src/lib.rs:208:12
    |
208 |     found: Option<Option<Match<'a>>>,
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: #[warn(option_option)] on by default
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.184/index.html#option_option

```

For this, `None` represents that the location has not been found yet and `Some(None)` represents that the location has been found, but there was nothing there. I'd probably just mark that as allow since a custom enum does not really make sense either :/.


The other seems pretty straightforward though but I assume you are finished with this for now? (given `There are some left that I'm not ready to do now:`)

zummenix:
Now as you explained it I'm thinking that we can move this knowledge into the code. `NotFound`, `Empty`, `Found(Match)` variants should mirror this. Though, I didn't look at a scope of changes.

Yes, other is easy. I will try to comeback to this PR during the week if I have time.

brendanzab:
Big fan of moving more domain knowledge into the code. Would greatly help future readers!

zummenix:
It's ready for a final review. I'm not happy with the naming of the `Location` and its variants and open to suggestions.

completion/src/lib.rs
```diff
@@ -204,10 +202,16 @@ impl<E: TypeEnv> OnFound for Suggest<E> {
     }
 }
 
+enum Location<'a> {
```
Marwes: Perhaps `MatchState` would be a better name (a bit bland but at least it is explicit).

zummenix: I like that!

PR #443, opened 2018-02-11
Use error_on_line_overflow option instead of error_on_line_overflow_comments

zummenix:
I'm using `rustfmt 0.3.6-nightly (e0e3e22 2018-01-18)` and it emits a warning:
`Warning: Unknown configuration option error_on_line_overflow_comments`

I'm not sure what version you use, so feel free to close this as you see fit.


Marwes:
I just use automatic formatting from the editor so I never see any warnings. I do remember seeing this some time ago though so 👍

PR #453, opened 2018-02-22
Fix typo in TUTORIAL.md

ethanpailes:
I noticed this when reading through the tutorial. It looks like this definition of the id function dropped an x.


Marwes:
Yup that's right, thanks!

PR #455, opened 2018-02-25
chore: Update log to 0.4 and env_logger to 0.5

Marwes:


Marwes:
Tried to break it of from #425 but it seems to require a stack overflow fix which I can't easily break out so leaving it part of that PR.

PR #460, opened 2018-03-11
Change vim-gluon link to point to gluon-lang/vim-gluon

zummenix:
I noticed that it has more commits.


Marwes:
Thanks! (I thought for sure that I had it changed way back...)

Issue #457, opened 2018-03-07
The error message on unresolved implicits are poor if implicits are imported via `{ ? }`


Marwes:
If an implicit argument could not be resolved the typechecker displays a list of all bindings that were considered. This works well for bindings that can be refereed to directly but when an implicit were imported via `{ ? }` there is no proper binding to refer to.

What happens instead is that the error will display `implicit?<number>`  which is the name of a generated variable.

```f#
let { ? } = import! std.option
let { List } = import! std.list
Cons 1 Nil == Nil
// ^ Unable to resolve impliciit ... candidates: implicit?123.eq, ...
```

Marwes:
One way of fixing this would be to make implicit imports (`?`) be bound on identifiers instead of records. That way there is always an identifier to refer to.

Another way would be to track the position of the import better and say something akin to
```
Unable to resolve implicit candidates:
eq: Eq String imported from line x, column y
let { ? } = import! std.string
       ^
```

Issue #388, opened 2017-11-09
Add functionality for calculating dependency graphs


Marwes:
This should be useful for

* Dead code analysis (warnings about unused bindings and optimization)
* Partial typechecking in the language server (only bindings which may have had their type changed need to be checked)
* Documentation (might be useful to display the graph in docs?)

So the implementation should ideally be fairly flexible so it can be reused. It should probably run on the parser AST rather than core AST since type definitions are stripped in the core AST (otherwise core would be the better choice since it's simpler).

brendanzab:
Why do we strip annotations in core? I would have thought keeping them around was a good idea?

Marwes:
Types are kept so that any core expression always knows its type. *type definitions* (ie Expr::TypeBindings`) are stripped however as those are no longer necessary in compilation.

```f#
type Test = Int
f 1
// ==>
f 1
```

We could keep them around if it becomes necessary but as it is right now they don't affect the rest of the compilation so there was no reason to keep them around.

PR #465, opened 2018-03-13
Use assert_eq with implicit arguments in the test module

### LINKED ISSUE #459: Use implicits in the test module
Lets us remove specialized functions such as these https://github.com/gluon-lang/gluon/blob/989ca97c7c5cde898592178acb70a9006d78ae42/std/test.glu#L21-L23

zummenix:
Closes #459

Wasn't able to convert `tests/pass/alternative.glu`, `tests/pass/map.glu`, `tests/pass/thread.glu`, so still use `old_assert_eq`

@Marwes here are my observations:
- errors need more love. I see you've already created #457. For example, it wasn't clear to me that I need to add `?` to import when using `*>` from `std.applicative`
- added some types to help the compiler; in particular in `tests` definitions, but it seems that to import `Test` from `std.test` is enough

### DIFF
--- a/tests/pass/alternative.glu
+++ b/tests/pass/alternative.glu
@@ -1,4 +1,4 @@
-let { run, writer, old_assert_eq, ? } = import! std.test
+let { run, Test, assert_eq, ? } = import! std.test
 let prelude @ { Alternative } = import! std.prelude
 let { Applicative, (*>) } = import! std.applicative
 let int = import! std.int
@@ -9,10 +9,14 @@ let test_alt ?alt show eq : [Alternative f] -> Show (f Int) -> Eq (f Int) -> _ =
     let { (<|>), or, empty } = import! std.prelude
     let { wrap } = alt.applicative
 
-    let assert = old_assert_eq show eq
+    let assert = assert_eq ?show ?eq
 
     assert empty empty *> assert (empty <|> wrap 1) (wrap 1) *> assert (empty <|> empty) empty
         *> assert (empty <|> empty <|> wrap 10) (wrap 10)
 
-test_alt option.show option.eq
-    *> test_alt list.show list.eq
+let tests: Test () =
+    test_alt option.show option.eq
+        *> test_alt list.show list.eq
+
+run tests
+
--- a/tests/pass/alternative.glu
+++ b/tests/pass/alternative.glu
@@ -1,4 +1,4 @@
-let { run, writer, assert_eq, ? } = import! std.test
+let { run, writer, old_assert_eq, ? } = import! std.test
 let prelude @ { Alternative } = import! std.prelude
 let { Applicative, (*>) } = import! std.applicative
 let int = import! std.int
@@ -9,7 +9,7 @@ let test_alt ?alt show eq : [Alternative f] -> Show (f Int) -> Eq (f Int) -> _ =
     let { (<|>), or, empty } = import! std.prelude
     let { wrap } = alt.applicative
 
-    let assert = assert_eq show eq
+    let assert = old_assert_eq show eq
 
     assert empty empty *> assert (empty <|> wrap 1) (wrap 1) *> assert (empty <|> empty) empty
         *> assert (empty <|> empty <|> wrap 10) (wrap 10)
--- a/tests/pass/lazy.glu
+++ b/tests/pass/lazy.glu
@@ -1,4 +1,4 @@
-let { run, writer, assert_ieq, ? } = import! std.test
+let { run, Test, assert_eq, ? } = import! std.test
 let prelude = import! std.prelude
 let { Applicative, (*>), wrap } = import! std.applicative
 let int = import! std.int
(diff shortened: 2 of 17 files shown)


tests/pass/list.glu
```diff
@@ -1,14 +1,12 @@
...
-let assert_list show eq : Show a -> Eq a -> List a -> List a -> _ = assert_eq list.show list.eq
-let assert_int_list = assert_list int.show int.eq
+let empty_list : List Int =
+    Nil
 
-let test_list =
-    assert_int_list (list.of []) Nil *>
-        assert_int_list (list.of [1, 2, 3]) (Cons 1 (Cons 2 (Cons 3 Nil)))
+let test_list : Test () =
+    assert_eq (list.of []) empty_list *>
```
zummenix: Apparently needs type information here, hence `empty_list : List Int` definition.

tests/pass/string.glu
```diff
@@ -1,58 +1,54 @@
...
+        *> assert_eq (string.from_utf8 [195b, 165b, 195b, 164b, 195b]) (Err ())
+        *> assert_eq (string.from_utf8 [195b, 165b, 195b, 195b, 182b]) (Err ())
 
 let tests =
-    slice_tests *> append_tests *> find_tests
+    slice_tests 
+        *> append_tests
+        *> find_tests
+        *> starts_ends_tests
+        *> trim_tests
```
zummenix: Oops, seems like some tests weren't run :)

tests/pass/list.glu
Marwes: Seems to be a bug in subsumption checking causing the wrong error message to be emitted. I don't think this could every compile without type annotations though since the `a` of `List a` will always be ambiguous.

tests/pass/string.glu
Marwes: :) Need https://github.com/gluon-lang/gluon/issues/388 and then it would be possible to scan for unused variables.

Marwes:
> errors need more love. I see you've already created #457. For example, it wasn't clear to me that I need to add ? to import when using *> from std.applicative

I seem to have forgotten to add that part in the tutorial as well (oops). Yeah I will add a hint to that error message as `HINT: Did you forget to import implicits with '?' ` or something

tests/pass/alternative.glu
```diff
@@ -9,7 +9,7 @@ let test_alt ?alt show eq : [Alternative f] -> Show (f Int) -> Eq (f Int) -> _ =
     let { (<|>), or, empty } = import! std.prelude
     let { wrap } = alt.applicative
 
-    let assert = assert_eq show eq
+    let assert = old_assert_eq show eq
```
Marwes: This could be `assert_eq ?show ?eq`

tests/pass/lazy.glu
```diff
@@ -7,5 +7,8 @@ let { lazy, force } = import! std.lazy
 
 let l = lazy (\_ -> 123 + 57)
 
-assert_ieq (force (lazy (\_ -> 2))) 2 *> wrap ()
-    *> assert_ieq 180 (force l)
+let tests : Test () =
```
Marwes: These type annotations does not appear to be necessary (no need to remove them though). I will look into why importing ``Test` is necessary though, way way back that used to be necessary in a few cases so there may be some old code backfiring on implicits now. Even if it ends up being needed to shoudl at least give abetter error than it does now.

tests/pass/list.glu
Marwes: (Haskell errors on `[] == []` as well)

tests/pass/alternative.glu
Marwes: Seems to be some problems writing it like  that 🤔  need to look into why.

tests/pass/string.glu
zummenix: Auto test runner would help. Currently we need to do two steps: write a test case and don't forget to add it in `run`.

Marwes: Yeah. Perhaps the rudimentary attribute system I added with `@implicit` could be used with `@Test` attributes and procedural macros to rewrite the AST.

```f#
/// @Test
let test1 = ...

/// @Test
let test2 = ...

()
```
gets rewritten to
```f#
/// @Test
let test1 = ...

/// @Test
let test2 = ...
(test1 *> test2)
```

(if the `let`s don't end with `()` it could error.

Or perhaps a `test!` macro which takes a record.

```f#
test! {
    test1 = ...
    test2 = ...
}
```

```f#
test "test1 (..) *> test "test2" (..)
```

In any case it should still be easy to programmatically construct test. I want to further integrate the tests inside the `.glu` files with https://github.com/Marwes/tensile so that it is possible to see success and failure of each individual test case as well as filter on individual tests (currently it is only possible to filter on a per-file basis).

(Tensile is based on http://haskell.github.io/test-framework/)

Marwes: https://github.com/gluon-lang/gluon/issues/468

zummenix:
`tests/pass/map.glu` and `tests/pass/thread.glu` are left. I think I will finish them tomorrow.

tests/pass/list.glu
Marwes: https://github.com/gluon-lang/gluon/pull/475 makes this error on an infinite loop which is correct but we could do better https://github.com/gluon-lang/gluon/issues/476

zummenix:
I think it's ready for the final review.

PR #466, opened 2018-03-13
Add basic REPL tests using rexpect.

### LINKED ISSUE #51: Make it possible to test the REPL
As more features gets added to the REPL it would be nice to improve the testing for it. The REPL can however be tricky to test as one needs to be able to send commands to the terminal programmatically. In python I found the [pexpect](https://github.com/pexpect/pexpect) library which seems to be able to do this. It is a bit of a heavy weight solution to take on a python dependency just for some tests though and it would really nice if a library like pexpect could be ported to RUST. I don't have a good estimate to how much work such a port would be but it might be an interesting project to do (which would likely be useful outside of gluon as well)!

memoryruins:
Closes #51

### DIFF
--- a/repl/tests/tests.rs
+++ b/repl/tests/tests.rs
@@ -73,6 +73,10 @@ static PROMPT: &str = "> ";
 
 macro_rules! test {
     ($b: block) => {
+        if ::std::env::var("GLUON_PATH").is_err() {
+            ::std::env::set_var("GLUON_PATH", "..");
+        }
+
         || -> Result<()> { 
             $b
         }().unwrap_or_else(|err| panic!("{}", err));
@@ -84,7 +88,7 @@ fn prompt() {
     test!({
         let mut repl = spawn(COMMAND, Some(TIMEOUT))?;
         repl.exp_string(PROMPT)?;
-        
+
         Ok(())
     });
 }
@@ -97,14 +101,13 @@ fn exit() {
 
         repl.send_line(":q")?;
         repl.exp_eof()?;
-        
+
         Ok(())
     });
 }
 
 
 #[test]
-#[ignore] // TODO fix intermittent test failure
 fn hello_world() {
     test!({
         let mut repl = spawn(COMMAND, Some(TIMEOUT))?;
@@ -114,7 +117,7 @@ fn hello_world() {
 
         repl.send_line("io.println \"Hello world\"")?;
         repl.exp_string("Hello world")?;
-    
+
         Ok(())
     });
 }
@@ -133,7 +136,7 @@ fn expression_types() {
 
         repl.send_line(":t \"gluon\"")?;
         repl.exp_string("String")?;
-        
+
         Ok(())
     });
 }
@@ -146,7 +149,7 @@ fn names() {
 
         repl.send_line(":i std.prelude.show")?;
         repl.exp_string("std.prelude.show: forall a . [std.prelude.Show a] -> a -> String")?;
-        
+
         Ok(())
     });
 }
@@ -162,7 +165,7 @@ fn comments() {
 
         repl.send_line("1 + 2 /* Calls the + function on 1 and 2 */")?;
         repl.exp_string("3")?;
-    
+
         Ok(())
     });
 }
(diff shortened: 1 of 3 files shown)


repl/tests/tests.rs
```diff
@@ -60,3 +64,112 @@ fn issue_365_run_io_from_command_line() {
...
 }
+
+
+static TIMEOUT: u64 = 10_000;
+static PROMPT: &str = ">";
+
+#[test]
+fn prompt() {
+    let mut repl = spawn("../target/debug/gluon -i", Some(TIMEOUT)).unwrap();
+    repl.exp_string(PROMPT).unwrap();
```
Marwes: The error messages seems rather bad when using unwrap. Perhaps wrap the tests in a result returning function and do `unwrap_or_else(|err| panic!("{}", err))` on that function. Or add a macro like

```rust
macro_rules! t {
    ($e: expr) => { $e.unwrap_or_else(|err| panic!("{}", err)) }
}
t!(repl.exp_string(PROMPT))
```

memoryruins: Agreed about unwrap! That macro will be very useful.

repl/tests/tests.rs
```diff
@@ -60,3 +64,120 @@ fn issue_365_run_io_from_command_line() {
...
+    let mut repl = test!(spawn(COMMAND, Some(TIMEOUT)));
+    test!(repl.exp_string(PROMPT));
+
+    test!(repl.send_line(":q"));
+    test!(repl.exp_eof());
+}
+
+
+#[test]
+#[ignore] // TODO fix intermittent test failure
```
memoryruins: The ``hello_world`` test appears to fail 1/10th of the time with a timeout error and a message containing the expansion of std.io's import and a prompt at the end of the string. Setting #[ignore] and left a TODO note to come back to. Tried ``test!(repl.exp_string(PROMPT));`` before sending another line, also tried letting the process sleep in case it was sending too fast; however, It imports ``std.array`` test without the need to wait/sleep and hasn't experienced any intermittent test failures.

memoryruins:
Strangely panics on that many on the CI, yet continually passes on both my laptop and desktop.

brendanzab:
Oooh, this is nice! Wonder if the type system tests could be done in a more integrationy way like this...?

brendanzab:
I guess one issue might be that the repl requires gluon, so if you broke something you wouldn't be able to test it until getting it working enough to run the repl…

brendanzab:
I updated the description to say, "Closes #51". This will auto close the issue once this is merged.

repl/tests/tests.rs
Marwes: You are getting panics sometimes because one of the already existing tests sets an environment variable which adjusts the search path for modules https://github.com/gluon-lang/gluon/pull/466/files#diff-6bf6c6682fd277dbda0194c680c05591R45 . It should just be setting that variable for the command itself instead of globally in the current executable (I think I hacked that in pretty late in the night...).

If you add that environment variable to all the rexpect tests as well they should work consistently.

(The reason for the environment variable is because tests get run from the `./repl` sub folder instead of running from the root folder of the project)

Marwes:
> Oooh, this is nice! Wonder if the type system tests could be done in a more integrationy way like this...?

Hmm, any example on what you mean? Some sort of way of specifying that `let binding 'abc' should have type 'T'` and one could specify multiple such queries for each test?

repl/tests/tests.rs
memoryruins: Ayyy once again your suggestion worked, and it made travis pass!

Marwes:
Looks good! Thanks for the PR (and for pointing me to rexpect!)

memoryruins:
Absolutely! @philippkeller your great work on rexpect is very appeciated 🙏

philippkeller:
nice! That's exactly the use case I had in mind with rexpect. Great that this is actually used :-) If you have any issues or wishes, feel free to open issues at the [github repo](https://github.com/philippkeller/rexpect)

PR #475, opened 2018-03-14
fix(check): Don't generalize tail expressions before replacing implic…

Marwes:
…its.rs


Marwes:
cc https://github.com/gluon-lang/gluon/issues/476

Issue #431, opened 2018-01-28
Create a mdbook out of the tutorial


Marwes:
Inspired by https://github.com/lalrpop/lalrpop/issues/292 I think it would be a good idea to make the current tutorial into a mdbook.

While there is a lot of content missing from the current tutorial to make it a "proper book". What is currently there won't be hurt by a little bit more structure!

https://github.com/rust-lang-nursery/mdBook

Marwes:
Book is up at http://gluon-lang.github.io/gluon/. It needs some work to break apart some of it and reorder and rewrite other parts but the deployment process is at least there. Eventually I want to move it over to https://gluon-lang.org together with the generated API documentation https://github.com/gluon-lang/gluon/pull/456

PR #477, opened 2018-03-14
docs: Make the tutorial a mdbook 

Marwes:
Part of #431 . The book needs to be reorganized a bit but I just want to get travis building and deploying this first.


memoryruins:
Love that that you're putting it in a mdbook!

Wow, after running the hello_world test passed many times locally and then it passing on travis before, it re-appears here.

PR #486, opened 2018-03-17
Wrap repl tests in interface.

memoryruins:
A simple interface to ensure various lines are not forgetten to improve reliability of repl tests. Let's see how these fair, especially so more tests can be written!

### DIFF
--- a/repl/tests/rexpect.rs
+++ b/repl/tests/rexpect.rs
@@ -3,11 +3,8 @@
 extern crate rexpect;
 
 use rexpect::spawn;
+use rexpect::session::PtySession;
 use rexpect::errors::*;
-
-static COMMAND: &str = "../target/debug/gluon -i";
-static TIMEOUT: u64 = 10_000;
-static PROMPT: &str = "> ";
 
 macro_rules! test {
     ($b: block) => {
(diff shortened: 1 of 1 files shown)


repl/tests/rexpect.rs
```diff
@@ -22,126 +19,137 @@ macro_rules! test {
...
+
+        if let Some(string) = expect {
+            self.session.exp_string(string)?;
+        }
+
+        self.session.exp_string(self.prompt)?;
+        Ok(())
+    }
+
+    fn quit(&mut self) -> Result<()> {
```
Marwes: 👍

repl/tests/rexpect.rs
```diff
@@ -22,126 +19,137 @@ macro_rules! test {
...
+        let mut session = spawn(command, Some(timeout))?;
+        session.exp_string(prompt)?;
+
+        Ok(REPL { session, prompt })
+    }
+
+    /// Ensures certain lines are expected to reduce race conditions.
+    /// If no ouput is expected or desired to be tested, pass it an Option::None,
+    /// causing rexpect to wait for the next prompt.
+    fn test(&mut self, send: &str, expect: Option<&str>) -> Result<()> {
```
Marwes: Since we wrap `rexpect` I think we might as well do `unwrap_or_else` in these wrapper methods. That would let the `test` macro be removed (making it simpler to add more tests later).

```rust
fn test(&mut self, send: &str, expect: Option<&str>) {
    self.test_(send, expect).unwrap_or_else(|err| panic!("{}", err));
}
fn test_(&mut self, send: &str, expect: Option<&str>) -> Result<()> {
```

repl/tests/rexpect.rs
```diff
@@ -22,126 +19,137 @@ macro_rules! test {
...
+
+impl REPL {
+    /// Defines the command, timeout, and prompt settings.
+    /// Wraps a rexpect::session::PtySession. expecting the prompt after launch.
+    fn new() -> Result<REPL> {
+        let command = "../target/debug/gluon -i";
+        let timeout: u64 = 10_000;
+        let prompt: &'static str = "> ";
+
+        let mut session = spawn(command, Some(timeout))?;
```
Marwes: Could use [spawn_command](https://docs.rs/rexpect/0.3.0/rexpect/session/fn.spawn_command.html) here.

```rust
spawn_command(Command::new(command).arg("-i").env("GLUON_PATH", ".."), Some(timeout))
```

repl/tests/rexpect.rs
```diff
@@ -22,126 +19,137 @@ macro_rules! test {
...
+    prompt: &'static str,
+}
+
+impl REPL {
+    /// Defines the command, timeout, and prompt settings.
+    /// Wraps a rexpect::session::PtySession. expecting the prompt after launch.
+    fn new() -> Result<REPL> {
+        let command = "../target/debug/gluon -i";
+        let timeout: u64 = 10_000;
+        let prompt: &'static str = "> ";
```
Marwes: Perhaps the prompt should match on the regex `^> ` to make it a little bit more unlikely that one writes a flaky test? (It would still be possible to break by doing `io.println "> "` however so we would still need to be a little bit careful in the tests).

It should work well enough either way since any test can be written in such a way that it does not print out something that looks like the prompt.

repl/tests/rexpect.rs
memoryruins: The .env method returns a &mut Command, while spawn_command only wants a Command, Moved the env part of the macro to new().

repl/tests/rexpect.rs
memoryruins: I wasn't able to make a regex behave as planned, but I agree that there should be a more robust option than the ">" string. It will definitely be good to keep that in mind when writing more tests.

PR #505, opened 2018-04-02
feat: Integrate codespan for error reporting

Marwes:
Got a few changes for codespan itself which I will make a PR for as well before this is working.


brendanzab:
Wooo! Will have to publish to crates.io at some stage too.

PR #540, opened 2018-05-19
Allow types to be passed through the record! macro

Marwes:


Marwes:
Got this just about working locally, just need to add tests to the http example so that it does not bitrot again.

Issue #539, opened 2018-05-19
Marshalling Rust enums


Laegluin:
How can I marshal a Rust enum to Gluon? I only need an untagged enum, but I assume there is some way to marshal enums that are only made up of types that implement `VmType`. I know you can simply define the enum in Gluon, but that is really inconvenient when you it to be part of a native module, so that you can import from the same module as the functions that take it as argument.

If this is possible, it would probably not hurt to add a marshalling chapter in the book, explaining how to work with complex types. I'll try to do that as soon as I get it working :)

Marwes:
Hello!

So there is a marshalling example at https://github.com/gluon-lang/gluon/blob/master/examples/marshalling.rs which I believe should cover what you want, it is slightly awkward in how you need to use the generated code but it should at least be automatic once you get past that.

Having a chapter covering the topc in the book would definitely be a good idea as well.

Laegluin:
Thanks, that seems like I can make it work!

So as far as I can tell, I'll have to generate the source code for the enum and then load it as a module I can import from the actual module I want to expose. Does that mean I have to always write a Gluon wrapper for my module?

Lets say I want to add some module like this:

```rust
let module = record! {
    function => primitive!(2 some_function),
};
```

Ideally I could just add an enum that my function needs like this:

```rust
let module = record! {
    function => primitive!(2 some_function),
    Enum => some_macro!(Enum),
};
```

If I understand the example correctly, what I have to do instead is:

- generate the code for the enum and add it as module
- add the module containing the functions I want to add
- create a third module that imports both the enum and the functions and exposes them as a single module

It just seems like a lot of work for something that seems to be a quite common use case. Is it simply not possible to add a type using the `record!` macro?

Marwes:
> It just seems like a lot of work for something that seems to be a quite common use case. Is it simply not possible to add a type using the record! macro?

Yeah, that is not very user friendly. It is possible to extend `record!` to support types as well though, I started to add it with https://github.com/gluon-lang/gluon/pull/540 as it will make it possible to export some types in the standard library as well (which currently cannot be refereed to from gluon scripts). Still need to verify that it works as I want on the http example though which needs to do this as well.

Laegluin:
Looks great!

[...]

PR #541, opened 2018-05-21
[WIP] Derive Getable/Pushable

Laegluin:
As mentioned in #539, I started playing around with deriving `Getable`/`Pushable`. The current implementation for `Getable` works quite well for normal enums, but does not work properly with lifetimes yet (which make a lot of sense for things like `&str`). I'm not sure how I can marshal generic parameters (looking at things like `api::generic::*` I assume there it's possible somehow).

`Pushable` should be a lot easier to implement because the data needs to be copied anyway, meaning there are no lifetimes getting in the way.


Laegluin:
> Lifetimes should be possible but I believe I need to change the trait and some code around it as it is currently only allowed using unsafe + some extra assumptions about the input. I'd recommend holding of on that for the moment until I can add a safe API around that.

Yeah, I just stumbled upon that. Doesn't matter much though, most of the trouble is generating the correct lifetime bounds and that part I managed to get working. I have simply disabled the test for now.

> Generic parameters in Rust get a Getable/Pushable bound and generic parameters on gluon's side are stored as opaque values only (which have their own custom Getable/Pushable) impls already.

Generating the code is not the problem, but you always have to instantiate your generic type before you marshal it. You cannot do something like this:

```rust
enum Enum<T> {
}
```

codegen/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = "0.14"
+quote = "0.6"
+proc-macro2 = "0.4"
+
+gluon = { version = "0.7.1", path = ".." } # GLUON
```
Marwes: I don't think gluon is actually required for this crate, right? (Only the generated code refer to it)

Having the dev-dependencis on gluon is ok though.

codegen/src/getable.rs
```diff
@@ -0,0 +1,223 @@
...
+    // for the enum
+    match &variant.fields {
+        Fields::Unit => quote! {
+            #tag => #ident::#variant_ident
+        },
+        // both constructors that need to marshall values extract them by using the index
+        // of the field to get the content from Data::get_variant;
+        // the data variable was assigned in the function body above
+        Fields::Unnamed(FieldsUnnamed { unnamed, .. }) => {
+            let cons = gen_tuple_cons(unnamed.into_iter().collect());
```
Marwes: Could do without the collect here if the function were to take an iterator or the original `unnamed` slice

codegen/src/getable.rs
```diff
@@ -0,0 +1,223 @@
...
+
+            quote! {
+                #tag => #ident::#variant_ident#cons
+            }
+        }
+        Fields::Named(FieldsNamed { named, .. }) => {
+            let cons = gen_struct_cons(named.into_iter().collect());
```
Marwes: Ditto

codegen/src/getable.rs
```diff
@@ -0,0 +1,223 @@
...
+
+            quote! {
+                if let Some(val) = data.get_variant(#idx) {
+                    <#field_ty as ::gluon::vm::api::Getable<'__vm>>::from_value(vm, val)
+                } else {
+                    panic!("Enum does not contain data at index '{}'. Do the type definitions match?", #idx)
+                }
+            }
+        })
+        .collect();
```
Marwes: Unnecessary collect (I think)

codegen/src/getable.rs
```diff
@@ -0,0 +1,223 @@
...
+
+            quote! {
+                #field_ident: if let Some(val) = data.get_variant(#idx) {
+                    <#field_ty as ::gluon::vm::api::Getable<'__vm>>::from_value(vm, val)
+                } else {
+                    panic!("Enum does not contain data at index '{}'. Do the type definitions match?", #idx)
+                }
+            }
+        })
+        .collect();
```
Marwes: Ditto

Marwes:
> Generating the code is not the problem, but you always have to instantiate your generic type before you marshal it. You cannot do something like this:

So for generic types, the rust types do need to be instantiated (otherwise nothing is executed at runtime). So the `Generic<T>` and `generic::A`, `generic::B` etc types are provided for this purpose.

So for `enum Enum<T> { .. }` would be written as either `Enum<generic::A>` or `Enum<Generic<generic::A>>` and it will create the correct type. If you actually want to use the type as a value then there is also the `OpaqueValue<T>` wrapper which also handles rooting (`Generic` will probably be made into an implementation detail eventually) so you would use it as `OpaqueValue<Generic<A>>`

codegen/Cargo.toml
Laegluin: I just checked, I do actually use gluon for the `VmTag` typedef. I can just replace it with `u32` of course, but that would kind defeat the purpose.

Marwes: I think it is fair to just use `usize` for it. If someone happens to have more than `VmTag` variants then they will still get a warning (if they somehow manage to compile a 4>GB file that is :). Not needing the dependency is nice though as it opens up the possibility to use this crate in the gluon and gluon_vm crates aswell.

Laegluin:
Removed the collects with https://github.com/gluon-lang/gluon/pull/541/commits/442c1cf04824eca0d3ef690b18b8836bc8869b81. In theory there's two more collects I can remove, but that's really only readable when using impl Trait and that would drop support for all older rust versions.

codegen/Cargo.toml
Laegluin: Yeah, fair enough, I'll remove it :D

Marwes:
> Removed the collects with 442c1cf. In theory there's two more collects I can remove, but that's really only readable when using impl Trait and that would drop support for all older rust versions.

`impl Iterator` is already in use so I am not personally worried about needing 1.26 https://github.com/gluon-lang/gluon/blob/120a91e80d56c5669886a63f6bbb197e674bd13b/base/src/metadata.rs#L76 . I should maybe back that change out though for 0.8 since it is the only thing that require 1.26 (otherwise I think there is support for 1.20-ish or something). Releases after 0.8 will need >= 1.26 though as I can remove a workaround/experiment around minimizing `Box<Future>` uses.

Laegluin:
> I should maybe back that change out though for 0.8 since it is the only thing that require 1.26 (otherwise I think there is support for 1.20-ish or something)

Yeah it probably makes sense to give people a few weeks to switch to the new version. I imagine most will be pretty eager to upgrade to 1.26 anyway. I'll just leave it as is for now, for a macro it probably does not make a big difference either way.

> So for generic types, the rust types do need to be instantiated (otherwise nothing is executed at runtime). So the Generic<T> and generic::A, generic::B etc types are provided for this purpose.

When I do that I can make the rust code compile, but the gluon side fails during typechecking. I'll just ignore it for now, there's probably something stupid I'm missing :P

Marwes:
> I'll just leave it as is for now, for a macro it probably does not make a big difference either way.

True 👍 .

> When I do that I can make the rust code compile, but the gluon side fails during typechecking. I'll just ignore it for now, there's probably something stupid I'm missing :P

Can you post an example/error message? I may have an idea about what the issue is but can't say for sure.

Laegluin:
Tried with this enum:

```rust
enum GenericEnum<T> {
    Val(T)
}
```

`make_source` doesn't work with it, since it doesn't implement the serde traits, so I defined the gluon type by hand:

```ocaml
type GenericEnum a = | Val a
```

The function I added looks like this:

```rust
fn generic_enum_to_str(val: GenericEnum<Generic<generic::A>>) -> String {
    format!("{:?}", val)
}
```

Using it like this

```ocaml
assert (generic_enum_to_str (Val "str") == "Val(\"str\")")
```

causes this error:
```
Expected: types.GenericEnum
Found: types.GenericEnum String
The use of self recursion in type `types.GenericEnum` could not be unified.
- <test>:6:38
6 |         assert (generic_enum_to_str (Val "str") == "Val(\"str\")")
  |       
```

Laegluin:
The `Pushable` derive already works quite nicely, but there's a lot of duplication. I'll probably add derives for `Userdata` and `VmType` just for consistency, so you can derive everything you need for marshalling, and then clean up a bit.

codegen/tests/derive_getable.rs
```diff
@@ -0,0 +1,246 @@
...
+{
+    type Type = GenericEnum<'static, T>;
+
+    fn make_type(vm: &Thread) -> ArcType {
+        vm.global_env()
+            .get_env()
+            .find_type_info("types.GenericEnum")
+            .unwrap()
+            .into_owned()
+            .into_type()
```
Marwes: > make_source doesn't work with it, since it doesn't implement the serde traits, so I defined the gluon type by hand:

So the problem here is that you only refer to a plain `GenericEnum` here. But `GenericEnum` needs to be applied to another type before it becomes `GenericEnum<T>/GenericEnum t`.

What you need to do is to do the same thing as the http example does here https://github.com/gluon-lang/gluon/blob/120a91e80d56c5669886a63f6bbb197e674bd13b/examples/http.rs#L58-L69 or the option type does here https://github.com/gluon-lang/gluon/blob/120a91e80d56c5669886a63f6bbb197e674bd13b/vm/src/api/mod.rs#L907-L919  and apply the type to `T::make_type(vm)`

codegen/src/pushable.rs
```diff
@@ -0,0 +1,179 @@
...
+        .rev();
+
+    // since the number of fields is statically known, we can allocate all values in an array
+    let array_init = iter::repeat(quote! { ::gluon::vm::internal::Value::int(0) }).take(len);
+    let stack_init = quote! {
+        let mut fields = [ #(#array_init),* ];
+
+        for i in 0..#len {
+            fields[i] = ctx.stack.pop();
+        }
```
Marwes: Could generate this as just
```rust
let array_init = iter::repeat(quote! { ctx.stack.pop() }).take(len);
let stack_init = quote! {
        let fields = [ #(#array_init),* ];
```

and avoid the initialization step altogether

Laegluin: Yep, that's a lot cleaner and doesn't rely on the compiler optimizing things correctly. Already fixed it with https://github.com/gluon-lang/gluon/pull/541/commits/90a428cd6977511ba7687a9e261023aaf14b57ed :)

codegen/tests/derive_getable.rs
Laegluin: Ah yes, that makes much more sense. I probably won't be able to try before friday though.

Laegluin:
Just tested all the generic stuff, works perfectly once you actually do it right!

Marwes:
Haven't had the time to take another look so far over the weekend (will have Sunday evening probably).

Tests are failing ontravis because they are run with --features test which changes how the standard library is imported. You can look at what the repl's test does to initialize the vm (it adds the path `..` to ensure the standard library gets found).

Laegluin:
This should be pretty much the complete implementation. The only missing now are docs (a chapter in the book and detailed module level explanation of how the derives work and what you can do with them). I should be able to get around to that tomorrow.

codegen/src/vm_type.rs
```diff
@@ -0,0 +1,124 @@
...
+        #where_clause #(#trait_bounds,)* #(#lifetime_bounds),*
+        {
+            type Type = Self;
+
+            fn make_type(vm: &::gluon::vm::thread::Thread) -> ::gluon::base::types::ArcType {
+                let env = <::gluon::vm::thread::Thread as ::gluon::vm::thread::ThreadInternal>::global_env(
+                    vm,
+                ).get_env();
+
+                let info = env.find_type_info(#gluon_type);
```
Marwes: It is actually possible to call `find_type_info` directly on `vm` here (not necessary to go through `THreadInteral`

Marwes:
Unless you have something more to add this looks good to me! If there are any changes necessary it is likely we will only find them after using this for a bit.

codegen/src/vm_type.rs
Laegluin: Must have overlooked that function, fixed it :)

Laegluin:
> Unless you have something more to add this looks good to me! If there are any changes necessary it is likely we will only find them after using this for a bit.

I'm still working on a chapter for the book for all of this, once I finished that I think this should be ready! Long term it would probably be nice to integrate the `VmType` derive with the changes added in #540, but I'll just open a new issue when I have some time to look at that.

Laegluin:
> Yeah, I thought about that but I figure there are some issues with the current Deserialize/Serialize approach that needs to be hashed out before "blessing" it with a more automatic approach of just using derive with it.

Yeah, I think the best approach might be to have a trait for generating the `Type` that can be derived and then providing a wrapper struct that implements that trait when `T: Serialize + Deserialize`.

Laegluin:
This should be everything! It's probably best if you check everything for mistakes and weird wording, then I can go ahead and rebase it.

book/src/marshalling-types.md
```diff
@@ -0,0 +1,252 @@
...
+
+Gluon provides several traits for safely marshalling types to and from Gluon code:
+
+- [VmType][] provides a mapping between Rust and Gluon types. It specifies the Gluon type
+  the implementing Rust type represents. All types that want to cross the Gluon/Rust boundary
+  must implement this trait.
+
+- [Getable][]: Types that implement `Getable` can be marshalled from Gluon to Rust. This
+  means you can use these types anywhere you are receiving values from Gluon, for example
+  as parameters for a function.
```
Marwes: Might be good to make it explicit that it is for functions implement in rust, so perhaps `for a function implemented on the Rust side`?

book/src/marshalling-types.md
```diff
@@ -0,0 +1,252 @@
...
+
+- [Pushable][] is the counterpart to `Getable`. It allows implementing types to be marshalled
+  to Gluon. Values of these types can returned from embedded rust functions.
```
Marwes: `and be used as parameters to gluon functions`?

book/src/marshalling-types.md
```diff
@@ -0,0 +1,252 @@
...
+the Gluon type the Rust type maps to, using the `#[gluon(vm_type = "<gluon_type>")]` attribute.
+
+`Userdata` can be derived for any type as long as it is `Debug + Send + Sync` and has a `'static`
+lifetime.
+
+### Implementing by hand
+
+The following examples will all assume a simple struct `User<T>`, which is defined in a different
+crate. To implement the marshalling traits, we to have create a wrapper and implement the traits
+for it.
```
Marwes: Since the examples were added to `examples/marshalling.rs` it would useful to have a link to them so that readers can lookup the complete implementation!

book/src/marshalling-types.md
Laegluin: Being explicit certainly won't hurt, I'll add it.

book/src/marshalling-types.md
Laegluin: Didn't even consider that, the reverse also applies to `Getable` now that I think about it.

book/src/marshalling-types.md
Laegluin: Yup, makes sense!

Laegluin:
Should be all fixed now :)

PR #567, opened 2018-06-27
Fix repl record destructuring

mikeyhew:
This fixes the record-restructuring bug in the repl, where records are treated like tuples and
 `let {foo} = {bar = “bar”, foo = “foo”}` followed by `foo` displays `”bar”`.

I couldn’t access any of the existing `GcPtr<DataStruct>::get_field` or `Data::lookup_field`
methods from the repl crate, so I added one to `RootedValue`. It makes things consistent
since there is already a `get` method that takes an index.

There are a few FIXME notes where I could use some advice on what to do.


Marwes:
> I couldn’t access any of the existing GcPtr<DataStruct>::get_field or Data::lookup_field
methods from the repl crate, so I added one to RootedValue. It makes things consistent
since there is already a get method that takes an index.

👍  Sorry! Misread that when I searched for that method. Thought it was under `Value`

repl/src/repl.rs
```diff
@@ -362,17 +363,41 @@ fn set_globals(
...
-                .iter()
-                .zip(::vm::dynamic::field_iter(&value, typ, vm));
-            for (field, (field_value, field_type)) in iter {
-                match field.value {
-                    Some(ref field_pattern) => {
-                        set_globals(vm, field_pattern, &field_type, &field_value)?
+            // FIXME: not sure if this is necessary, or if this has already
+            // been done. ::vm::dynamic::field_iter did this
+            // so I'm doing it too
+            let resolved_type = resolve::remove_aliases_cow(&*vm.global_env().get_env(), typ);
```
Marwes: 👍 This is necessary. Since `type` could be an alias `MyRecord` we need to remove it to look at the type inside the alias. 

```
type MyRecord = { x : Int }
```

repl/src/repl.rs
```diff
@@ -362,17 +363,41 @@ fn set_globals(
...
+                    value.get_field(field_name.declared_name())
+                    // FIXME: can I use the type checker for this error?
+                    // FIXME: apparently this error message isn't actually getting displayed. not sure why
+                    .ok_or_else(|| VMError::Message(format!(
+                        "The record doesn't have a field by the name `{}`",
+                        field_name.declared_name()
+                    )))?;
+                let field_type = match **resolved_type {
+                    Type::Record(ref row) => match **row {
+                        Type::ExtendRow {ref fields, ..} => {
```
Marwes: It is actually not necessary to match on the type here, you can instead call `resolved_type.row_iter()` and you will get a iterator over the fields (I don't know why `dynamic::field_iter` does an explicit match...)

repl/src/repl.rs
```diff
@@ -362,17 +363,41 @@ fn set_globals(
...
+            // been done. ::vm::dynamic::field_iter did this
+            // so I'm doing it too
+            let resolved_type = resolve::remove_aliases_cow(&*vm.global_env().get_env(), typ);
+
+            for pattern_field in fields.iter() {
+                let field_name: &Symbol = &pattern_field.name.value;
+                let field_value: RootedValue<&Thread> =
+                    value.get_field(field_name.declared_name())
+                    // FIXME: can I use the type checker for this error?
+                    // FIXME: apparently this error message isn't actually getting displayed. not sure why
```
Marwes: This should never happen as the typechecker does validate that the pattern is valid.

Marwes:
Failure seems to be a flaky test.

repl/src/repl.rs
mikeyhew: It doesn't output any errors though:

```
>let {x} = {}
>
```

Marwes: 😓 Should have checked what the code actually does after that FIXME. It appears it only compiles with the pattern if the binding is a function which is an error.

If you replace the `eval_expr` creation in `fn eval_line_` with this it will do the necessary rewrite for any type of pattern and actually check it.

```rust
            // We can't compile function bindings by only looking at `let_binding.expr`
            // so rewrite `let f x y = <expr>` into `let f x y = <expr> in f`
            let id = match unpack_pattern.value {
                Pattern::Ident(ref id) if !let_binding.args.is_empty() => {
                    id.clone()
                }
                _ => {
                    let id = Symbol::from("repl_temp");
                    let_binding.name = pos::spanned(let_binding.name.span, Pattern::As(id.clone(), Box::new(let_binding.name)));
                    TypedIdent::new(id)
                }
            };
            let id = pos::spanned2(0.into(), 0.into(), Expr::Ident(id.clone()));
            let expr = Expr::LetBindings(vec![let_binding], Box::new(id));
            let eval_expr = pos::spanned2(0.into(), 0.into(), expr);

```

repl/src/repl.rs
mikeyhew: Thanks! Updated.

mikeyhew:
Got conflicts, do you prefer if I rebase or merge? Either's fine

mikeyhew:
Merging ended up being a lot easier, since there were conflicts every commit it seemed. I can squash later if you want to keep history clean.

repl/src/repl.rs
```diff
@@ -299,18 +300,24 @@ fn eval_line_(
             compiler = compiler.run_io(true);
             expr.run_expr(&mut compiler, vm, "line", line, None).boxed()
         }
-        Some(ReplLine::Let(let_binding)) => {
+        Some(ReplLine::Let(mut let_binding)) => {
```
mikeyhew: I like this new `ReplLine` enum, it will definitely make adding type expressions easier

Marwes:
Squashing these commits seems like a good idea, but I can do that from github's side (by ignoring bors but this PR should be based on master already).

Marwes:
> As you can see, the ideal version using method chaining didn’t work.
The fact that it doesn’t work with NLL seems like a bug, but I gave up
trying to reproduce it

@mikeyhew As an aside, what did the code which caused this error look like?

mikeyhew:
@Marwes this is what I tried first:

```rust
set_globals(&vm, &unpack_pattern, &value.typ, &value.value.as_ref())
.and(Ok(value))
```

with `#[feature(nll)]`, it produces this error:

screen shot 2018-06-27 at 6 48 11 pm

and by changing it to this, it goes away:

```rust
let temp = set_globals(&vm, &unpack_pattern, &value.typ, &value.value.as_ref());
temp.and(Ok(value))
```

here's a branch that demonstrates the error: https://github.com/gluon-lang/gluon/compare/master...mikeyhew:weird-nll-bug

cc @nikomatsakis who asked about it in Discord

Marwes:
Yeah, that is a weird error. Wonder if it has something to do with the expression being returned in the `and_then` closure (tried a bit but couldn't reproduce it either).

PR #586, opened 2018-07-09
feat: Add json deserialization

Marwes:


Marwes:
bors r+

May need to clean up something things later but this works well enough.

Issue #566, opened 2018-06-27
The book doesn't explain how to sequence IO actions


lilyball:
After going through the book, there's no explanation for how to sequence IO actions. It doesn't even explain why the naive approach doesn't work, e.g.

```gluon
let io = import! std.io
io.println "one" // nothing is printed
io.println "two" // this one works
```

I've had previous exposure to Haskell so I understand the IO monad, and the book does explain that when you have multiple statements it evaluates them both but discards the values from all but the last (side note: is this useful for anything at all besides `assert`? What else has side-effects?), so I was able to figure out what was going on. But someone who hasn't seen Haskell before would probably be very confused.

Not only that, but it's not obvious how to proceed. As someone who's used Haskell before, my first thought was to try `do` notation, but it appears that Gluon's `do` notation is literally just an alternative way to write a single call to `flat_map` rather than being like Haskell and handling both `>>=` and `>>` depending on whether the line is an assignment or not (also Gluon's `do` doesn't appear to handle anything other than 2 lines.

Then I tried to use `>>` but quickly determined that Gluon doesn't have a `>>` at all. I could write `io.println "one" >>= \_ -> io.println "two"` but that's pretty gross.

I finally hit upon what I believe to be the correct solution:

```gluon
let io @ { ? } = import! std.io
let { (*>) } = import! std.applicative
io.println "one" *> io.println "two"
```

But this was non-obvious.

Ultimately, I'd like to see a few changes:

1. The book should explicitly mention how to sequence multiple IO actions (and explain why the naive approach doesn't work).
2. Assuming `*>` is the ideal way to handle this, the `*>` operator should be in the prelude.
3. For that matter, `flat_map` (and probably `>>=` and `=<<`) should be in the prelude too. Having to import something in order for a language construct (`do`) to even work is weird.
4. You might consider defining `>>` for Monad (I'm not sure if this should just be defined in terms of `>>=` or if it should be an alias for `*>`)
5. And finally, `do` notation should probably be enhanced to be more like Haskell in that it supports more than 2 expressions (and should probably work with 1 expression too, even though that makes it useless, because it's better than having a confusing syntax error), detects assignments vs other statements, and uses `>>=` and `>>` as appropriate. Such a change would likely require indenting the subsequent lines in the `do` block, otherwise you wouldn't know where the block ends, but that seems fine to me (the rule could just be the second line has to be indented at least one space past the `do` token, and subsequent lines would be indented to the same level).

kballard:
[...]

PR #595, opened 2018-07-31
feat: Add seq expressions

Marwes:
This adds the `seq EXPR1 in EXPR2` syntax which desugars to `flat_map (\_ -> EXPR2) EXPR1`. In the long run we should probably desugar this to

```
let x = lazy (\expr -> EXPR2)
flat_map x EXPR1
// or with **>> : [Applicative f] -> f a -> Lazy (f b) -> f b
// EXPR1 **>> x
```

cc #566

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,9 +184,9 @@ do digits =
     wrap [a, b, c, d]
 
 let print_digits = for digits (\d ->
-        do _ = io.print " "
+        do io.print " "
         io.print (show d))
-do _ = io.print "Four digits:" *> print_digits *> io.println ""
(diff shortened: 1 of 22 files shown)


ebkalderon:
Is this PR still being worked on? This is a pretty nice ergonomics improvement to Gluon that I would hope to see land someday.

Marwes:
@ebkalderon It is still on the table but I found out that `do EXPR in EXPR` and `do PATTERN = EXPR in EXPR` are ambiguous so I need another keyword for it (`seq EXPR in EXPR` is the best I got atm).

Issue #495, opened 2018-03-21
"if .. then .. else" is formatted inconsistently


Marwes:
See https://github.com/gluon-lang/gluon#24

```f#
// Format 1
if x
then y
else z

// Format 2
if x then
    y
else
    z
```

OvermindDL1:
I like the OCaml'y style of:
```ocaml
if blah
then bloop
else bleep
```
Unless used as a kind of early out or so then this common:
```ocaml
if not blah then bloop else
...more stuff...
```

Marwes:
I am in two minds on this, `Format 1` is nice and compact for the simple cases. On the other hand `Format 2` works better when the true and false branches are more complicated.

```
// Format 1
if x
then
    let y = 1
    y
else z

// Format 2
if x then
    let y = 1
    y
else
    let y = 1
    y
```

OvermindDL1:
When the branches are complex then in OCaml I've always done:

```ocaml
if x then
    let y = 1 in
    y
else
    let y = 2 in
    y
```

So the Format 2 in your example.

_/me really dislikes whitespace or newline sensitivity as a side note..._

Marwes:
I have been bitten enough by indentation based syntax to prefer non-whitespace sensitivity but there are some unfortunate ambiguities that would need resolving to let that happen.

```ocaml
(* Fails since the last alternative is associated with the inner match *)
let x = match 1 with
    | 2 ->
        match () with
        | _ -> 1
    | _ -> 0
```

```ocaml
(* Syntax error *)
let x =
    let y = "a"
    y
print_string x;;

(* Add `in` to make it work *)
let x =
    let y = "a"
    in
    y
in
print_string x;;
```

Perhaps there is a a better syntax that would resolve these cases though 🤔

OvermindDL1:
> I have been bitten enough by indentation based syntax to prefer non-whitespace sensitivity but there are some unfortunate ambiguities that would need resolving to let that happen.

Well the OCaml usual way for those are:

```ocaml
let x = match 1 with
    | 2 ->
        begin match () with
        | _ -> 1
        end
    | _ -> 0
```
Also do not that the OCaml community-standard formatter/indenter `ocp_indent` would have made that case blaringly obvious as it would have moved the last case in your example to the right.  :-)

As for the `in`, I really like the `in` bits as it makes it visually, syntactically, and greppably obvious where a statement ends, though I would have formatted it like:
```ocaml
let _ =
  let x =
    let y = "a" in
[...]

PR #596, opened 2018-07-31
[RFC] feat: Remove the indentation aware syntax

Marwes:
Removes the indentation aware syntax and instead forces explicit `in` tokens to be inserted to separate `let/type/do` bindings from their body. By some tweaks in the grammar `in` is not required to separate the bindings itself.

```f#
type Test = Int
let id x = x
and id2 y = y
do action
let x = id 2
in
wrap (x + 1)
```
(Could also be written as)

```f#
type Test = Int
in
let id x = x
and id2 y = y
in
do action
in
let x = id 2
in
wrap (x + 1)
```

I feel like this could be a good compromise between a non-indentation based syntax while also avoiding the verboseness of `in` everywhere (I think this is more or less how OCaml is structured, but it still avoids a concept of top-level).

Since blocks are indentation aware they were also removed so it is no longer possible to sequence two function calls.

```f#
f x
g y z

// Now parses as
f x g y z
```

`match` expressions can no longer be nested directly as that causes ambiguities.

```f#
let x = match 1 with
    | 2 ->
        match () with // Error
        | _ -> 1
    | _ -> 0
```
To fix this the inner `match` must be wrapped in parentheses
```f#

let x = match 1 with
    | 2 ->
        (match () with
        | _ -> 1)
    | _ -> 0
```

Before and after of part of the standard library https://github.com/gluon-lang/gluon/pull/596/commits/b5cfc7476359801c73a81e1165d7244aa151de4a

cc #495 cc @OvermindDL1

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,9 +185,9 @@ do digits =
     wrap [a, b, c, d]
 
 let print_digits = for digits (\d ->
-        do _ = io.print " "
+        do io.print " "
         io.print (show d))
-do _ = io.print "Four digits:" *> print_digits *> io.println ""
+do io.print "Four digits:" *> print_digits *> io.println ""
 
 let guess_loop _ =
     do line = io.read_line
--- a/base/src/ast.rs
+++ b/base/src/ast.rs
@@ -262,7 +262,7 @@ pub struct ExprField<Id, E> {
 
 #[derive(Clone, PartialEq, Debug)]
 pub struct Do<Id> {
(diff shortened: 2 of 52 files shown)


Marwes:
Before and after of part of the standard library https://github.com/gluon-lang/gluon/pull/596/commits/393a555d3eb0ce3c29ebeca61c31631b1649c1ea

PR #606, opened 2018-08-07
Issue 599

### LINKED ISSUE #599: The implementation of higher-rank types is unsound
I was playing around with higher-rank types and discovered that an inner-quantified type could escape, leading to unsoundness:
```
let foo f: forall b. (forall a. a -> b) -> b = f ()
let id x: forall a. a -> a = x
let false: forall a. a = foo id // Oops
let some_int: Int = false
some_int // Undefined behaviour
```
`foo id` should not typecheck, because `forall a. a -> a` and `forall a. a -> b` should not unify.

Marwes:
Fixes #599

### DIFF
--- a/check/src/unify_type.rs
+++ b/check/src/unify_type.rs
@@ -1092,7 +1092,7 @@ pub fn subsumes(
     state: State,
     l: &ArcType,
     r: &ArcType,
-) -> Result<ArcType, Errors<Error<Symbol>>> {
+) -> Result<ArcType, (ArcType, Errors<Error<Symbol>>)> {
     debug!("Subsume {} <=> {}", l, r);
     let mut unifier = UnifierState {
         state: state,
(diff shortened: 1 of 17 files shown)


check/src/unify_type.rs
```diff
@@ -1177,7 +1189,7 @@ impl<'a, 'e> Unifier<State<'a>, ArcType> for UnifierState<'a, Subsume<'e>> {
             }
 
             (_, &Type::Forall(_, _, Some(_))) => {
-                let r = r.instantiate_generics(&mut FnvMap::default());
```
Marwes: This was the actual unsound line. Unfortunately fixing this (and the direct fallout) exposed a unfication ambiguity where I have two sound programs but only one or the other will successfully typecheck depending on some minor tweaks to the typechecker... Researching options for now.

Marwes: See the unify_with_inferred_forall_in_record test https://github.com/gluon-lang/gluon/pull/606/files#diff-697e0a05f79902051453dccb6fd941c2R1050

PR #610, opened 2018-08-15
Correct record test in vm

topelinux:
Use len instead of tag for verify record with 0 field.


Marwes:
Yep, that was wrong, thanks!

bors r+

PR #611, opened 2018-08-19
Derive clone and debug for function

c410-f3r:
Is it OK?


Marwes:
Used to be that function types (`fn (i32) -> String`) didn't implement clone but it looks like they do know 👍 

bors r+

c410-f3r:
Thanks!

PR #617, opened 2018-08-31
Add test and update docs for VmType derive

Laegluin:
While playing around with the new `VmType` derive I noticed there was no test for newtypes. I also updated the docs for the new features, at least where I noticed missing things :)


Marwes:
bors r+

(Flaky test :/)

PR #621, opened 2018-09-02
Print a hint/warning towards necessary feature gates when importing modules

### LINKED ISSUE #620: Print a hint/warning towards necessary feature gates when importing modules
`regex`, `serialization` and `rand` are features that enable additional modules in gluon but if you forget to enable them then you will get a confusing error message about a missing module. If the error pointed to the feature gate then it would be easier to discover.

The documentation should also be updated to mention the feature gates.

Laegluin:
Closes #620

### DIFF
--- a/book/src/standard-types-and-functions.md
+++ b/book/src/standard-types-and-functions.md
@@ -1,6 +1,11 @@
 # Standard types and functions
 
-https://gluon-lang.org/doc/nightly/std/index.html
+The API documentation for the standard library can be found [here][std-docs]. Some of the modules
+are only available if Gluon is compiled with the required features:
+
+- `std.regex` requires the `regex` feature (enabled by default)
+- `std.random` requires the `rand` feature (enabled by default)
+- All `std.json.*` modules require the `serialization` feature
 
 TODO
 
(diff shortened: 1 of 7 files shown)


Laegluin:
This works nicely for things like `std.regex`, but I'm not really happy with the errors when the error is transitive like with `std.random`

Not quite sure how to best do that without overcomplicating the build script though

Marwes:
Yeah that seems awkward :/

Laegluin:
> Seems good to me! What is the problem with errors in nested modules? That it is still possible to import! std.regex and then get an error about std.regex.prim ?

Yeah, the only way I can think of to make it better is to modify the import macro to allow you to disable some of the std modules. Everything else would be messing around with the build script more than I like.

Laegluin:
Something like this: 7a0d372

Marwes:
There would essentially be a need to add substitution primitive modules with the error message for each file that uses the primitive module to make the error occur at the first imported module (so three modules for json ... https://github.com/gluon-lang/gluon/tree/master/std/json).

I don't think it is worth it, I have some patches to reduce the amount of cascading errors as well so ideally, the lack of say `std.regex.prim` shouldn't print out that many errors on top of the actual missing module error.

Laegluin:
Yeah, I'll just remove the last commit real quick and then the rest should be ready to merge.

Marwes:
Yup, thanks!

bors r+

PR #624, opened 2018-09-09
Implement VmType and Getable for Box<T>

ftilde:
The type of `Box<T>` is just `T` and for `Getable` we call `from_value` for `T` and box the result. This is useful because in recursive types in rust we often have to use Box in order for the type to have a finite size.

I *think* this should not cause any problems and with this implementation the user can just choose how to represent their gluon types in rust (i.e., with or without boxing), but feel free to correct me on that. Alternatively, one could implement a `Box`-like type in gluon and map those types, but I currently don't see a reason for that.

Unfortunately, `Pushable` cannot be implemented due to the conflicting implementation for `UserData` and I could not think of a workaround. Apparently, users can implement `UserData` (or any other trait for that matter) for `Box<SomeLocallyDefinedType>`.

I also looked into adding tests for this implementation, but I'm unsure where they should go. Maybe `tests/de.rs`?

### DIFF
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -351,3 +351,24 @@ fn use_rust_created_record_as_polymorphic() {
     let result = f.call(record_no_decl!{ x => 1 }).unwrap();
     assert_eq!(result, 1);
 }
+
+#[test]
+fn get_value_boxed_or_unboxed() {
+    let _ = ::env_logger::try_init();
+    let vm = make_vm();
+
+    let text = r#"
+        27
+    "#;
+
+    // The user should be able to decide whether or not any gluon value should
+    // be boxed on the rust side.
+    let (unboxed, _) = Compiler::new()
+        .run_expr::<i32>(&vm, "test", text)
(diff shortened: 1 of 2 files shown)


Marwes:
Thanks for the PR!.

> I also looked into adding tests for this implementation, but I'm unsure where they should go. Maybe tests/de.rs?

I'd put a test for this in tests/api.rs but the implementation is so simple in this case that it isn't strictly required. So add one if you feel like it.

It is unfortunate that it does not work for `Pushable`, I believe this is because `Box` is a bit of a "special" type in the compiler so it gets treated differently during trait coherence ... :/ .

It might be a good idea to remove the blanked impl for `Userdata` now that `derive`s have been added as the implementation could just be generated through that instead. Need to sleep on it however.

Marwes:
Looks good, thanks!

bors r+

Marwes:
REPL tests are timing out for some reason so I just merged this https://github.com/gluon-lang/gluon/issues/625

PR #646, opened 2018-10-17
Fix Type::variant call for VmType derive

Laegluin:
Fixes `Type::variant` still being referred to as `Type::variants` in the `VmType` derive, which is currently broken for enums because of it.


Marwes:
Woops, thanks! Let me know if you want a release with this asap.

bors r+

Laegluin:
It's not super urgent for me personally, but it's probably best to release the fix soonish, since it can definitely break quite a few things.

PR #647, opened 2018-10-17
Fix try link

joprice:
Looks like the trailing slash is needed to resolve the try page.


Marwes:
Thanks! Seems I accidentally re-introduced this when porting the server to gluon. Going to fix it in the server side as well.

PR #648, opened 2018-10-21
Add global debugging status and decide whether to print out the contents of functions

### LINKED ISSUE #467: Don't print out the contents of functions to the repl

oooooba:
Fixes #467.
In this PR, a general purpose global debugging status is added to GlobalVmState, and it is decided whether to print out the contents of functions based on the state.

### DIFF
--- a/base/src/lib.rs
+++ b/base/src/lib.rs
@@ -108,3 +108,29 @@ pub fn filename_to_module(filename: &str) -> String {
 
     name.replace(|c: char| c == '/' || c == '\\', ".")
 }
+
+#[derive(Debug, Clone)]
+pub enum DebugLevel {
+    None,
+    Low,
+    High,
+}
+
+impl Default for DebugLevel {
+    fn default() -> DebugLevel {
+        DebugLevel::None
+    }
+}
+
+impl ::std::str::FromStr for DebugLevel {
+    type Err = &'static str;
+    fn from_str(s: &str) -> ::std::result::Result<Self, Self::Err> {
+        use self::DebugLevel::*;
+        Ok(match s {
+            "none" => None,
+            "low" => Low,
+            "high" => High,
+            _ => return Err("Expected on of none, low, high"),
+        })
+    }
+}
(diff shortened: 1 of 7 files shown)


base/src/lib.rs
```diff
@@ -108,3 +108,40 @@ pub fn filename_to_module(filename: &str) -> String {
...
+        Ok(match s {
+            "none" => None,
+            "low" => Low,
+            "high" => High,
+            _ => return Err("Expected on of none, low, high"),
+        })
+    }
+}
+
+impl ::std::string::ToString for DebugLevel {
```
Marwes: Could implement `fmt::Display` instead and get this for free!

vm/src/value.rs
```diff
@@ -539,16 +540,23 @@ pub struct ValuePrinter<'a> {
     pub value: Variants<'a>,
     pub max_level: i32,
     pub width: usize,
+    pub debug_level: &'a DebugLevel,
```
Marwes: Any reason why this is taken by value? It is fine either way I think but by value seems slightly simpler.

vm/src/vm.rs
```diff
@@ -583,4 +588,12 @@ impl GlobalVmState {
     pub fn get_env<'b>(&'b self) -> RwLockReadGuard<'b, VmEnv> {
         self.env.read().unwrap()
     }
+
+    pub fn get_debug_level<'b>(&'b self) -> RwLockReadGuard<'b, DebugLevel> {
```
Marwes: I think returning just `DebugLevel` here would be fine, or do you think there may be a reason not to clone/copy it?

Marwes: That's a good point but I think i'd still prefer copying. I may end up refactoring here to put more values under the same `RwLock` so being able to return the lock around just `DebugLevel` would need to change then.

vm/src/value.rs
Marwes: Ok, just wondering!

PR #654, opened 2018-11-01
feat: Implement extensible effects

Marwes:
Very much a WIP, currently working out the type system requirements.

The current WIP is largely based on the older paper http://okmij.org/ftp/Haskell/extensible/exteff.pdf as that appeared as an easier point to start at. Unfortunately the implementation requires that the open variant type (or open union in the paper) embeds a `Functor` which is impossible without additional type system and runtime support.

The followup paper on freer monads appear to avoid this issue (using GADTs) so the next step is to move the implementation towards that.

https://github.com/Marwes/gluon/blob/extensible_effects/test.glu

```
error: Implicit parameter with type `forall r . std.functor.Functor [| r]` could not be resolved.
- <test>:81:48
81 |                 flat_map loop (send (\k -> map k rest)) // ERROR
```

Something I aim to do differently is to move open unions/variants in the language proper to avoid the need for the unsafe coercions.

#### Extensible effects - An alternative to monad transformers
http://okmij.org/ftp/Haskell/extensible/exteff.pdf
https://github.com/suhailshergill/extensible-effects

#### Freer Monads, More Extensible Effects
https://github.com/lexi-lambda/freer-simple
http://okmij.org/ftp/Haskell/extensible/more.pdf

#### purescript-run
https://github.com/natefaubion/purescript-run


Marwes:
Refactored the effect implemention in the freer monad representation which wasn't as big of a problem as I expected. The functor issue is thus resolved and all cheating is removed. Everything does not fully typecheck however.

```
Expected: test.Eff (test.Error String a) Int
Found: test.Eff (test.Lift IO a) Int
2 errors were found during unification:
Types do not match:
    Expected: test.Error
    Found: test.Lift
Types do not match:
    Expected: String
    Found: IO
- <test>:107:12
107 |     do y = lift (io.applicative.wrap 123)

Expected: test.Eff (test.Lift IO a) (std.types.Result String Int)
Found: test.Eff (test.Rest b) (std.types.Result String Int)
Types do not match:
    Expected: test.Lift IO a
    Found: test.Rest b
- <test>:110:48
110 | let final : IO (Result String Int) = run_lift (run_result effect)
```

This fails because `test.Rest b` and `test.Lift IO a` are partially applied, if they were fully applied they do as they are both open variant rows (and unifies under row polymorphism). So what is needed is a way to write a set of effects as a row that is also be convertible to variants so they can be matched on when the effect row is applied.

```
// Applying an effect gives something we can `match` on, in other words a variant
// The applied type (`a`) can be used inside each variant as it is applied on each `effect` (`Effect1 a`, `Effect2 Int a`
[| Effect1, Effect2 Int | r |] a <==> (| Variant1 Content1 | Variant2 | r)
```

Doing this generically seems to at least need polymorphism over labels so `convert` can be written

```
// 'Label is a type variable of kind `Label` (instead of Type, Row, Type -> Type, ..)
let convert : [| 'Label a | r |] b -> (| 'Label a b | r) 
```

TBC

Marwes:
## Progress update

Effects basically work at the user level now and I have ported the repl, http server and lisp example to be effect based instead of the plain IO or custom monads that they used before. There may still be some API changes in the future or changes to the effect row syntax `[| state : State Int | r |]` but otherwise I'd expect everything to work without issues.

The internals are on a bit shakier ground.

I added a set of "magic macros" to convert between effect rows and variants which can actually be matched and dispatched on which isn't something I'd expect to last long term but it does the job for now. I might restrict their use the the standard library which would effectively prevent the definition of custom effects (except through composition of the ones provided). That might not be so bad, if there is a missing effect, chances are the standard library wants it anyway.

There is also a soundness issue with the effect representation. As it is, the `a` in `Eff r a` is "forgotten" when writing the effect interpreter so it is possible to violate type safety.

```
do a = my_effect
wrap (a + 1) // oops, the effect interpreter accidentally sent a `String` and not an `Int`
```

It is at least possible to define a safe interface as long as the interpreter is defined correctly but GADTs are needed to fix it properly so I have started looking into that (https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/gadt-pldi.pdf).

Lastly, while not an issue with effects, there is (yet another) bug with higher ranked types I found when trying to implement the ST effect (/ [monad](http://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Monad-ST.html)). I think I have a fix for it though, just need to work out some scoping issues with type holes. That should make it feasible to lift the `Ref` type into `ST` and pave the way for better handling of mutable state (I hope!).

I am hoping to get this done before the end of the year but GADTs may be rough. If there is an easy way to limit the unsoundness so it is contained to the standard library and it seems like this will take a while I will try and cut 0.10 before that. Otherwise that is the plan.

PR #655, opened 2018-11-05
feat: Add std.io.read/write and std.disposable

### LINKED ISSUE #653: Add implicits similar to rust's io::Read/Write
I've been playing around with some additions to `std.process`, but I noticed that for working with streams some abstractions are really needed:

- basically rust's io::Read/Write
- Something like `Disposable` for managing resources like file handles

Edit: Just added a rough impl to see if things make sense: https://github.com/Laegluin/gluon/tree/read-write/std

Laegluin:
This is a somewhat working prototype of #653.

Still missing:

- [x] docs
- [ ] tests
- [x] fix broken tests
- [x] add breaking changes to commit messages
- [x] there's a bug in `read_file_to_end` I must have introduced in the last few commits
- [ ] I'm honestly not exactly sure what gluon's multithreading model is; right now, buffered IO definitely has a few race conditions

Closes #653


src/io.rs
```diff
@@ -49,7 +49,16 @@ fn eprintln(s: &str) -> IO<()> {
 
 #[derive(Userdata)]
 #[gluon(crate_name = "::vm")]
-struct GluonFile(Mutex<File>);
+struct GluonFile(Mutex<Option<File>>);
```
Marwes: Could have sworn that rust had a `File::close(&mut self)` method but it seems that is not the case :/

src/io.rs
```diff
@@ -101,6 +94,86 @@ fn read_file_to_string(s: &str) -> IO<String> {
...
+
+    let mut file = file.0.lock().unwrap();
+    let file = unwrap_file!(file);
+    let mut buffer = Vec::with_capacity(count);
+
+    let result = unsafe {
+        buffer.set_len(count);
+        match file.read(&mut *buffer) {
+            Ok(bytes_read) if bytes_read == 0 => IO::Value(None),
+            Ok(bytes_read) => IO::Value(Some(buffer[..bytes_read].to_vec())),
```
Marwes: Could truncate the buffer here instead of allocating a new one.

src/io.rs
```diff
@@ -163,6 +236,10 @@ fn catch<'vm>(
     })
 }
 
+fn throw(msg: String) -> IO<OpaqueValue<RootedThread, A>> {
```
Marwes: This actually already exists as the `error` function. Do you feel there is a good reason to have it here as well?

vm/src/primitives.rs
```diff
@@ -30,6 +30,10 @@ pub mod array {
         array.len() as VmInt
     }
 
+    pub fn is_empty(array: Array<generic::A>) -> bool {
```
Marwes: Could implement this in gluon itself using `len`

std/io/write.glu
```diff
@@ -0,0 +1,130 @@
...
+}
+
+let write_slice ?write : [Write a] -> a -> Array Byte -> Int -> Int -> IO Int = write.write_slice
+
+let write writer buf : [Write a] -> a -> Array Byte -> IO Int =
+    write_slice writer buf 0 (array.len buf)
+
+let write_all writer buf : [Write a] -> a -> Array Byte -> IO () =   
+    let inner bytes_written =
+        if bytes_written >= array.len buf then
```
Marwes: Could hoist `array.len buf` outside the inner function

std/io/write.glu
```diff
@@ -0,0 +1,130 @@
...
+            wrap ()
+        else
+            do written = write_slice writer buf bytes_written (array.len buf)
```
Marwes: Ditto

Laegluin:
> Yeah, that is a bit unfortunate. Two gluon threads could definitely run on two OS threads accessing the same buffer at the same time which will lead to weird results (but not undefined behavior at least). Unsure what the best solution is here.

It would probably not be too hard to implement a simple mutex in gluon. `<-` would need to return the old value (maybe supporting a separate atomic bool makes more sense efficiency wise). Then it would be possible to implement a spin lock. Would be quite useful to have gluon be aware of `thread.sleep` so a sleeping thread can be suspended.

> I will try and take a second look at this as well but I am a bit short on time since I am planning and packing for a trip to Singapore in a few days.

I'm probably busy during the week anyway, have fun in Singapore! :)

src/io.rs
Laegluin: `error` is `String -> a` though, not `String -> IO a`, or am I missing something? Name's probably not the best, just added it quickly to make things work.

Marwes:
> It would probably not be too hard to implement a simple mutex in gluon. <- would need to return the old value (maybe supporting a separate atomic bool makes more sense efficiency wise). Then it would be possible to implement a spin lock. Would be quite useful to have gluon be aware of thread.sleep so a sleeping thread can be suspended.

After effects are added I will look into `Ref` again as it should really have an effect to indicate the mutability. Will look into more complete solutions then as well. Since futures are already used in gluon it might be best to try and have a "futures compatible Mutex" to avoid blocking unnecessarily.

Laegluin:
It seems that returning a `RuntimeResult<IO<T>>` has some interesting effects (it panics with 'Cannot call' whatever T is). Is that a bug or just a limitation when combining `IO` with other wrappers like `RuntimeResult`?

Marwes:
> It seems that returning a RuntimeResult<IO<T>> has some interesting effects (it panics with 'Cannot call' whatever T is). Is that a bug or just a limitation when combining IO with other wrappers like RuntimeResult?

Right! It seems that some method implementations aren't forwarded as they should be 

https://github.com/gluon-lang/gluon/blob/9addc89e89eb0858eba09af4d53df85acfd373f5/vm/src/api/mod.rs#L379-L400

https://github.com/gluon-lang/gluon/blob/master/vm/src/api/mod.rs#L1294-L1299

Laegluin:
> Right! It seems that some method implementations aren't forwarded as they should be

That's a lot simpler than I'd hoped, I'll just fix it real quick.

Marwes:
Back home again, I will look through this again tomorrow or the day after.

Laegluin:
I didn't do much anyway, I'll probably have some time this weekend.

Laegluin:
It seems there is a bug regarding enum using the `VmType` derive. This enum 
https://github.com/gluon-lang/gluon/blob/79e886d20f07a6a066cc5fcefa428d291804254b/src/io.rs#L69-L78

produces some interesting type error when trying to use it:

src/io.rs
```diff
@@ -101,6 +118,91 @@ fn read_file_to_string(s: &str) -> IO<String> {
     }
 }
 
+fn read_file(file: &GluonFile, count: usize) -> IO<RuntimeResult<Option<Vec<u8>>, String>> {
+    if count == 0 {
```
Marwes: I don't think it is worth optimizing for this case, in practice it shouldn't happen (very often) making it a minor slow down in the common case and it also shadows any errors for zero-byte reads since it never touches the file. The standard library doesn't do the check either so better to follow suit.

tests/io.rs
```diff
@@ -43,10 +49,39 @@ fn read_file() {
...
+        r#"
+        let {{ assert }} = import! std.test
+        let {{ OpenOptions, open_file_with, write_slice_file, flush_file, ? }} = import! std.io
+
+        do file = open_file_with "{path}" [Write]
+        do bytes_written = write_slice_file file [1b, 2b, 3b, 4b] 0 4
+        assert (bytes_written == 4)
+        flush_file file
+        "#,
+        path = file.path().to_str().unwrap()
```
Marwes: I find it a bit cleaner (but slightly more verbose I suppose) to return a function which takes the path as an argument instead of injecting the path via a format string.

```
\path ->
    do file = open_file_with "{path}" [Write]
    ...

// ... 
run_expr::<OwnedFunction<fn (String) -> IO<()>>>(&thread, "<top>", &text)

f.call(file.path().to_str().unwrap()).unwrap();
```

Marwes:
This looks good otherwise! Could be nice with some tests around the buffered writing as that involves some extra custom code but otherwise 👍

tests/io.rs
Laegluin: You're absolutely right, for some reason I didn't even think of that. Was probably me unconsciously being lazy :p

src/io.rs
Laegluin: Yeah, it doesn't do anything helpful and being consistent makes more sense. Should the function even return an `Option` then? I guess it's not really necessary, but it is more convenient to use because you can pattern match on it.

Laegluin:
This should be everything except for the tests for buffered IO. I'll probably get to that on the weekend. One thing I already noticed is that `std.test` doesn't really compose with functions that are with `IO`. I am not quite sure how I would handle that cleanly.

Marwes:
> This should be everything except for the tests for buffered IO. I'll probably get to that on the weekend. One thing I already noticed is that std.test doesn't really compose with functions that are with IO. I am not quite sure how I would handle that cleanly.

The story around testing `IO` is not great atm, the `Test` monad should be able to handle `IO` but it becomes a bit annoying to implement as it would require a manual monad implementation or monad transformers. I have avoided fixing it since it should come for free with extensible effects.

For reading you could test `Buffer` on an memory reader but that does not exist for writers. Not sure of a great way to handle this without fixing `Test`. Could perhaps do it like this json test does and just use plain `assert` (avoiding the `Test` monad) https://github.com/gluon-lang/gluon/blob/9addc89e89eb0858eba09af4d53df85acfd373f5/std/json/de.glu#L67-L76.

I'd be fine merging either way, maybe with doc comment on the modules that aren't tested and then revisit once `Test` is improved.

Marwes:
Working to fix https://github.com/gluon-lang/gluon/pull/655#issuecomment-440797152. Extracting worked decently but there are some bugs I am still working on.

Laegluin:
> Could perhaps do it like this json test does and just use plain assert (avoiding the Test monad)

Yeah, I think for now that's probably the best solution. Shouldn't be hard to do in that case.

Laegluin:
This should be everything. The tests live in `tests/buffered_io.{glu,rs}` for now. I also removed `std.io.types` since #661 fixed the issues with the derive.

Marwes:
Nice! I will have a look tomorrow.

In other news, I just got effects into a place where I could refactor `std.test` to using effects. https://github.com/gluon-lang/gluon/pull/654/commits/a0ab9bf07f087f676ee1f5de8ce98145d7be6e68#diff-3c8ee448e6ab89d5d1aaa5522e18b6f8 

Should be able to port your tests to use that after this is merged and I rebase the other PR!

tests/buffered_io.rs
```diff
@@ -0,0 +1,16 @@
+extern crate env_logger;
+extern crate gluon;
+
+use gluon::vm::api::IO;
+use gluon::{new_vm, Compiler};
+
+#[test]
+fn buffered_io() {
+    env_logger::init();
```
Marwes: This will error if there are multiple tests that initialize env_logger so usually I do `let _ = env_logger::try_init()` to ignore the error in case of multiple initializations. It is fine for now though since there is only a single test. And i'd like to move the test to use std.test once that works with `IO` as well so then it becomes a non-issue.

Marwes:
Looks great! Awesome work!

bors r+

PR #670, opened 2018-12-31
Release 0.10

Marwes:


Laegluin:
Looking at the example for effects, what does the `| r` mean? It's never mentioned nor used, to me it seems kind of redundant (which I'm sure it isn't!).

Laegluin:
> that is there may be more effects in the row but we don't care about those

What happens to the ignored effects? I can't really think of a situation where I'd want to ignore an effect.

Marwes:
The effects are still handled but it allows individual functions to limit the effects that they need while still letting the be used in a larger program which has more effects.

```
// Only need Eff1 here
let eff_1 : Eff [| eff_1 : Eff1 | r |] () = ...

// Only need Eff2 here
let eff_2 : Eff [| eff_2 : Eff2 | r |] () = ...

// Here we can use either effect
let eff : Eff [| eff_1 : Eff1, eff_2 : Eff2 | r |] () =
    do _ = eff_1
    do _ = eff_2
    // etc
```

If the row were instead closed

```
// Only need Eff1 here
let eff_1 : Eff [| eff_1 : Eff1 |] () = ...

// Only need Eff2 here
let eff_2 : Eff [| eff_2 : Eff2 |] () = ...

// Error, type mismatch `[| eff_1 : Eff1 |]` does not contain `eff_2 : Eff2`
let eff : Eff [| eff_1 : Eff1, eff_2 : Eff2 | r |] () =
    do _ = eff_1
    do _ = eff_2
    // etc
```

If this still seems redundant, it may help to think about it this way. When we evaluate an effect we remove it from the row `[| eff_1 : Eff1 |] -> [| |]` so a closed row would end up as the empty row instead of a row with potentially more effects `[| | r |]`. If we have an empty row we have no effects to handle and we can trivially unpack it with `run_pure : Eff [| |] a -> a`

Laegluin:
Took some time, but I get it now :D

Might not be a bad idea to add the examples to the book as well, they definitely make it a lot clearer.

PR #677, opened 2019-01-16
fix: macro resolution in 2018 by adding `local_inner_macros`

JameeKim:
I was having a problem with using `record!` macro, so figured out the problem and made a fix. More info on this in the [Rust Edition Guide](https://doc.rust-lang.org/stable/edition-guide/rust-2018/macros/macro-changes.html).


Marwes:
Thanks! Wasn't aware of this issue when updating to rust 2018.

bors r+

PR #686, opened 2019-02-12
[WIP] feat(std): add monad transformer interface, StateT, and LazyT

Etherian:
## Description
Adds a monad-transformer interface `Transformer`, a State monad transformer `StateT`, and a Lazy monad transformer `LazyT` to the standard library.

## Related PRs
Lays groundwork for Streamlike and the new parser #687.

## Status
**In Development**

### Outstanding design questions
- [x] ~~Can and should `StateT` be replaced by the effects system?~~ ([not for the time being, at least](https://github.com/gluon-lang/gluon/pull/686#issuecomment-465513284))
- [ ] Should `Transformer` require `monad` field?
- [ ] Should StateT and LazyT (and other monad transformers) be merged with their non-transformer counterparts?

### To Do
- [x] ~~Fix StateT's implicit parameter bugs~~ (#688 & #689)
- [ ] add tests
- [ ] add inline docs

## Prior Art
- StateT
  - [Haskell](https://hackage.haskell.org/package/transformers-0.5.6.2/docs/src/Control.Monad.Trans.State.Lazy.html#StateT)
  - [Idris](https://github.com/idris-lang/Idris-dev/blob/master/libs/base/Control/Monad/State.idr)
  - [Implementing Applicative (<*>) for StateT](https://stackoverflow.com/questions/27903650/implementing-applicative-for-statet)
- MonadTrans
  -  [Haskell](https://hackage.haskell.org/package/transformers/docs/src/Control.Monad.Trans.Class.html#MonadTrans)
  - [Idris](https://github.com/idris-lang/Idris-dev/blob/master/libs/base/Control/Monad/Trans.idr)


std/statet.glu
```diff
@@ -0,0 +1,83 @@
...
+
+let monad ?mo : [Monad m] -> Monad (StateT s m) =
+    let flat_map f sr : (a -> StateT s m b) -> StateT s m a -> StateT s m b = \state ->
+        sr state >>= \sout ->
+            let {value, state = state'} = sout
+            f value state'
+
+    { applicative, flat_map }
+
+// FIXME implicit parameter troubles
```
Marwes: I assume that this is the error that occur on appveyor https://ci.appveyor.com/project/Marwes/gluon/builds/22319979#L870 ? Will look into it.

Etherian: Yep. Those are the errors I'm getting on my machine. Thanks for taking a look; I wasn't able to make sense of them.

Etherian:
> Left some nits for some more idiomatic code but otherwise looks good 👍 . Some tests, (doc or in `tests/pass`) would be nice but maybe that can wait till later as `State` could be implemented in terms of `StateT` and it would then get tested by the current tests.

Thanks. I'll make those changes as soon as I get the chance. Any tests I add might take a bit longer, since that's another area I don't have much experience with.

> Something I have been thinking about is that `std`'s top level (https://gluon-lang.org/doc/nightly/std/std.html) has become rather crowded so for the next version I will likely try and move some stuff into sub-modules. Maybe it makes sense to put these under `std.transformer` or `std.monad.transformer`? Without being sure about where to put them it might make more sense to wait with that however.

I think it makes sense to merge `std.statet` with `std.state` and `std.lazyt` with `std.lazy` like in Haskell and Idris. Then they could be moved to `std.monads` with all the other monads and monad transformers.

std/statet.glu
Marwes: #689 takes care of the second error

std/lazyt.glu
```diff
@@ -0,0 +1,34 @@
+//@NO-IMPLICIT-PRELUDE
+
+let { (<<) } = import! std.function
+let { lazy, force } = import! std.lazy
+let { Applicative, Functor, Monad } = import! std.prelude
+
+type LazyT m a = Lazy (m a)
+
+let functor ?fr : [Functor m] -> Functor (LazyT m) =
```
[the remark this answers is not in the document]
Etherian: I get odd errors when attempting to omit `?fr`, `?ap`, and such in `lazyt.glu` and `statet.glu`. Here's the error removing `?fr` from the above line causes:

```
Expected: Lazy a0
Found: a0
Variable `11` occurs in `Lazy a0`.
- <std.lazyt>:12:39
12 |     let map f ma = lazy (\_ -> map f (force ma))

Expected: a1
Found: Lazy a1
Variable `5` occurs in `Lazy a1`.
- <std.lazyt>:12:20
12 |     let map f ma = lazy (\_ -> map f (force ma))
```

All the other errors seem similar in nature.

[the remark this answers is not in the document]
Marwes: Ah, the issue is that the `map` you define is the one you end up calling inside it as well

```f#
let map f ma = lazy (\_ -> map f (force ma))
 // ^                      ^
 // same
```

But inside the `map` you want to call the one imported from `std.functor`. Rename either of them and it ought to work.

std/transformer.glu
```diff
@@ -0,0 +1,11 @@
+//@NO-IMPLICIT-PRELUDE
+
+let { Monad } = import! std.prelude
+
+#[implicit]
+type Transformer t = {
```
Marwes: `Transformer` only seems to be used once in this code so maybe the abstraction isn't needed? Or can it be used to greater effect?

Marwes:
>  Can and should `StateT` be replaced by the effects system?

It think there is still value in having monad transformers as they have their uses. If they end up being replaced by effects in 99% of cases monad transformers could still be factored out of the standard library and into an external library for those who need it.

std/transformer.glu
Etherian: Maybe `Transformer` could be factored out into its own PR, but I think if monad transformers are going to be included in the library, `Transformer` is an important abstraction to have. It and `wrap_monad` are equivalent to `MonadTrans` and `lift` in Haskell, where they are used liberally.

std/lazyt.glu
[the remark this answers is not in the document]
Etherian: Ah good, thanks. That was a simple fix. But should names be available in their own definition bodies outside of recursive contexts?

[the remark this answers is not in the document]
Marwes: `rec let` was added only a few months ago as a way to define recursive values. Before that only functions could be recursive and they were always recursive by default. (since there were no way to  to mark them as recursive/non-recursive).

In the PR that added `rec let` I kept the recursive by default behaviour to avoid a breaking changes while `rec let` was still experimental. Making them only recursive when explicitly using `rec let` may be a good idea though.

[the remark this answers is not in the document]
Marwes: https://github.com/gluon-lang/gluon/issues/692

std/transformer.glu
Marwes: Good point! Lets keep it 👍

Etherian:
[started late night 3 Mar 2019]

I was writing tests for StateT when I ran into some issues. (Ignore the error mentioned on line 26. Looking back at my logs, that appears to have been due to a typo.)

```
error: Implicit parameter with type `forall m . std.functor.Functor m` could not be resolved.
- <tests.pass.statet>:32:29
32 |     assert_eq (eval_state_t ((mx >>= f) >>= g) s) (eval_state_t (mx >>= (\x -> f x >>= g)) s)
   |                             ^^^^^^^^^^^^^^^^^^^^

et alter...
```
The type checker sometimes seems to have trouble finding buried `Functor` instances.

```
error: Implicit parameter with type `std.transformer.Transformer (std.statet.StateT b)` could not be resolved.
- <tests.pass.statet>:29:40
- 29 |     let mx : StateT _ _ _ = wrap_monad mx
-    |                                        ^^
```
I'm not sure what I'm doing wrong here.

```
Expected: std.list.List Int
Found: Int
Types do not match:
    Expected: std.list.List Int
    Found: Int
- <tests.pass.statet>:44:103
44 |         test "modify eval_state_t" <| \_ -> (assert_eq (eval_state_t (modify (\x -> x + 2) *> get) 0) 2),
   |                                                                                                       ^

et cetera...
```
There's not actually an issue here. I left the unfinished code in lines 41-44 to point out an anomaly in the error messages: they refer to the `List` type when that type is not relevant to those lines. The type checker seems to be assuming the missing types in those lines are `List`s instead of referring to them generically.

Marwes:
> error: Implicit parameter with type `forall m . std.functor.Functor m` could not be resolved.
> - <tests.pass.statet>:32:29
> 32 |     assert_eq (eval_state_t ((mx >>= f) >>= g) s) (eval_state_t (mx >>= (\x -> f x >>= g)) s)
>    |                             ^^^^^^^^^^^^^^^^^^^^

It appears that records are only traversed once to pull out the implicit instances so it finds `Applicative` but not `Functor`. I have a fix which recurses fully but unfortunately it cause ambiguity errors else where since the resolver then finds `Num -> Ord` and `Ord` and there is no way to figure out that those are actually the same instance.

Going to have to sleep on this one.


```
#[implicit]
type Functor f = {
    map : forall a b . (a -> b) -> f a -> f b
}
#[implicit]
type Applicative f = {
    functor : Functor f,
}
#[implicit]
type Monad m = { functor : Applicative m }

let any x = any x

let eval_state_t f : [Functor m] -> m a -> () = ()

let associativity mx : [Monad m] -> m a -> () =
    eval_state_t mx

()

```

> error: Implicit parameter with type `std.transformer.Transformer (std.statet.StateT b)` could not be resolved.
> - <tests.pass.statet>:29:40
> - 29 |     let mx : StateT _ _ _ = wrap_monad mx
> -    |         

Here you have forgotten to export the `transformer` binding from statet.glu, add `transformer` alongside `monad` etc and it works.


> error: Expected the following types to be equal
> Expected: std.list.List Int
> Found: Int
> 1 errors were found during unification:
> Types do not match:
>     Expected: std.list.List Int
>     Found: Int
> - <tests.pass.statet>:44:103
> 44 |         test "modify eval_state_t" <| \_ -> (assert_eq (eval_state_t (modify (\x -> x + 2) *> get) 0) 2),
>   |                                                                                                       ^

Implicit resolution seems to do something weird here, haven't looked into it further.

tests/pass/statet.glu
```diff
@@ -0,0 +1,56 @@
...
+    let s = ()
+    assert_eq (eval_state_t (mx >>= wrap) s) (eval_state_t mx s)
+
+let associativity ?mo mx f g : [Monad m] -> [Show (m a)] -> [Eq (m a)] -> m a -> _ -> _ -> _ = \_ ->
+    let mx : StateT _ m _ = wrap_monad mx
+    let s = ()
+    assert_eq (eval_state_t ((mx >>= f) >>= g) s) (eval_state_t (mx >>= (\x -> f x >>= g)) s)
+
+group "statet" [
+    // should this be moved to std.monad?
```
Marwes: That might be useful but we don't have to do that here and now.

tests/pass/statet.glu
```diff
@@ -0,0 +1,56 @@
...
+        test "modify exec_state_t" <| \_ -> (assert_eq (exec_state_t (modify (\x -> x + 2) *> modify (\x -> x * 4)) 0) <| Some 8),
+        test "modify eval_state_t" <| \_ -> (assert_eq (eval_state_t (modify (\x -> x + 2) *> get) 0) <| Some 2),
+        test "put get eval_state_t" <| \_ -> (assert_eq (eval_state_t (put "hello" *> get) "") <| Some "hello"),
+        #[derive(Eq, Show)]
+        type StateOut s a = { value : a, state : s }
+        test "put get run_state_t" <| \_ -> (assert_eq (run_state_t (put "hello" *> get) "") <| Some {value = "hello", state = "hello"}),
+        test "gets eval_state_t" <| \_ -> (assert_eq (eval_state_t (gets <| (::) 1) (2 :: 3 :: Nil)) <| Some (1 :: 2 :: 3 :: Nil)),
+    ],
+    // TODO
+    // should this be moved to std.transformer?
```
Marwes: It could go in `tests/pass/transformer.glu` instead I suppsose? I don't think it is critical to add though as it is probably hard to make the test fail accidentally.

Etherian:
I decided to not include the unfinished tests, so the PR should be ready to merge. I do have a question, though, about one of the issues I had while trying to write them. Why does the implicit parameter of `(+)` not automatically resolve in the code below?

```gluon
let { wrap, (<*>) } = import! std.applicative
let { Option, ? } = import! std.option
let int @ { ? } = import! std.int
let { wrap_monad } = import! std.transformer

let mplus : Option _ = wrap (+) // requires adding ?int.num
let ma : Option Int = wrap 10
let mb : Option Int = None
let c : Option Int = mplus <*> ma <*> mb
c
```

Marwes:
> Why does the implicit parameter of (+) not automatically resolve in the code below?

Implicits are resolved just before generalisation just as in https://arxiv.org/pdf/1512.01895.pdf . The short of it is that since type inference and implicit resolution depend on each other _when_ we chose to do implicit resolution affects type inference and therefore whether the program is well typed.

Since we also want typing to be predictable, choosing to do implicit resolution just before generalization on the closest let binding is the easiest to follow solution. Most of the time it is not an issue since the type will be concrete (say `(+) : Int -> Int -> Int`) and in the other cases it can be fixed with a type signature `let mplus : [Num a] -> Option a = wrap (+)`.

As an aside, we could chose to infer that signature instead of demanding a concrete type. However I would argue that it is usually not what you want. Introducing an implicit argument makes the binding a function so using it elsewhere involves a function call, making it more costly than one might expect.

While this is basically https://wiki.haskell.org/Monomorphism_restriction which has a fair share of critque, gluon has scoped type variables by default so I think it avoids most of actual problems.

Looks like tests are failing after the rebase since you now need to import `Lazy` (and some other types) explicitly https://travis-ci.org/gluon-lang/gluon/jobs/527568322#L1771 . Went through and made the standard library types act more like library types.

Etherian:
Thanks for the explanation! Do you think it would be viable to suggest the inferred signature in the error message? That way we could keep the difference in signature explicit while also smoothing the learning curve for newbies like myself.

> Looks like tests are failing after the rebase since you now need to import Lazy (and some other types) explicitly https://travis-ci.org/gluon-lang/gluon/jobs/527568322#L1771 . Went through and made the standard library types act more like library types.

Ah, that explains it.

I won't have access to my computer over the weekend, but if you made the tweaks I could accept them over the phone. Or I could just do it Monday.

Marwes:
> Do you think it would be viable to suggest the inferred signature in the error message?

Good idea! Created an issue https://github.com/gluon-lang/gluon/issues/725

PR #688, opened 2019-02-12
fix(check): Subsume implicit functions with forall correctly

Marwes:
The previous, naive subsume_implicit function did not traverse the
functions simultaneously so the implicit arguments would be lost if the
left side had a `forall` at the top.

Fixes an issue found in #686

cc @Etherian


Marwes:
There is another error in #686 as well which might be a compiler error but I am not sure yet (a type signature might be able to work around it).

PR #735, opened 2019-05-20
feat(codegen): Map Rust's struct enums to records in Gluon. (fix #729)

### LINKED ISSUE #729: Struct enums are not implemented as enums of records in the codegen.
Let's consider the following enum in Rust:
```Rust
enum Point {Cartesian {x: f64, y: f64}, Polar {r: f64, phi: f64} }
```
Currently the codegen assumes that the corresponding Gluon definition looks like this:
```Haskell
type Point = | Cartesian Float Float | Polar Float Float
```
However, a reasonable person would expect the following Gluon type instead:
```Haskell
type Point = | Cartesian {x: Float, y: Float} | Polar {r: Float, phi: Float}
```
Is there any reason to not use this more intuitive form?

If you agree with my opinion, I would be willing to work on a PR to fix it.

Context:
https://github.com/gluon-lang/gluon/blob/master/codegen/tests/derive_getable.rs#L107
https://github.com/gluon-lang/gluon/blob/master/codegen/tests/derive_pushable.rs#L235

mluszczyk:
@Marwes @Laegluin: please review.

### DIFF
--- a/codegen/src/getable.rs
+++ b/codegen/src/getable.rs
@@ -243,10 +243,10 @@ fn gen_variant_match(ident: &Ident, tag: usize, variant: &Variant) -> TokenStrea
             }
         }
         Fields::Named(FieldsNamed { named, .. }) => {
-            let cons = gen_struct_variant_cons(named);
+            let cons = gen_struct_variant_cons(ident, variant_ident, named);
 
             quote! {
-                #tag => #ident::#variant_ident#cons
+                #tag => #cons
             }
         }
     }
@@ -273,7 +273,7 @@ where
     }
 }
 
-fn gen_struct_variant_cons<'a, I>(fields: I) -> TokenStream
+fn gen_struct_variant_cons<'a, I>(ident: &Ident, variant_ident: &Ident, fields: I) -> TokenStream
 where
     I: IntoIterator<Item = &'a Field>,
 {
@@ -283,9 +283,9 @@ where
             .ident
             .as_ref()
             .expect("Struct fields always have names");
-
+        let quoted_field_ident = format!("{}", quote! { #field_ident });
         quote! {
-            #field_ident: if let Some(val) = data.get_variant(#idx) {
+            #field_ident: if let Some(val) = inner_data.lookup_field(vm, #quoted_field_ident) {
                 <#field_ty as _gluon_api::Getable<'__vm, '__value>>::from_value(vm, val)
             } else {
                 panic!("Enum does not contain data at index '{}'. Do the type definitions match?", #idx)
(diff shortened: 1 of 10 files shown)


vm/src/api/de.rs
```diff
@@ -856,6 +856,38 @@ impl<'de, 'a, 't> VariantAccess<'de> for Enum<'a, 'de, 't> {
     where
         V: Visitor<'de>,
     {
-        de::Deserializer::deserialize_seq(self.de, visitor)
+        let typ = resolve::remove_aliases_cow(self.de.state.env, &mut NullInterner, self.de.typ);
+        match (self.de.input.as_ref(), &**typ) {
+            (ValueRef::Data(data), &Type::Variant(ref row)) => {
```
Marwes: Quite a lot of rightward drift here, would be easier to read if it is rewritten in this style

```rust
let (data, row) = match (self.de.input.as_ref(), &**typ) {
    (ValueRef::Data(data), &Type::Variant(ref row)) => (data, row),
    _=> return Err(..)
}
```

codegen/src/pushable.rs
```diff
@@ -89,11 +89,23 @@ fn derive_enum(
...
-        quote! {
-            #pattern => {
-                #push_impl
+        match &variant.fields {
+            Fields::Named(_) => {
+                let push_impl = gen_push_impl(None, &field_idents, &field_types);
+                quote! {
+                    #pattern => {
+                        #push_impl
+                        ctx.context().push_new_data(vm, #tag as _gluon_types::VmTag, 1)?;
```
Marwes: This should be `push_new_record` instead of a plain data. You will need to create `InternedStr` via `intern` for each field. You can test that it works by passing the record to a row polymorphic function which extracts one of the fields.

```f#
let extract r : { key: String | r } -> String = r.key
```

vm/src/api/de.rs
```diff
@@ -856,6 +856,38 @@ impl<'de, 'a, 't> VariantAccess<'de> for Enum<'a, 'de, 't> {
...
+                match row.row_iter().nth(data.tag() as usize) {
+                    Some(field) => {
+                        let typ = resolve::remove_aliases_cow(
+                            self.de.state.env, &mut NullInterner, &field.typ);
+                        match (data.get_variant(0).unwrap().as_ref(), &**typ) {
```
Marwes: Could Avoid the `unwrap` here and fold the `None` case into "Unable to deserialize"

codegen/src/getable.rs
```diff
@@ -283,19 +283,27 @@ where
...
             .as_ref()
             .expect("Struct fields always have names");
-
+        let quoted_field_ident = format!("{}", quote! { #field_ident });
         quote! {
-            #field_ident: if let Some(val) = data.get_variant(#idx) {
+            #field_ident: if let Some(val) = inner_data.lookup_field(vm, #quoted_field_ident) {
                 <#field_ty as _gluon_api::Getable<'__vm, '__value>>::from_value(vm, val)
             } else {
                 panic!("Enum does not contain data at index '{}'. Do the type definitions match?", #idx)
```
Laegluin: Just a nitpick, but this error message requires some rewording. That would also remove the need for the extra `enumerate` as far as I can tell.

codegen/src/pushable.rs
Marwes: This was correct as it was, my bad. Tested that it worked and fixed the other nits in #773 . Thanks!

PR #738, opened 2019-05-25
feat(std): Add Indexable interface

Etherian:
## Description
Adds an `Indexable` interface which abstracts over collections whose elements can referred to by integer indexes, e.g. Strings, Arrays, and Lists.

## Status
**In Development**

### Outstanding design questions:
*None currently*

### To Do
- [ ] Add useful functions?
- [ ] Add implementations for standard types
- [ ] Add inline documentation

### Prior Art
N/A

### DIFF
--- a/std/indexable.glu
+++ b/std/indexable.glu
@@ -1,0 +1,11 @@
+type Indexable ind = {
+    index : ind a -> Int -> a
+}
+
+let index ?ind : [Indexable ind] -> ind a -> Int -> a = ind.index
+
+{
+    Indexable,
+
+    index,
+}


Marwes:
Do you have a clear use case in mind for this? I find that it is hard to get a good API without having 2-3 example uses of something.

Some ideas

* Generalizing to arbitrary keys? (supporting `Map`)
* Length function? Doesn't connect to indexable, but without knowing which indices are in bounds it becomes hard to use
* Returning an `Option` to handle index out of bounds?

Etherian:
I'm closing this for now because, while I think having some sort of unified indexing interface would be very handy, I don't think I'm knowledgable enough to specify it properly. I had been intending to use this interface in my parser lib, but I've decided a simpler method will suffice.

PR #752, opened 2019-06-24
feat: Automatically unroot values stored in `Gc` allocated values

### LINKED ISSUE #746: Allow userdata to contain gluon values without reference cycles
I'm using gluon to add some user functionality to a program, like this:

    process_file {
      // do some built in processing
      // IF no script was specified at the command line, stop here.
      // create new gluon vm
      // create user library
      // compile and run gluon script => gluon script exports some callbacks
      // repeatedly call the gluon callbacks
      // vm should be freed at this point
    }

When process_file is called for >100 files without a script, the memory usage stays at ~10MB all the time (measured using the time command). When a script is specified, the memory usage at the end of the program (according to valgrind) is >100MB. When the 100 files are specified multiple times at the command line, the memory usage rises accordingly. i.e., 4 times these 100 files, 400MB memory used at the end.

Where to look for the memory leak?

Marwes:
`RootedThread` and `RootedValue` are reference counted pointers so
storing them in a value allocated on the GC heap will create a cycle
which will never be dropped.

To avoid this we detect when such a value is allocated on the GC heap
and remove the value from the root list.

Fixes #746

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -219,7 +219,7 @@ dependencies = [
  "rayon 1.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "reqwest 0.9.17 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.91 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_derive 1.0.91 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.93 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
  "url 1.7.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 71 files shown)


hce:
Hmm: I'm using rust stable. I'm getting this error when compiling gluon as part of my project:

   Compiling gluon_format v0.11.2 (/Users/hc/gluon-test/gluon/format)
error: expected one of `+`, `,`, `::`, or `{`, found `where`
   --> gluon/vm/src/thread.rs:525:18
    |
525 |         #[derive(DeserializeState)]
    |                  ^^^^^^^^^^^^^^^^ expected one of `+`, `,`, `::`, or `{` here

error: aborting due to previous error

Marwes:
I fixed that in `serde_derive_state`. You can do `cargo update -p serde_derive_state` to fix it (I should update the `Cargo.toml` with the min version).

Marwes:
Update the `Cargo.toml` file on this branch so it will pick the correct version if you update your git dependency.

PR #760, opened 2019-07-23
feat(std): Export std.array.monoid and std.stream.uncons

lePerdu:
These are defined, but not exported.


Marwes:
Thanks!

bors r+

PR #762, opened 2019-07-25
Add `#[gluon_userdata(clone)]` attribute

### LINKED ISSUE #761: Can't export Userdata value from module
## Use Case
I've playing around with using gluon in a text editor. It's using [`xi_rope::Rope`](https://docs.rs/xi-rope/0.3.0/xi_rope/rope/type.Rope.html) to store text, which is marshalled to gluon as `Userdata` (in a new-type wrapper). This works fine, but when I tried to implement `Monoid` for it, gluon gave a runtime error saying "Userdata cannot be cloned" when importing the module, because of the `empty` field of `Monoid`. The same error occurs when exporting a Userdata literal from an `ExternModule`. I have a stripped-down example below:

## Minimal example
```rust
#[derive(VmType, Trace, Userdata, Clone, Debug)]
#[gluon_trace(skip)]
#[gluon(vm_type = "mystring.MyString")]
struct MyString(String);

fn load_mystring(vm: &Thread) -> VmResult<ExternModule> {
    vm.register_type::<MyString>("mystring.MyString", &[])?;

    ExternModule::new(vm, record! {
        from_str => primitive!(
            1, "mystring.from_str", |s| MyString(s)),
        // This causes an error
        empty => MyString(String::new()),
    })
}

fn main() {
    let vm = gluon::new_vm();
    add_extern_module(&vm, "mystring", load_mystring);

    let mut compiler = Compiler::new();
    // Import to module to bring in the MyString type
    match compiler.run_expr::<OpaqueValue<&Thread, Hole>>(
        &vm, "", "import! mystring")
    {
        Ok(_) => {}
        Err(e) => { panic!("{}", e); }
    };

    match compiler.run_expr::<UserdataValue<MyString>>(
        &vm, "", r#"
            let mystring = import! mystring
            mystring.empty
            // This works fine: mystring.from_str "abc"
        "#)
    {
        Ok((result, _)) => { println!("{:?}", result); }
        Err(e) => { panic!("{}", e); }
    };
}
```

Error:
```
thread 'main' panicked at 'error: Userdata cannot be cloned
- <>:1:9
  |
1 | import! mystring
  |         ^^^^^^^^
  |
', src/main.rs:48:21
```

Commenting out the `empty` export and returning the expression `mystring.from_str "abc"` works fine.

lePerdu:
If set and the deriving type implements `Clone`, the gluon type will be
cloneable.

Fixes #761

### DIFF
--- a/codegen/src/attr.rs
+++ b/codegen/src/attr.rs
@@ -9,7 +9,10 @@ use syn::{
 
 fn get_gluon_meta_items(attr: &syn::Attribute) -> Option<Vec<syn::NestedMeta>> {
     if attr.path.segments.len() == 1
-        && (attr.path.segments[0].ident == "gluon" || attr.path.segments[0].ident == "gluon_trace")
+        && (   attr.path.segments[0].ident == "gluon"
+            || attr.path.segments[0].ident == "gluon_trace"
+            || attr.path.segments[0].ident == "gluon_userdata"
+        )
     {
         match attr.interpret_meta() {
             Some(List(ref meta)) => Some(meta.nested.iter().cloned().collect()),
@@ -31,6 +34,7 @@ pub struct Container {
     pub vm_type: Option<String>,
     pub newtype: bool,
     pub skip: bool,
+    pub clone: bool,
 }
 
 impl Container {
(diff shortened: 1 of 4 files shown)


Marwes:
Thanks! I will try and update the docs later and expand the 'unable to clone' error message to point to this.

bors r+

PR #775, opened 2019-08-12
feat(repl): add a --no-std option to the standalone interpreter / REPL

### LINKED ISSUE #751: Add a flag to gluon.exe to ignore its internal library modules
It can be difficult to test changes to the standard library using the standalone interpreter / REPL because it finds the modules in its internal standard library before checking the file system for the updated files.

Adding a command line flag to skip its internal modules, or maybe just check the filesystem first, would be a nice convenience.

---

This might be a good introduction for me to Gluon's Rust codebase, if someone could point me to where and what needs to be changed.

Etherian:
## Description
Adds a command-line switch `--no-std` to the standalone interpreter / REPL which causes it to ignore its internal standard library modules when searching for modules to import.

This should make less awkward to use the REPL to test changes to the standard library.

resolves #751

### DIFF
--- a/repl/src/main.rs
+++ b/repl/src/main.rs
@@ -79,7 +79,7 @@ impl ::std::str::FromStr for Color {
             "always" => Always,
             "always-ansi" => AlwaysAnsi,
             "never" => Never,
-            _ => return Err("Expected on of auto, always, always-ansi, never"),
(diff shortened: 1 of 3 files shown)


Marwes:
LGTM, thanks!

bors r+

PR #778, opened 2019-08-18
fix: Don't depend on ansi_term in windows to make windows 7 work

### LINKED ISSUE #777: "procedure entry point CreateFile2 could not be located" when running gluon.exe
After pulling the most recent commits on `master` and updating my Rust toolchain to 1.37.0, building the REPL yields a broken executable. When run from PowerShell it immediately exits with code -1073741511, and when run from `cmd.exe` (well, actually [Clink](http://mridgers.github.io/clink)) an error dialog appears, stating 
> The procedure entry point CreateFile2 could not be located in the dynamic link library KERNEL32.dll.

It's odd that the binary would be requesting CreateFile2, since I'm building for and on Windows 7 and [the CreateFile2 function was added in Windows 8](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfile2#requirements).

This behavior is unlikely to be related to Gluon itself but may be caused by a bug in one of its dependencies, so I thought I should mention it. ~~I'll also be opening an issue in the Rust repository in case there is a bug in the toolchain.~~ [was unnecessary]

Marwes:
Fixes #777

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ branches:
 
 matrix:
   include:
-  - rust: nightly-2019-05-20
+  - rust: nightly-2019-08-30
   - rust: beta
   - rust: stable
     env: ARCH=i686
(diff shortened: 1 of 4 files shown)


Etherian:
These changes work, by the way. I'm using them in my local repo.

PR #779, opened 2019-08-24
[WIP] feat(std): add time module to standard library

Etherian:
## Description
Adds a `time` module to the standard library which exposes the functionality of Rust's `std::time`.

## Status

**In Development**

## Outstanding design questions
- [ ] How should `std.time` be structured -- `std.time.Instant.now`, `std.time.instant_now`, `std.instant.now`, or something else?
- [ ] Should `std.time` be behind a feature? Should that feature be default?

## Future Directions
- Add more full-featured date-and-time functionality to the standard library? [Chrono](https://github.com/chronotope/chrono)?

PR #780, opened 2019-08-30
fix: Add tests and fix the regression with clone_userdata

Marwes:
cc #779
cc @Etherian


Etherian:
With this PR, compilation succeeds and `std.time` imports and functions properly. I did run into a couple bugs along the way that I'll report in a bit.

What do you think is causing the breakage on Beta?

Marwes:
No clue... Haven't been able to reproduce locally either...

Marwes:
Figured it out, it was `sccache`. Updated it and it should work after I merge #778

PR #791, opened 2019-10-15
feat(std): add a few functions to std.test & std.effect.error

Etherian:
Adds `some_or_throw` to `std.effect.error` and  `assert_ok`, `assert_err`, `assert_errorless`, and `assert_erroring` to `std.test`. I could also add `assert_lt`, `assert_lte`, `assert_gt`, and `assert_gte` if you like.

### DIFF
--- a/std/test.glu
+++ b/std/test.glu
@@ -70,8 +70,7 @@ let assert_success : [Show e]
         -> Eff [| error : Error e, writer : Test | r |] a
         -> Eff [| writer : Test | r |] ()
     =
-    run_error
-        >> flat_map assert_ok
+    run_error >> flat_map assert_ok
 
 let assert_throws : forall e .
     [Show a] -> Eff [| error : Error e, writer : Test | r |] a -> Eff [| writer : Test | r |] ()
--- a/std/test.glu
+++ b/std/test.glu
@@ -2,21 +2,23 @@
 
 let string = import! std.string
 let { wrap } = import! std.applicative
-let prelude @ { Semigroup } = import! std.prelude
+let { flat_map } = import! std.monad
 let float = import! std.float
 let int = import! std.int
 let list @ { List, ? } = import! std.list
 let { Foldable, foldl } = import! std.foldable
 let { Option } = import! std.option
-let { (<>) } = import! std.semigroup
+let { Result } = import! std.result
+let { Semigroup, (<>) } = import! std.semigroup
 let { error } = import! std.prim
-let { id } = import! std.function
+let { (>>), id } = import! std.function
 let { ? } = import! std.io
 
 let { assert } = import! std.assert
 
 let effect @ { Eff, ? } = import! std.effect
 let { Writer, run_writer, tell } = import! std.effect.writer
+let { Error, run_error } = import! std.effect.error
 let { Lift, run_lift } = import! std.effect.lift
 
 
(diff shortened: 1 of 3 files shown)


std/test.glu
```diff
@@ -38,6 +40,22 @@ let assert_neq l r : [Show a] -> [Eq a] -> a -> a -> Eff [| writer : Test | r |]
...
+    match res with
+    | Ok _ -> wrap ()
+    | Err e -> tell (Cons ("Assertion failed: found error " <> show e) Nil)
+
+let assert_err res : [Show a] -> Result e a -> Eff [| writer : Test | r |] () =
+    match res with
+    | Ok x -> tell (Cons ("Assertion failed: expected error, found " <> show x) Nil)
+    | Err _ -> wrap ()
+
+let assert_errorless : [Show e] -> Eff [| error : Error e, writer : Test | r |] a -> Eff [| writer : Test | r |] () =
```
Marwes: Maybe `assert_success`?

std/test.glu
```diff
@@ -38,6 +40,22 @@ let assert_neq l r : [Show a] -> [Eq a] -> a -> a -> Eff [| writer : Test | r |]
...
+    run_error >> flat_map assert_ok
+
+let assert_erroring : forall e . [Show a] -> Eff [| error : Error e, writer : Test | r |] a -> Eff [| writer : Test | r |] () =
```
Marwes: Maybe `assert_throws`?

std/test.glu
```diff
@@ -38,6 +40,22 @@ let assert_neq l r : [Show a] -> [Eq a] -> a -> a -> Eff [| writer : Test | r |]
...
 
+let assert_ok res : [Show e] -> Result e a -> Eff [| writer : Test | r |] () =
+    match res with
+    | Ok _ -> wrap ()
+    | Err e -> tell (Cons ("Assertion failed: found error " <> show e) Nil)
+
+let assert_err res : [Show a] -> Result e a -> Eff [| writer : Test | r |] () =
+    match res with
+    | Ok x -> tell (Cons ("Assertion failed: expected error, found " <> show x) Nil)
+    | Err _ -> wrap ()
```
Marwes: `Could return the error instead of `()`

std/test.glu
```diff
@@ -38,6 +40,22 @@ let assert_neq l r : [Show a] -> [Eq a] -> a -> a -> Eff [| writer : Test | r |]
     if l /= r then wrap ()
     else tell (Cons ("Assertion failed: " <> show l <> " == " <> show r) Nil)
 
+let assert_ok res : [Show e] -> Result e a -> Eff [| writer : Test | r |] () =
+    match res with
+    | Ok _ -> wrap ()
```
Marwes: `Could return the value instead of `()`

Etherian: I considered that, but then what would the failing branch return?

Marwes: Ah, right. Because the error is just written, there is no error thrown. I think it would make sense to move these asserts into `std.effect.error` and have them throw instead. Then make tests instead catch errors and write them. The current setup is kind of strange in that it continues even if an assert fails.

That's another PR though as there might be fallout from such a change.

Marwes:
Tests are failing because the file isn't formatted anymore

Marwes:
Since formatting is still somewhat janky, feel free to disable it for this file and log an issue so it can be fixed later.

Etherian:
How would I disable the formatting check?

Marwes:
There is an explicit `#[test]` which you can `#[ignore]` https://travis-ci.org/gluon-lang/gluon/jobs/598480363#L2256 . But if `gluon fmt std/test.glu` works then that is better than disabling

PR #862, opened 2020-08-09
Add a few QoL functions to the standard library

Etherian:
Adds
- `unwrap_or` and `to_result` to `Option`
- `unwrap_ok_or`, `unwrap_err_or`, and `to_option` to `Result`
- Kleisli composition operators (`<=<` and `>=>`)


Marwes:
Looks like you forgot to import `<<`

Etherian:
Fixed.

On the subject, what do you think of adding `std.function` to the prelude? I feel like I import most of it in every script.

Marwes:
> On the subject, what do you think of adding std.function to the prelude? I feel like I import most of it in every script.

👍 Probably just `<<` and `<|` I think those are the most common to use. Maybe also Monad, Category etc as well?

Etherian:
> > On the subject, what do you think of adding std.function to the prelude? I feel like I import most of it in every script.
> 
> +1 Probably just `<<` and `<|`? I think those are the most common to use. Maybe also Monad, Category etc as well?

I agree on both counts.

PR #883, opened 2020-09-14
doc: Explain the vm_type attribute better

### LINKED ISSUE #882: Confusing docs on marshalling types?
I'm trying to marshal a Rust struct as a Gluon record, and I've derived `VmType` and `Pushable` for the type. However, it seems like when I use the `#[gluon(vm_type = "...")]` attribute (and also registering the type), the resulting record would not have any names. For a testing struct like:
```
#[derive(Clone, Debug, Trace, VmType, Pushable)]
#[gluon_trace(skip)]
pub struct Test {
    pub int: i64,
    pub string: String,
}
```

This would result in a record like `{ 42, "asdf" }`, and trying to access fields will result in `error: Type 'test.Test' is not a type which allows field accesses`. This is confusing because the docs specifies: 

>  `VmType`, `Getable` and `Pushable` can be derived for types that consist only of types that already implement the respective traits. In the case of `VmType` you also have to specify the Gluon type the Rust type maps to, using the `#[gluon(vm_type = "<gluon_type>")]` attribute.

Edit: This is under the assumption that I need to return `Test` from an exported function, in which case the compiler tells me that I need `VmType` on `Test`.

Marwes:
@crides does this explain it?

Closes #882

### DIFF
--- a/book/src/marshalling-types.md
+++ b/book/src/marshalling-types.md
@@ -39,19 +39,36 @@ code using the `api::typ::make_source` function.
 
 ### Using derive macros
 
-Add the `gluon_codegen` crate to your `Cargo.toml` and import it:
+Add the `gluon_codegen` crate to your `Cargo.toml` this lets you import and derive the
+`VmType`, `Getable`, `Pushable` and `Userdata` traits.
+
+`VmType`, `Getable` and `Pushable` can be implemented on any type which only consists of types which in turn implements
+these traits whereas `Userdata` can be derived for any type as long as it is `Debug + Send + Sync` and has a `'static`
+lifetime.
+
+Sometimes when deriving `VmType` you do not want to define a new type. In this case you can use the `vm_type` attribute
+to point to another, compatible type. See the marshalling example for the complete source for the examples below.
 
 ```rust,ignore
-#[macro_use]
-extern crate gluon_codegen;
+// Using `vm_type` to point to compatible type defined in gluon
+#[derive(Debug, PartialEq, VmType, Getable)]
+#[gluon(vm_type = "std.list.List")]
+enum List<T> {
+    Nil,
+    Cons(T, Box<List<T>>),
+}
+
+// Defines an opaque type with Userdata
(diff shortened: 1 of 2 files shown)


crides:
Yes, that does clear up the confusion!

If I may go a bit further on the original question, how is registering the type different from including the type inside the module? Does the naming of the type (e.g. `example.Either`) matter  i.e. does it have to be the same path when the type is imported (e.g. if I register the type as `example.Either`, does the type `Either` have to be inside the module `example`)? Thanks!

Marwes:
If you look at the `WindowHandle` type in `marshalling` you can see that `register_type` is called with the path that vm_type wants.

crides:
So using the `WindowHandle` example. There are 3 paths in the example:

```rust
#[derive(Userdata, Trace, Clone, Debug, VmType)]
#[gluon_userdata(clone)]
// Lets gluon know that the value can be cloned which can be needed when transferring the value between threads
#[gluon(vm_type = "WindowHandle")] //1
struct WindowHandle {
    id: Arc<u64>,
    metadata: Arc<str>,
}

fn load_mod(thread: &gluon::Thread) -> vm::Result<ExternModule> {
    thread.register_type::<WindowHandle>("WindowHandle", &[])?; //2

    let module = record! {
        type WindowHandle => WindowHandle, //3
        create_hwnd => primitive!(2, create_hwnd),
        id => primitive!(1, id),
        metadata => primitive!(1, metadata),
        default_hwnd => create_hwnd(0, "default".into()),
    };

    ExternModule::new(thread, module)
}
```

The 3 are: 1) the `vm_type` attribute, 2) the first argument to `register_type`, and 3) the actual import path for the type (i.e. if I import the module as `window`, the actual path to the type would be `window.WindowHandle`). So 1 and 2 have to be the same\*. So can 3 be anything? How would 3 be used? Would the actual path to the type be 1& 2 or 3?

\*side question: can you point some rust type using the `vm_type` attribute to some Gluon type, and not register the type?

PR #886, opened 2020-09-19
 feat: Compile block expressions as monadic sequences

Marwes:
Rather than removing block expressions entirely, this changes them
to behave as sequence (`seq`) expressions.

```f#
io.println "Hello"
io.println "World"

// Is now equivalent to
seq io.println "Hello"
io.println "World"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -811,6 +811,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "expect-test"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ceb96f3eaa0d4e8769c52dacfd4eb60183b817ed2f176171b3c691d5022b0f2e"
+dependencies = [
+ "difference",
+ "once_cell",
+]
+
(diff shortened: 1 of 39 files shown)


Etherian:
This sounds like `do` blocks from Haskell but more implicit. I may be biased by my narrow perspective as an amateur hobbyist, but I have always found `do` blocks hard to understand and use because they obscure the mechanics and intuitions of monads.

I would personally prefer the compiler suggest lightweight combinator(s), like `*>` or `seq`, when offered consecutive monad expressions instead of silently accepting them.

Marwes:
Fair point. As a counter argument, I have always thought of `do` blocks as a way to bridge the gap between imperative code and monadic one. So having the "default" way of writing sequential statements "just work"  lets people coming from that angle still get stuff working without needing to grok monads.

Marwes:
bors r+

Made it so the formatter keeps `seq` if it is written so both ways of writing it are equally viable now. Probably will end up removing one of the ways but for now this prevents people from writing it without `seq` and being confused that nothing happens.

Issue #664, opened 2018-12-12
Add formatting macros


Marwes:
Similar to https://doc.rust-lang.org/nightly/std/fmt/index.html. Would use https://gluon-lang.org/doc/nightly/std/show.html to format values.

PR #897, opened 2020-12-14
Format macro

bch29:
Includes changes from #896 

Addresses #664 

This is a preliminary PR to help diagnose why this isn't working. I am happy to accept discussion/changes to how the macro works.


src/format_macro.rs
```diff
@@ -0,0 +1,258 @@
...
+const OPEN_BRACE: &'static str = "{";
+const CLOSE_BRACE: &'static str = "}";
+
+struct FindExprResult<'a> {
+    prefix: &'a str,
+    expr: &'a str,
+    remaining: &'a str,
+}
+
+fn find_expr<'a>(format_str: &'a str) -> Result<Option<FindExprResult<'a>>, macros::Error> {
```
Marwes, line 204: Allowing arbitrary expressions inside the braces has some more wrinkles than just counting the braces. Consider that braces may also appear inside a string or character

```f#
format! "test {x + '}'} abc"
```
This would pick the first closing brace as the closer whereas it ought to be the second.

That could perhaps be fixed by parsing as an expression after finding the `{`, however then there is another problem with line comments.

```f#
format! "test {x // print x }"
```
The comment probably shouldn't contain the `}` here but if we naively parse everything after the `{` it would be.

It may be better to just start by allowing single variables in the format string (and possible positional arguments?)

src/format_macro.rs
```diff
@@ -0,0 +1,258 @@
...
+    let prefix_end_ix = match format_str.find(OPEN_BRACE) {
```
Marwes, line 205: This will skip over any `}` in the "prefix" which I think should be an error

src/format_macro.rs
bch29: This is a good point. We could forbid comments in the embedded expressions: try to parse starting from the '{' then walk the resulting expression and raise an error if a comment is found.

Alternatively, allowing single variables could be a fair compromise because it would be forward compatible with full expressions, if we allow capturing variables in local scope

```gluon
let x = 1
in format! "x is {x}"
```

instead of explicitly passing to the macro

```gluon
format! "x is {x}" {x = 1}
```

Thoughts on this?

As for positional arguments, I am not convinced they are worth having but I can add them if you disagree.

bch29:
One question I had: this works somewhat unpleasantly with strings at the moment, because `show` adds quotes.

```
let x = "hello"
format! "x is {x}"
```

becomes `x is "hello"`.

I could look up type information and handle strings differently, but that feels ugly. Perhaps we should add a `Display` typeclass for this purpose, which would render strings without quotes?

src/format_macro.rs
Marwes: > As for positional arguments, I am not convinced they are worth having but I can add them if you disagree.

True, though that would require some lexer tweaks since it will just ignore comments atm.

> Alternatively, allowing single variables could be a fair compromise because it would be forward compatible with full expressions, if we allow capturing variables in local scope

Yeah, :+1: .

> As for positional arguments, I am not convinced they are worth having but I can add them if you disagree.

Yeah, perhaps. Lets wait with positional as they complicates things in combination with captures.

bch29:
Added a `Display` typeclass. I have left `Show` as-is: I think if it is changed to `Debug` it should be done in a separate PR.

I now support `{x}` for `display` and `{x:?}` for `show`, similar to rust.

Now, arbitrary expressions that do not contain braces or `:` are allowed in the format subexpressions. This was actually simpler for me to implement than restricted to single variable names and it allows some nice flexibility, for example:

```
let foo = { x = 1, y = 2}
format! "x is {foo.x}, y is {foo.y}"
```

Let me know whether you agree that this is a good idea.

Marwes (approved):
Some test failures related to `display` not being imported but otherwise LGTM 👍

scarf005:
Hi, could there be any chance this PR could get traction? being able to interpolate strings sounds really useful

PR #948, opened 2023-03-19
mark freeze_expr function as unsafe

kitcatier:
https://github.com/gluon-lang/gluon/blob/27cec6c8ef90de7341d22a2beeade532680aa40c/vm/src/core/mod.rs#L602-L614
https://github.com/gluon-lang/gluon/blob/27cec6c8ef90de7341d22a2beeade532680aa40c/vm/src/core/mod.rs#L616-L635
Hi, it is not a good choice to mark the entire function body as unsafe, which will make the caller ignore the safety requirements that the function parameters must guarantee, the developer who calls the freeze_expr function may not notice this safety requirement.

And inside the function body, mem::transmute is used twice, which is an unsafe operation. The first time is to convert the type of allocator.clone() from Arc<Allocator> to Arc<Allocator<'static>>, and the second time is to convert expr of type CExpr to type CExpr<'static>. These operations will make the type system unable to guarantee the type safety of the code, so it is necessary to use the unsafe keyword to mark the function and remind users to use it with caution.

Marking them unsafe also means that callers must make sure they know what they're doing.

### DIFF
--- a/vm/src/core/interpreter.rs
+++ b/vm/src/core/interpreter.rs
@@ -690,15 +690,15 @@ impl<'a, 'e> Compiler<'a, 'e> {
                         key,
                         match value.as_ref().map(|b| &b.bind) {
                             Some(Binding::Expr(Reduced::Local(expr))) => Binding::Expr(
-                                Reduced::Local(crate::core::freeze_expr(allocator, expr)),
+                                Reduced::Local(unsafe { crate::core::freeze_expr(allocator, expr) }),
                             ),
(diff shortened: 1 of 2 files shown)


Marwes:
These functions should be safe to the best of my knowledge. The idea that I tried to explain with the comment is that the reference to the expression is guaranteed to be from the arena allocator (enforced by the lifetime annotations), using that knowledge we can return a new value without any lifetimes since the allocator which the reference lives in will live at least as long as the reference itself, thus it is impossible to do any use-after-free with it.

If you have a counter example that would allow that, please let me know.

PR #961, opened 2023-09-24
Correct typo

ShalokShalom:
Correct typo


Marwes:
Thanks! Looks like a fat-fingered that when fixing CI a couple of weeks ago.

PR #964, opened 2024-03-03
Fix a typo in README.md

DenisGorbachev:


correabuscar:
was wondering what's up with that dangling `I` in readme, weird ;p

